PR #22, opened 2017-12-12
add tests to autotools

adammoody:
Include test programs from tests in autotools build.

### Description

Build and link test programs with unifycr library.

### Types of changes

### Checklist:


README.md
```diff
@@ -27,44 +27,49 @@ Below is the guide on how to install and use
...
 
-```
-shell $ ./autogen.sh             # this will generate the build script
-shell $ ./configure --prefix=/install/path/you/want --enable-debug
-shell $ make
-shell $ make install
-```
+In order to use gotcha for I/O wrapping install the latest release:
+https://github.com/LLNL/GOTCHA/releases
+
```
nedbass: Not to harp too much on the packaging theme, but do we want to make spack the officially recommended way of installing dependencies? Then we can drop the advice about using --with-gotcha and setting LD_LIBRARY_PATH. Instead the README.md can say something like:

```
To install leveldb and gotcha and set up your build environment, we recommend
using the spack package manager. For example, the following commands will
build and install leveldb and gotcha using spack and configure your environment
so UnifyCR can find the libraries and headers it needs:

git clone https://github.com/spack/spack
. spack/share/spack/setup-env.sh
spack install leveldb
spack install gotcha
spack install environment-modules
spack load environment-modules
source <(spack module loads gotcha leveldb)
```

adammoody: Sure.  I'm a fan of anything that makes things easier for the user, so this sounds good to me.  I'll keep some of the other instructions around for a bit in case we have users who can't use spack for some reason.

adammoody:
Hmm, well I pushed a change for this, but it's not coming through.  I also am having trouble adding new PRs.  I think my fork may have lost it's association with this repo somewhere in the change from private to public.

adammoody:
Reforked and resubmitted this PR

PR #36, opened 2017-12-15
Build test cases and gotcha, get make dist working

adammoody:
- build gotcha-enabled library
- build test cases
- support make dist
- fix a few bugs exposed during gotcha build


client/src/unifycr-sysio.c
```diff
@@ -358,6 +358,7 @@ int UNIFYCR_WRAP(__xstat)(int vers, const char *path, struct stat *buf)
     }
 }
 
+#if 0
```
nedbass: I suggest deleting these code blocks. Keeping them commented out hurts readability by adding clutter, and it's not necessary since it can always be retrieved from the git history if needed. If we plan to re-enable this later let's open a github issue to track that work.

client/src/unifycr-sysio.h
```diff
@@ -57,9 +57,9 @@ UNIFYCR_DECL(rename, int, (const char *oldpath, const char *newpath));
 UNIFYCR_DECL(truncate, int, (const char *path, off_t length));
 UNIFYCR_DECL(stat, int,( const char *path, struct stat *buf));
 UNIFYCR_DECL(__lxstat, int, (int vers, const char* path, struct stat *buf));
-UNIFYCR_DECL(__lxstat64, int, (int vers, const char* path, struct stat64 *buf));
+//UNIFYCR_DECL(__lxstat64, int, (int vers, const char* path, struct stat64 *buf));
```
nedbass: delete

client/src/unifycr-sysio.h
```diff
@@ -57,9 +57,9 @@ UNIFYCR_DECL(rename, int, (const char *oldpath, const char *newpath));
...
 UNIFYCR_DECL(__xstat, int, (int vers, const char* path, struct stat *buf));
-UNIFYCR_DECL(__xstat64, int, (int vers, const char* path, struct stat64 *buf));
+//UNIFYCR_DECL(__xstat64, int, (int vers, const char* path, struct stat64 *buf));
```
nedbass: delete

client/src/unifycr-sysio.h
```diff
@@ -89,7 +89,7 @@ UNIFYCR_DECL(mmap64, void*, (void *addr, size_t length, int prot, int flags, int
 UNIFYCR_DECL(munmap, int,(void *addr, size_t length));
 UNIFYCR_DECL(msync, int, (void *addr, size_t length, int flags));
 UNIFYCR_DECL(__fxstat, int, (int vers, int fd, struct stat *buf));
-UNIFYCR_DECL(__fxstat64, int, (int vers, int fd, struct stat64 *buf));
+//UNIFYCR_DECL(__fxstat64, int, (int vers, int fd, struct stat64 *buf));
```
nedbass: delete

configure.ac
```diff
@@ -146,6 +148,9 @@ PKG_CHECK_MODULES([LEVELDB], [leveldb], ,
         ],
         AC_MSG_ERROR(['leveldb is required to build unifycr.'])))
 
+# look for gotcha library, sets GOTCHA_INCLUDE, GOTCHA_LIB
+X_AC_GOTCHA
```
nedbass: It might be helpful to use a more descriptive prefix like UNIFYCR_ for our m4 macros to make internal ones easy to identify.

m4/lx_find_gotcha.m4
```diff
@@ -0,0 +1,18 @@
+AC_DEFUN([X_AC_GOTCHA], [
```
nedbass: I'd just name this file gotcha.m4, and call the macro UNIFYCR_AC_GOTCHA as mentioned above.

adammoody:
@nedbass, can you take a fresh look?

nedbass:
@adammoody thanks for refreshing. If you can it would be nice to clean up the minor style issues below. It would be also be good to clarify in the commit message if we are permanently dropping those stat wrappers or if we plan to eventually fix the build issue.

(I fought with that build issue for a bit but still don't quite understand how to fix it. It seems like it has to do `struct stat64` being an unknown type. Not sure if we're supposed to define some magic constant from features.h to enable large file support, but nothing I tried worked.)

```
WARNING: Possible unwrapped commit description (prefer a maximum 75 chars per line)
#6: 
- Add m4 script to find and link with gotcha.  Build and install libunifycr_gotcha library.

ERROR: "foo* bar" should be "foo *bar"
#318: FILE: client/src/gotcha_map_unifycr_list.h:43:
+UNIFYCR_DEF(setbuf, void, (FILE* stream, char* buf));

ERROR: "foo* bar" should be "foo *bar"
#353: FILE: client/src/unifycr-stdio.h:55:
+UNIFYCR_DECL(setbuf, void, (FILE *stream, char* buf));

WARNING: Avoid unnecessary line continuations
#410: FILE: client/src/unifycr-sysio.c:875:
+int UNIFYCR_WRAP(lio_listio)(int mode,\
```

adammoody:
Updated the comment.  I'll leave the style issues to the style PR that's next in line.

PR #37, opened 2017-12-15
Apply style changes to server and client code

adammoody:
Holds off on meta changes for now to keep the MDHIM code unchanged to enable easy patching.  I will cherry pick the non-MDHIM files from there later.

Dependent on build PR.  First couple of commits are from there.  This will be rebased on that later.


.travis.yml
```diff
@@ -22,4 +22,4 @@ script:
   - spack load environment-modules
   - source <(spack module loads gotcha leveldb)
   - sh autogen.sh && ./configure && make
-  - ./scripts/checkpatch.sh
+  # - ./scripts/checkpatch.sh
```
nedbass: Did you mean to comment this out? Or is this an artifact of exhausting our travis-ci.com trial builds?

adammoody: It's a temporary thing to get past the false positive it's tripping on in the WRAP macros.  I suspect we'll also want to leave it off for the style PR.  Wanted to get those two through as one-offs after which we can reenable it.

nedbass: Well but we can always merge a PR even if some aspects fail the style check. But I'd still prefer to see the output in case it finds valid stuff. I'd expect it to complain about the // comments, for example.

adammoody: back on

nedbass:
@adammoody or @kathrynmohror can you enable this repo on travis-ci.org now that it's public? I think travis-ci.com isn't tesing your latest commits since the repo visibility changed.

adammoody:
I think I got travis going again.  Had to delete the service and then set it up anew.  Probably lost our history in the process.

kathrynmohror:
Travis doesn't show this pull request in the interface. According to the directions, it might be because of a merge error: https://docs.travis-ci.com/user/pull-requests/#My-Pull-Request-isn%E2%80%99t-being-built

client/src/unifycr.c
```diff
@@ -1096,65 +1115,66 @@ int unifycr_fid_open(const char* path, int flags, mode_t mode, int* outfid, off_
...
+                /* initialize the global metadata
+                 * */
+                unifycr_filemeta_t *meta = unifycr_get_meta_from_fid(fid);
+                meta->real_size = ptr_meta->file_attr.st_size;
+                ptr_meta->fid = fid;
+                ptr_meta->gfid = gfid;
+
+                ins_file_meta(&unifycr_fattrs,
+                              ptr_meta);
+                free(ptr_meta);
```
nedbass: Not for this PR, but may want to check return statements above for possible memory leak.

client/src/unifycr.c
```diff
@@ -1578,67 +1599,67 @@ static void* unifycr_superblock_bgq(size_t size, const char* name)
...
-static int unifycr_abtoull(char* str, unsigned long long* val)
+static int unifycr_abtoull(char *str, unsigned long long *val)
 {
     /* check that we have a string */
     if (str == NULL) {
         debug("scr_abtoull: Can't convert NULL string to bytes @ %s:%d",
-            __FILE__, __LINE__
-        );
+              __FILE__, __LINE__
+             );
```
nedbass: Don't have to do it in this PR, but it would be nice to go back and move up all these closing parentheses on their own lines at some point.

nedbass:
@adammoody if you squash the patch stack that should get rid of a lot of the warnings and errors from the  style checker. As it is, each patch gets inspected, so style issues cleaned up in later patches still trigger warnings.

client/src/unifycr.c
```diff
@@ -737,11 +755,11 @@ int unifycr_fid_write(int fid, off_t pos, const void* buf, size_t count)
...
 
     /* get meta for this file id */
-    unifycr_filemeta_t* meta = unifycr_get_meta_from_fid(fid);
+    unifycr_filemeta_t *meta = unifycr_get_meta_from_fid(fid);
 
     /* determine storage type to write file data */
-    if (meta->storage == FILE_STORAGE_FIXED_CHUNK\
-    		|| meta->storage == FILE_STORAGE_LOGIO) {
+    if (meta->storage == FILE_STORAGE_FIXED_CHUNK
+        || meta->storage == FILE_STORAGE_LOGIO) {
```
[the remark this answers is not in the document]
adammoody: fixed

adammoody:
Seems like astyle gets confused one return (cast) -1; statements.  I've gone back and surrounded those with () and removed the extra space.

client/src/unifycr-sysio.c
```diff
@@ -491,30 +492,30 @@ int unifycr_fd_write(int fd, off_t pos, const void* buf, size_t count)
...
-			return UNIFYCR_ERR_IO;
-		}
     /* finally write specified data to file */
     int write_rc = unifycr_fid_write(fid, pos, buf, count);
-   
-	if (meta->storage == FILE_STORAGE_LOGIO) {
-			unifycr_filemeta_t *meta = unifycr_get_meta_from_fid(fid);
-			if (write_rc == 0)
-					meta->size = newpos;
-					meta->real_size = pos+count;
```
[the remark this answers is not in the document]
adammoody: Looks like a bug to me.  Nice catch!  I moved statement back inside the if block.  Reran the test cases, and those still complete ok.  Probably should have left that to a different PR, but it's fixed in this PR for now.

server/src/unifycr_request_manager.c
```diff
@@ -359,148 +366,156 @@ int rm_receive_remote_message(int app_id,\
...
-		int cli_id, int dbg_rank, int thrd_id, shm_meta_t *msg) {
-	LOG(LOG_DBG, "recv_msg:dbg_rank:%d, app_id:%d, cli_id:%d, thrd_id:%d,\
-			  fid:%d, offset:%ld, len:%ld\n", \
-			  dbg_rank, app_id, cli_id,  thrd_id, msg->src_fid,\
-				  	  msg->offset, msg->length);
+void print_recv_msg(int app_id,
+                    int cli_id, int dbg_rank, int thrd_id, shm_meta_t *msg)
+{
+    LOG(LOG_DBG, "recv_msg:dbg_rank:%d, app_id:%d, cli_id:%d, thrd_id:%d, \
+        fid:%d, offset:%ld, len:%ld\n",
```
nedbass: Unrelated to your changes, but this quoted string has a formatting error. Should remove line continuation and use two quoted strings.

client/src/unifycr-sysio.c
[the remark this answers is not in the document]
nedbass: Sounds good. I'd just add a note about the bug fix to the commit message.

nedbass:
Still lots more cleanup to do, but I think this is good for one pass. Most of the style checker warnings for this PR can be safely ignored, but it would be nice if we made another pass to fix all the 'space required here ...' and 'space prohibited here ...' class of style errors.

PR #51, opened 2017-12-21
Dead code removal in client/src

nedbass:
Remove a bunch of commented out code and address a few nearby code formatting issues.

I updated checkpatch.pl in this PR because it doesn't correctly support the indentation convention for this project and it was causing false warnings for the other patches.

Some functions in unifycr.c are now completely stubbed out. Rather than keeping the code commented out there, I propose we create github issues to track the need to implement (or possibly remove) the stubs.

### DIFF
--- a/client/src/unifycr-stdio.c
+++ b/client/src/unifycr-stdio.c
@@ -165,30 +165,6 @@ int unifycr_unsupported_stream(
     return rc;
 }
 
-#if 0
-static int unifycr_stream_alloc(int fd)
-{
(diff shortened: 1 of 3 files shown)


adammoody:
This generally looks fine to me.  One note about the memcopy.  We have a few variants of that around because it is handy for testing when saving data in shared memory.  We found that different memcopy functions performed quite differently depending on the architecture and compiler.  I think for testing I was lazy and just commented them in/out.  We should find a good way to do that if it still matters, e.g., compile time guards or something.

PR #61, opened 2017-12-23
Fix a slew of compiler warnings

nedbass:
See individual commit messages for details. These commits should *not* be squashed together if this PR is merged, in case we want to revert any individual changes.

This fixes all warnings presently showing up in the travis build. However more will show up if we enabled `-Wall` in the client and meta directories.


client/src/unifycr-sysio.c
```diff
@@ -828,8 +828,12 @@ ssize_t UNIFYCR_WRAP(write)(int fd, const void *buf, size_t count)
...
             }
         } else {
+            ssize_t rc;
+
             fd += unifycr_fd_limit;
-            pwrite(fd, buf, count, filedesc->pos);
+            rc = pwrite(fd, buf, count, filedesc->pos);
+            if (rc < 0)
+                perror("pwrite failed");
         }
```
adammoody: Here we should set errno appropriately and return (ssize_t)-1 to the user and let them handle it, similar to what we do higher up in this routine:

            /* ERROR: invalid file descriptor */
            errno = EBADF;
            return (ssize_t) (-1);

Reading the man page for pwrite, it sets errno, so we just need to return (ssize_t) -1 and perhaps make a comment that pwrite sets errno for us.

adammoody:
That's a lot of good clean up.  Thanks, @nedbass.

client/src/unifycr-sysio.c
nedbass: Sounds good. I'll also update it to handle the short-write case where pwrite() returns < count.

client/check_fns/unifycr_translate.py
```diff
@@ -47,17 +47,17 @@ def write_gotcha_file(filename, func_list, modulename):
...
-	f.write("\n#define NFUNCS " + str(len(func_list)) + "\n\n")
 
 	# write map struct
 	f.write("struct gotcha_binding_t wrap_" + modulename + "[] = {\n")
 	for function in func_list:
 		f.write("\t{ " + function.name_quote + ", " + function.wrapper + ", &" + function.wrapee + " },\n")
 	f.write("};\n")
 
-        # close struct and file
+	# define number of functions we are wrapping for gotcha struct
```
dsikich: @nedbass thanks for adding this. I was thinking about if it would be good to generate the header at compile time with this script based on the unifycr_list, and having the header generated be part of build output? That way, people will not mistakenly update the header and not the list? And people wouldn't have to generate the header manually?

nedbass: @dsikich I had the same thought while I was working on this patch. But I hesitate because I don't fully understand why this approach of generating the header was chosen. My concern is that it adds complexity to the build process, and I suspect it may be preferable to just ditch the translate script and maintain the header "by hand".

nedbass: But if we *do* decide that translation script is the right approach, then I absolutely agree it should be incorporated in the build system, and the header should be removed from revision control since it is technically a "build product".

dsikich: @nedbass @adammoody I generated the header originally when I was integrating the code with gotcha, which had a lot of extra code that needed to be generated for each function we are trying to wrap. It needed the struct, the number of functions, a wrapped prototype, and a pointer to the original function declared for each wrapped function. Since there were so many functions being wrapped I just generated all of the prototypes, pointer declarations, and gotcha struct. 

Now that the majority of the code for that is generated, I don't think it would be too hard to maintain the header by hand. But, each time a new function to be wrapped is added all three of those things will need to be added and/or updated (the prototype for the wrapped function, the declaration of the pointer to the function, and the struct entry). I don't know what would be better and/or easier honestly. I prefer whichever way makes it easiest to add new functions to be wrapped.

nedbass: I guess my preference would be to keep the header  in revision control and maintain it by hand so that developers can find everything using tools like 'git grep', ctags, etc., even if they didn't happen to run 'make' yet. But we should have a tool that will warn if any of the required elements for a wrapped function are missing.

dsikich: @nedbass Sounds good to me.

PR #78, opened 2018-01-05
fix debug output in client

dsikich:
The client was printing debug information about gotcha wrappers by
default.

        - update unifycr-defs so it defaults to not printing debug info
        - update unifycr-internal so that debug function prints file and
          line info
        - change unifycr.c to use debug function instead of printf

### DIFF
--- a/client/src/unifycr-defs.h
+++ b/client/src/unifycr-defs.h
@@ -1,8 +1,15 @@
+/* holds debug level for unifycr, this defaults to zero
+ * in unifycr_init, but can be changed at runtime by setting
+ * the UNIFYCR_DEBUG env variable to a value greater than 1
+ */
+extern int unifycr_debug_level;
+
 #define UNIFYCR_MAX_FILES        ( 128 )
 
 /* eventually could decouple these so there could be
  * more or less file descriptors than files, but for
- * now they're the same */
+ * now they're the same
+ */
 #define UNIFYCR_MAX_FILEDESCS    ( UNIFYCR_MAX_FILES )
 
 #define UNIFYCR_MAX_FILENAME     ( 128 )
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -176,6 +176,9 @@ extern pthread_mutex_t unifycr_stack_mutex;
 /* keep track of what we've initialized */
 int unifycr_initialized = 0;
 
+/* keep track of debug level */
+int unifycr_debug_level;
+
 /* global persistent memory block (metadata + data) */
 void *unifycr_superblock = NULL;
 static void *free_fid_stack = NULL;
@@ -279,9 +282,46 @@ int unifycr_err_map_to_errno(int rc)
         return EBADF;
     case UNIFYCR_ERR_ISDIR:
         return EISDIR;
-    default:
-        return EIO;
+    case UNIFYCR_ERR_NOMEM:
+        return ENOMEM;
     }
+    return EIO;
+}
+
+/* given an errno error code, return corresponding UnifyCR error code */
+int unifycr_errno_map_to_err(int rc)
+{
+    switch (rc) {
+    case 0:
+        return UNIFYCR_SUCCESS;
+    case ENOSPC:
+        return UNIFYCR_ERR_NOSPC;
+    case EIO:
+        return UNIFYCR_ERR_IO;
+    case ENAMETOOLONG:
+        return UNIFYCR_ERR_NAMETOOLONG;
+    case ENOENT:
+        return UNIFYCR_ERR_NOENT;
+    case EEXIST:
+        return UNIFYCR_ERR_EXIST;
+    case ENOTDIR:
+        return UNIFYCR_ERR_NOTDIR;
+    case ENFILE:
+        return UNIFYCR_ERR_NFILE;
+    case EINVAL:
+        return UNIFYCR_ERR_INVAL;
+    case EOVERFLOW:
+        return UNIFYCR_ERR_OVERFLOW;
+    case EFBIG:
+        return UNIFYCR_ERR_FBIG;
+    case EBADF:
+        return UNIFYCR_ERR_BADF;
+    case EISDIR:
+        return UNIFYCR_ERR_ISDIR;
+    case ENOMEM:
+        return UNIFYCR_ERR_NOMEM;
+    }
+    return UNIFYCR_FAILURE;
 }
 
 /* returns 1 if two input parameters will overflow their type when
@@ -419,7 +459,7 @@ inline int unifycr_intercept_fd(int *fd)
          * so intercept the call and shift the fd */
         int newfd = oldfd - unifycr_fd_limit;
         *fd = newfd;
-        debug("Changing fd from exposed %d to internal %d\n", oldfd, newfd);
+        DEBUG("Changing fd from exposed %d to internal %d\n", oldfd, newfd);
         return 1;
     }
 }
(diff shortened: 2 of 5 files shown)


nedbass:
It may be useful to be able to set the debug level at runtime with an environment variable rather than at build time.

dsikich:
@nedbass I'm also getting  a warning about a line over 80 characters (the debug print macro) in unifier-internal.h. I was using a "\" character to continue the line before it hit 80 characters, but it looks like it is reading it as one line. Is there a better way to do that and/or get around that?

dsikich:
@nedbass oh, yeah that would be easier. I will update with a check for the env variable at runtime instead.

nedbass:
I often use a macro something like this, where `unifycr_debug_level` would be set in `unifycr_init()`. The "unifycr:" prefix might help distinguish these debug messages from the application's output.

```
#define DEBUG(fmt, ...)                                        \
do {                                                           \
    if (unifycr_debug_level > 0)                                       \
        printf("unifycr: %s:%d: %s: " fmt "\n",                \
               __FILE__, __LINE__, __func__, ## __VA_ARGS__);  \
} while (0)
```

dsikich:
@nedbass I'm not really sure about the warnings I am getting. When I continue a string on another line (to not go over 80 characters) it still says it is over 80 characters. If I use a \ character it also warns about that? I'm not sure. I've tried many variations, but I'm not seeing how to get the warning to go away.

dsikich:
@nedbass okay, tests are passing. I had to fix my ~/.vimrc config.

client/src/unifycr.c
```diff
@@ -2770,7 +2789,7 @@ chunk_list_t *unifycr_get_chunk_list(char *path)
     return NULL;
 }
 
-/* debug function to print list of chunks constituting a file
+/* DEBUG function to print list of chunks constituting a file
```
nedbass: same here

client/src/unifycr.c
```diff
@@ -2218,7 +2236,7 @@ static int unifycr_sync_to_del()
     memcpy(cmd_buf + 2 * sizeof(int),
            &local_rank_idx, sizeof(int));
     memcpy(cmd_buf + 3 * sizeof(int),
-           &dbg_rank, sizeof(int)); /*add debug info*/
+           &dbg_rank, sizeof(int)); /*add DEBUG info*/
```
nedbass: I suspect this was an inadvertent search and replace hit (not that it hurts anything :)

client/src/unifycr-defs.h
```diff
@@ -1,8 +1,15 @@
+/* holds debug level for unifycr, this defaults to zero
+ * in unifycr_init, but can be changed at runtime by setting
+ * the DEBUG env variable to a value greater than 1
```
nedbass: I'd prefer to name it `UNIFYCR_DEBUG` for consistency with other unifycr environment variables and  to reduce likelihood of name collision.

dsikich:
@nedbass Thanks for your help. No worries, this will be very useful!

client/src/unifycr.c
```diff
@@ -1679,21 +1686,30 @@ static int unifycr_abtoull(char *str, unsigned long long *val)
 
 static int unifycr_init(int rank)
 {
-    if (! unifycr_initialized) {
+    if (!unifycr_initialized) {
+
+		/* unifycr DEBUG level default is zero */
+        unifycr_debug_level = 0;
```
nedbass: Indentation looks off here. Probably from mixing tabs and spaces.

I use this in my .vimrc to make that easier to spot:

```
set list            " * Turn on printing of invisible chars
set listchars=tab:>-,trail:! " * Customize the characters used to print the
                    "   invisible chars
```

client/src/unifycr.c
```diff
@@ -1679,21 +1686,30 @@ static int unifycr_abtoull(char *str, unsigned long long *val)
...
 
         int i;
+
         for (i = 0; i < GOTCHA_NFUNCS; i++) {
             if (*(void **)(wrap_unifycr_list[i].function_address_pointer) == 0) {
-                printf("This function name failed to be wrapped: %s\n",
-                       wrap_unifycr_list[i].name);
+                DEBUG("This function name failed to be wrapped: %s\n",
+                      wrap_unifycr_list[i].name
+                      );
```
nedbass: (style) move closing paren to line above

client/src/unifycr.c
```diff
@@ -2138,19 +2154,21 @@ int unifycrfs_mount(const char prefix[], size_t size, int rank)
...
+            DEBUG("rank:%d, fail to init shared receive memory.", dbg_rank);
             return UNIFYCR_FAILURE;
         }
 
         rc = unifycr_sync_to_del();
         if (rc < 0) {
-            debug("rank:%d, fail to convey information to the delegator.", dbg_rank);
+            DEBUG("rank:%d, fail to convey information to the delegator.",
+				  dbg_rank
+				  );
```
nedbass: (style) move closing paren to line above

client/src/unifycr.c
```diff
@@ -665,7 +668,9 @@ int unifycr_fid_create_file(const char *path)
      * and return appropriate error if it is greater
      */
     strcpy((void *)&unifycr_filelist[fid].filename, path);
-    debug("Filename %s got unifycr fd %d\n", unifycr_filelist[fid].filename, fid);
+    DEBUG("Filename %s got unifycr fd %d\n",
+		  unifycr_filelist[fid].filename, fid
+		  );
```
nedbass: (style) move closing paren to line above

client/src/unifycr.c
```diff
@@ -1148,23 +1153,24 @@ int unifycr_fid_open(const char *path, int flags, mode_t mode, int *outfid,
...
-            debug("unifycr_superblock = %p; free_fid_stack = %p; free_chunk_stack = %p; unifycr_filelist = %p; chunks = %p\n",
-                  unifycr_superblock, free_fid_stack, free_chunk_stack, unifycr_filelist,
-                  unifycr_chunks
-                 );
+            DEBUG("Couldn't find entry for %s in UNIFYCR\n", path);
+            DEBUG("unifycr_superblock = %p; free_fid_stack = %p;"
+                  "free_chunk_stack = %p; unifycr_filelist = %p; chunks = %p\n",
+                  unifycr_superblock, free_fid_stack, free_chunk_stack,
+                  unifycr_filelist, unifycr_chunks
+                  );
```
nedbass: (style) move closing paren to line above

client/src/unifycr.c
dsikich: @nedbass yes, thanks. I can see the tabs now!

client/src/unifycr.c
```diff
@@ -2239,8 +2244,9 @@ static int unifycr_sync_to_del()
     memcpy(cmd_buf + 7 * sizeof(int) + 6 * sizeof(long),
            &data_size, sizeof(long));
 
+	/*adjust to add debug info*/
```
nedbass: check indentation

dsikich:
@nedbass the newly merged pull requests conflict with a file now. I will submit a new PR off of an updated dev branch.

nedbass:
> @nedbass the newly merged pull requests conflict with a file now. I will submit a new PR off of an updated dev branch.

Sounds good. Just FYI, you can always force update your branch after a rebase, it's not necessary to open a new PR.

dsikich:
Okay, thanks good to know. In this case though I’ve already submitted a new one (#82).

PR #88, opened 2018-01-25
fix hard coding of external meta and data dirs

dsikich:
The external data and meta directories should be set based on the
path to the external SSD's on that system. The path was using a
default (/l/ssd/), which only exists on a particular cluster.

   - The client now fails when the external meta and data directory
     are not set

   - To run the client successfully the UNIFYCR_EXTERNAL_META_DIR
     and the UNIFYCR_EXTERNAL_DATA_DIR need to be set to valid paths

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -1931,7 +1931,9 @@ static int unifycr_init(int rank)
         if (env) {
             strcpy(external_data_dir, env);
         } else {
-            strcpy(external_data_dir, EXTERNAL_DATA_DIR);
+            DEBUG("UNIFYCR_EXTERNAL_DATA_DIR not set to an existing writable"
+                  " path (i.e UNIFYCR_EXTERNAL_DATA_DIR=/mnt/ssd):\n");
+            return UNIFYCR_FAILURE;
         }
 
         sprintf(spillfile_prefix, "%s/spill_%d_%d.log",
@@ -1953,7 +1955,9 @@ static int unifycr_init(int rank)
         if (env) {
             strcpy(external_meta_dir, env);
         } else {
-            strcpy(external_meta_dir, EXTERNAL_META_DIR);
+            DEBUG("UNIFYCR_EXTERNAL_META_DIR not set to an existing writable"
+                  " path (i.e UNIFYCR_EXTERNAL_META_DIR=/mnt/ssd):\n");
+            return UNIFYCR_FAILURE;
         }
 
         /*ToDo: add the spillover feature for the index metadata*/
@@ -2643,7 +2647,10 @@ int unifycrfs_mount(const char prefix[], size_t size, int rank)
     }
 
     /* initialize our library */
-    unifycr_init(rank);
+    int ret = unifycr_init(rank);
+
+    if (ret != UNIFYCR_SUCCESS)
+        return ret;
 
     if (fs_type == UNIFYCR_LOG || fs_type == UNIFYCR_STRIPE) {
         char host_name[UNIFYCR_MAX_FILENAME] = {0};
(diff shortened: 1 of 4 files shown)


dsikich:
@adammoody thanks, I realized I probably could have moved the environment check to one if/else block? But I initially just added on (or removed) from what was already there.

client/src/unifycr.c
```diff
@@ -1931,7 +1931,10 @@ static int unifycr_init(int rank)
         if (env) {
             strcpy(external_data_dir, env);
         } else {
-            strcpy(external_data_dir, EXTERNAL_DATA_DIR);
+            fprintf(stderr, "UNIFYCR_EXTERNAL_DATA_DIR not set to an SSD path "
+                            "(i.e UNIFYCR_EXTERNAL_DATA_DIR=/mount/ssd):%s : %d\n",
+                            __FILE__, __LINE__);
+            exit(UNIFYCR_FAILURE);
```
nedbass: Rather than calling exit() here, what would you think about propagating an error back so unifycr_mount() fails? Shared libraries shouldn't normally call exit().

client/src/unifycr.c
```diff
@@ -1931,7 +1931,10 @@ static int unifycr_init(int rank)
         if (env) {
             strcpy(external_data_dir, env);
         } else {
-            strcpy(external_data_dir, EXTERNAL_DATA_DIR);
+            fprintf(stderr, "UNIFYCR_EXTERNAL_DATA_DIR not set to an SSD path "
+                            "(i.e UNIFYCR_EXTERNAL_DATA_DIR=/mount/ssd):%s : %d\n",
+                            __FILE__, __LINE__);
```
nedbass: * This should probably use the DEBUG macro rather than print directly to the console.
* The path doesn't technically *have* to be on a SSD, so I might word this like "to an existing writable directory"
* I'd suggest /mnt instead of /mount for the example path to be more conventional

client/src/unifycr.c
dsikich: @nedbass maybe returning UNIFYCR_FAILURE? It looks like that is what the other calls are doing in unifycrfs_mount?

client/src/unifycr.c
dsikich: @nedbass Okay, I had been using the DEBUG macro originally, but wasn't sure if it should show if DEBUG wasn't turned on. I will update.

client/src/unifycr.c
nedbass: Yes, I agree.

Longer term I think we should design a more robust way to return error information to the application, since there's a variety of reasons why `unifycr_mount()` might fail. A simple option might be to set errno, and provide documentation on how to interpret the error codes. We might want to think about some auxiliary interface that could return diagnostic messages that the application could print.

client/src/unifycr.c
dsikich: @nedbass There is a unifycr map to error numbers in unifycr.c of some kind it looks like? But they are only sometimes used. Also, I updated with the change to DEBUG and removed the reference to SSD.

client/src/unifycr.c
```diff
@@ -2643,7 +2647,10 @@ int unifycrfs_mount(const char prefix[], size_t size, int rank)
     }
 
     /* initialize our library */
-    unifycr_init(rank);
+    int ret = unifycr_init(rank);
+
+    if (ret != UNIFYCR_SUCCESS)
+        return UNIFYCR_FAILURE;
```
nedbass: Small nit here, but I'd suggest returning `ret` here. My thinking is that in the future `unificy_init()` might return different error codes depending on the failure type, and we'd want to pass that back to the application.

client/src/unifycr.c
nedbass: @dsikich thanks. I think error mapping and reporting would be a good area for us to revisit in a new  issue.

client/src/unifycr.c
dsikich: @nedbass Thanks, good catch. I updated it.

PR #90, opened 2018-02-01
fix spack instruction comments in docs

dsikich:
I noticed I forgot to take out extra comments that didn't make sense when I added more information about spack module commands

### DIFF
--- a/docs/build-intercept.rst
+++ b/docs/build-intercept.rst
@@ -16,9 +16,7 @@ LevelDB, and GOTCHA.
 ***************************
 
 To install leveldb and gotcha and set up your build environment, we recommend
-using the `Spack package manager <https://github.com/spack/spack>`_. For example,
-the following commands will build and install leveldb and gotcha using Spack and
-configure your environment so UnifyCR can find the libraries and headers it needs:
+using the `Spack package manager <https://github.com/spack/spack>`_.
 
 The instructions assume that you do not already have a module system installed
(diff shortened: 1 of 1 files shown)


nedbass:
This looks good. Regarding the example commands, I think the `. spack/share/spack/setup-env.sh` step sets $PATH, so can we put that before the `spack` commands and skip the steps that set $PATH directly?

dsikich:
@nedbass When I setup my spack environment I wasn't able to run the load command unless I used this sequence: install the environment-modules then use the setup-env script. I suppose I just added spack to the path first to be able to just type "spack install..". Were you able to run "spack load..." using that sequence of commands?

dsikich:
@nedbass I will also forward you the email thread from a graduate student using UnifyCR who had trouble with the old sequence and installing environment-modules as well.

nedbass:
> @nedbass When I setup my spack environment I wasn't able to run the load command unless I used this sequence: install the environment-modules then use the setup-env script.

Yes I seem to recall that being the case as well. That's why our [travis.yml](https://github.com/LLNL/UnifyCR/blob/dev/.travis.yml) does it in that order. I just forgot since environment-modules is already installed in my development environment. In that it might be cleaner just to say ./spack/bin/spack in the example so all the commands will work when cut-and-pasted.

dsikich:
@nedbass Oh, yeah ok that makes sense. I updated the commands.

PR #96, opened 2018-02-21
add spillover size calculation fix from HPE

dsikich:
This will add the spillover size calculation fix from HPE.


nedbass:
I was going to recommend attributing these HPE patches to their original authors, but I'm too late :smile:

dsikich:
@nedbass yeah I thought about that, but I'm not really sure who actually wrote the the code? So I just put HPE.

adammoody:
Good point, @nedbass.  We'll try to track down author names if we can for future patches that come in this way.

PR #97, opened 2018-02-22
fix wrong return code of fsync from Teng Wang

dsikich:
Trying to get UnifyCR up to date with Teng's bug fixes in burstfs here: https://github.com/LLNL/burstfs/commit/2edb8ad73f8c95659a89130cada6da82cc69d3ca

### DIFF
--- a/client/src/unifycr-sysio.c
+++ b/client/src/unifycr-sysio.c
@@ -1561,8 +1561,8 @@ int UNIFYCR_WRAP(fsync)(int fd)
                                 return -1;
                             } else {
                                 /**/
-                                if (ack_msg[0] != COMM_META ||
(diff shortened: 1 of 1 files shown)


nedbass:
For proper attribution you can use `git commit --author "Teng Wang <EMAIL ADDRESS>"`.

It would be good to reference the burstfs commit in the commit message.  It would also be nice to include some details in the commit message about the bug being fixed, although I know you didn't write the patch so may not have that information at hand.

dsikich:
@nedbass thanks, I will update the commit message and author. But yeah, there isn't much to go on in terms of info about the patch itself.

dsikich:
@nedbass okay, the build was failing because I was using the whole commit has in the message which was too long :-)

PR #99, opened 2018-02-22
fix non-sequential arraylist insertion

dsikich:
This handles the case when the insertion to the arraylist does not
follow the sequential order.

This patch is from burstfs here:
https://github.com/LLNL/burstfs/commit/46dd467


dsikich:
@nedbass @adammoody is this PR ok? Do I need to make some changes?

PR #100, opened 2018-02-26
update build docs

dsikich:
An example of a leveldb pkgconfig file was added to the 'building without
spack' section of the documentation. The change in the configure line was
also documented. This problem arises when someone uses a leveldb build
from source and is not using spack to build their dependencies.


dsikich:
@nedbass @adammoody @kathrynmohror Hi, is this PR ok? If so, I'd just like to get it into the docs.

Issue #86, opened 2018-01-17
Resource manager integration


sandrain:
This issue is to develop additional software components that allows UnifyCR to interact with cluster resource managers such as SLURM and LSF. Such interactions are necessary to launch and terminate the UnifyCR daemon with user-specified configurations. In addition, 'unifycr' command line utility will be developed for support environments without having any resource managers.

### Expected functionalities

1. Reading the system-wide options if any, e.g., shared memory size.
2. Recognizing the user-specified configuration options, which will be fed to the UnifyCR daemon.
3. Launching the UnifyCR daemon across a job allocation, i.e., a set of compute node, with the given options
4. Performing any initialization and cleanup tasks, e.g., preloading/draining data files from/to PFS

### User-specified options
The following is the initial list of the options that will be supported:

1. ```mount=<name>```: Specifies the mount point name to be used by UnifyCR. Job environment variable
UNIFYCR MT is set to the value of name.
2. ```transfer in=<path or filename>```: Specifies a file name or directory to transfer into UnifyCR instance at the beginning of the job.
3. ```transfer out=<path or filename>```: Specifies a file name or directory to transfer files into from UnifyCR instance at the end of the allocation.
4. ```cleanup```: Specifies if UnifyCR should clean up any storage used at the end of the allocation.
5. ```consistency model```: Planned option to allow user to specify desired consistency model of those supported by UnifyCR. The default is “laminated consistency.”

### Implementation
Currently, three software components are planned:

1. unifycr command line utility.
2. Resource manager-specific hooks, e.g., SLURM plugin, LSF prolog/epilog scripts.
3. Initial template for configuration files, e.g., /etc/unifycr.conf

### New source code directories
The following directories will be appended to the existing source code tree:

1. ```/etc```: configuration file template
2. ```/rm```: resource manager-specific plugins or scripts
3. ```/util```: command-line tools including the ```unifycr``` utility

### Resources
1. [The planned design](https://github.com/LLNL/UnifyCR/files/1640729/4.1.resourcemanager-unifycr.txt)
2. [SLURM SPANK plugins](https://slurm.schedmd.com/spank.html)
3. [SLURM SPANK plugin examples 1](https://github.com/grondo/slurm-spank-plugins)
4. [SLURM SPANK plugin examples 1](https://lc.llnl.gov/bitbucket/users/sikich1/repos/slurm-lua-plugin/browse)
[...]

PR #103, opened 2018-03-01
unifycr command line tool and sysconf file implementation.

sandrain:
### Description
This includes a basic implementation of the unifier command line utility, addressing issue #86 .

### Motivation and Context
See issue #86.

### Current status
It successfully read configurations from sysconf file (/etc/unifycr/unifycr.conf), environmental variables, and command line arguments. Also, it generates the dynamic configuration file under specified path (runstatedir/unifycr/unifycr-runtime.conf, which can be changed by sysconf file). It is not launching the daemon, yet. In specific, the following is the remaining work list:

1. Launching the server daemon (do we need to execute mpirun and daemonize?)
2. Node list parsing using scontrol (SLURM), or manually.
3. Plugins for individual resource managers.
4. Changes to the unifycrd (server daemon) to read configurations from the dynamically created runtime configuration file.

The test program is going to be included after the item 4. above (changing the unifycrd).

### Types of changes

### Checklist:

### DIFF
--- a/util/unifycr/src/unifycr-cons.c
+++ b/util/unifycr/src/unifycr-cons.c
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2017, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyCR.
+ * For details, see https://github.com/LLNL/UnifyCR.
+ * Please read https://github.com/LLNL/UnifyCR/LICENSE for full license text.
+ */
+
+/*
+ * Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ * Copyright (c) 2017, Florida State University. Contributions from
+ * the Computer Architecture and Systems Research Laboratory (CASTL)
+ * at the Department of Computer Science.
+ *
+ * Written by: Hyogi Sim
+ * LLNL-CODE-728877. All rights reserved.
+ *
+ * This file is part of burstfs.
+ * For details, see https://github.com/llnl/burstfs
+ * Please read https://github.com/llnl/burstfs/LICENSE for full license text.
+ */
+
+/*
+ *
+ * Copyright (c) 2014, Los Alamos National Laboratory
+ *	All rights reserved.
+ *
+ */
+
+#ifndef _CONFIG_H
+#define _CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "unifycr.h"
+
+/**
+ * @brief this should match with the enum definition of unifycr_cm_t in
+ * unifycr.h
+ */
+static char *consistency_strs[N_UNIFYCR_CM] = { "none", "laminated", "posix" };
+
+/**
+ * @brief return consistency enum entry matching with the given string.
+ *
+ * @param str input string of the consistency model
+ *
+ * @return matching enum entry on success, UNIFYCR_CM_INVALID (-1) otherwise
+ */
+unifycr_cm_t unifycr_read_consistency(const char *str)
+{
+    int i = 0;
+
+    for (i = 0; i < N_UNIFYCR_CM; i++)
+        if (!strcmp(str, consistency_strs[i]))
+            return i;
+
+    return UNIFYCR_CM_INVALID;
+}
+
+/**
+ * @brief return the consistency string based on the given code.
+ *
+ * @param con the consistency code (integer)
+ *
+ * @return string that describes the consistency model
+ */
+const char *unifycr_write_consistency(unifycr_cm_t con)
+{
+    if (con < 0 || con >= N_UNIFYCR_CM)
+        return "invalid";
+    else
+        return consistency_strs[con];
+}
+
+
--- a/util/unifycr/src/unifycr-rm.c
+++ b/util/unifycr/src/unifycr-rm.c
@@ -1,0 +1,198 @@
+/*
+ * Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2017, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyCR.
+ * For details, see https://github.com/LLNL/UnifyCR.
+ * Please read https://github.com/LLNL/UnifyCR/LICENSE for full license text.
+ */
+
+/*
+ * Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ * Copyright (c) 2017, Florida State University. Contributions from
+ * the Computer Architecture and Systems Research Laboratory (CASTL)
+ * at the Department of Computer Science.
+ *
+ * Written by: Hyogi Sim
+ * LLNL-CODE-728877. All rights reserved.
+ *
+ * This file is part of burstfs.
+ * For details, see https://github.com/llnl/burstfs
+ * Please read https://github.com/llnl/burstfs/LICENSE for full license text.
+ */
+
+/*
+ *
+ * Copyright (c) 2014, Los Alamos National Laboratory
+ *	All rights reserved.
+ *
+ */
+
+#ifndef _CONFIG_H
+#define _CONFIG_H
+#include <config.h>
+#endif
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <ctype.h>
+#include <errno.h>
+
+#include "unifycr.h"
+
+/**
+ * @brief get the node list from the $PBS_NODEFILE
+ *
+ * @param resource
+ *
+ * @return
+ */
+static int read_resource_pbs(unifycr_resource_t *resource)
+{
+    int ret = 0;
+    int i = 0;
+    int n_nodes = 0;
+    FILE *fp = NULL;
+    char **nodes = NULL;
+    char buf[1024] = { 0, };
+    char *nodefile = getenv("PBS_NODEFILE");
+
+    if (!nodefile)
+        return -EINVAL;
+
+    fp = fopen(nodefile, "r");
+    if (!fp)
+        return -errno;
+
+    while (fgets(buf, 1023, fp))
+        n_nodes++;
+
+    if (ferror(fp)) {
+        ret = -errno;
+        goto out;
+    }
+
+    nodes = calloc(sizeof(char *), n_nodes);
+    if (!nodes) {
+        ret = -errno;
+        goto out;
+    }
+
+    rewind(fp);
+
+    while (fgets(buf, 1024, fp) != NULL) {
+        buf[strlen(buf) - 1] = '\0';    /* discard the newline */
+        nodes[i] = strdup(buf);
+        i++;
+    }
+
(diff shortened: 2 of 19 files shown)


sandrain:
I am checking the build log, and it seems like checkpatch.sh fails due to the warnings.

dsikich:
@sandrain something that helped me was running the scripts/checkpatch.sh on my latest commit before I push. If that passes, then the Travis build should also pass when you push to the remote repo.

sandrain:
@dsikich yes, I did that this time. I didn't fixed some warnings, because I think using typedef is not unreasonable. I suspect that that requirement is for linux kernel? Anyway, thanks. I can fix all warnings but am not quite sure whether it is necessary.

nedbass:
We should disable the typedef checks. I'll look into that tomorrow. We're still fine tuning the style checker to eliminate false positives like that.

On Feb 28, 2018, at 19:17, Hyogi Sim <notifications@github.com<mailto:notifications@github.com>> wrote:


@dsikich<https://github.com/dsikich> yes, I did that this time. I didn't fixed some warnings, because I think using typedef is not unreasonable. I suspect that that requirement is for linux kernel? Anyway, thanks. I can fix all warnings but am not quite sure whether it is necessary.

nedbass:
@sandrain you can add this diff to your patch to quiet the Linux-specific warnings. That leaves just a few that remaining warnings that should be fixed.

```diff
diff --git a/scripts/checkpatch.sh b/scripts/checkpatch.sh
index 2f08c2f..a9c526a 100755
--- a/scripts/checkpatch.sh
+++ b/scripts/checkpatch.sh
@@ -23,6 +23,8 @@ checkpatch_ignore+=",CONST_STRUCT"      # Don't nag about const structs
 checkpatch_ignore+=",SPLIT_STRING"      # Allow long strings to be split
 checkpatch_ignore+=",ARRAY_SIZE"        # Don't require use of ARRAY_SIZE macro
 checkpatch_ignore+=",USE_NEGATIVE_ERRNO" # We don't return negative errnos
+checkpatch_ignore+=",NEW_TYPEDEFS"
+checkpatch_ignore+=",ENOSYS"

 checkpatch_cmd+=" --ignore $checkpatch_ignore"
```

sandrain:
@nedbass so travis pulls different commit (be671a9), rather than my latest above (a49fe3e). Is it because I use --amend locally to make up a single commit out of two?

nedbass:
> @nedbass so travis pulls different commit (be671a9), rather than my latest above (a49fe3e). Is it because I use --amend locally to make up a single commit out of two?

No, it's because travis fetches it as a merge commit (`git fetch origin +refs/pull/102/merge: ; git checkout FETCH_HEAD`). The style check fails if any patch in your PR has style issues. So having follow up patches to address the style issues doesn't make travis happy.  The typical workflow is to squash those clean up fixes into the first commit and force update your branch. No need to open a new PR.

util/unifycr/src/unifycr-cons.c
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * @return matching enum entry on success, UNIFYCR_CM_INVALID (-1) otherwise
+ */
+unifycr_cm_t unifycr_read_consistency(const char *str)
+{
+    int i = 0;
+    unifycr_cm_t cons = UNIFYCR_CM_INVALID;
+
+    for (i = 0; i < sizeof(consistency_strs)/sizeof(char *); i++)
+        if (!strncmp(str, consistency_strs[i], strlen(consistency_strs[i])))
```
nedbass: Strictly speaking, this should probably use strcmp() to ensure an exact match, otherwise you could get unexpected results for strings that share a common prefix.

util/unifycr/src/unifycr-cons.c
```diff
@@ -0,0 +1,85 @@
...
+            cons = i;
```
nedbass: (nit) could return on first match

util/unifycr/src/unifycr-cons.c
```diff
@@ -0,0 +1,85 @@
...
+/**
+ * @brief return the consistency string based on the given code.
+ *
+ * @param con the consistency code (integer)
+ *
+ * @return string that describes the consistency model
+ */
+char *unifycr_write_consistency(unifycr_cm_t con)
+{
+    return con == UNIFYCR_CM_INVALID ? "invalid" : consistency_strs[con];
```
nedbass: Should check if con is outside of array boundaries.

util/unifycr/src/unifycr-rm.c
```diff
@@ -0,0 +1,201 @@
...
+
+    nodes = calloc(sizeof(char *), n_nodes);
+    if (!nodes) {
+        ret = errno;
+        goto out;
+    }
+
+    rewind(fp);
+
+    while (fgets(buf, 1023, fp) != NULL) {
```
nedbass: This should be sizeof(buf) for the size parameter.  fgets() always returns a null terminated string on success, so the next line is not needed.

util/unifycr/src/unifycr-rm.c
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    str = getenv("LSB_JOBID");
+    if (str) {
+        ret = read_resource_lsf(resource);
+        goto out;
+    }
+
+    fprintf(stderr, "cannot detect resource manager. "
+                    "please provide node info with -n option\n");
```
nedbass: I think there is currently no such -n option.

util/unifycr/src/unifycr.c
```diff
@@ -0,0 +1,269 @@
...
+
+    program = basename(argv[0]);
+
+    if (argc < 2)
+        usage(1);
+
+    cmd = argv[1];
+
+    for (i = 0; i < sizeof(actions)/sizeof(char *); i++) {
+        if (strncmp(cmd, actions[i], strlen(actions[i])) == 0)
```
nedbass: Here again I'd prefer strcmp() to ensure an exact match.

util/unifycr/src/unifycr.c
```diff
@@ -0,0 +1,269 @@
...
+    char *cmd = NULL;
+
+    program = basename(argv[0]);
+
+    if (argc < 2)
+        usage(1);
+
+    cmd = argv[1];
+
+    for (i = 0; i < sizeof(actions)/sizeof(char *); i++) {
```
nedbass: `sizeof(actions)/sizeof(actions[0])` would be a more general construct.

util/unifycr/src/unifycr.c
```diff
@@ -0,0 +1,269 @@
...
+    if (debug) {
+        int tmp;
+        char cmdbuf[1024] = { 0, };
+
+        printf("\n## the final runstate file %s/unifycr-runstate.conf ##\n",
+               opts_sysconf.runstatedir);
+
+        sprintf(cmdbuf, "cat %s/unifycr-runstate.conf",
+                        opts_sysconf.runstatedir);
+        tmp = system(cmdbuf);
```
nedbass: I think it would be better to open and print the file contents directly, rather than calling `cat`.

util/unifycr/src/unifycr.c
```diff
@@ -0,0 +1,269 @@
...
+
+    ret = unifycr_read_resource(&resource);
+    if (ret) {
+        fprintf(stderr, "failed to detect a resource manager\n");
+        goto out;
+    }
+
+    if (debug) {
+        printf("\n## job allocation (%llu nodes) ##\n",
+               _llu(resource.n_nodes));
```
nedbass: For code readability I'd prefer an explicit typecast here, rather than a custom macro.

util/unifycr/src/unifycr-cons.c
```diff
@@ -0,0 +1,85 @@
...
+ * @param str input string of the consistency model
+ *
+ * @return matching enum entry on success, UNIFYCR_CM_INVALID (-1) otherwise
+ */
+unifycr_cm_t unifycr_read_consistency(const char *str)
+{
+    int i = 0;
+    unifycr_cm_t cons = UNIFYCR_CM_INVALID;
+
+    for (i = 0; i < sizeof(consistency_strs)/sizeof(char *); i++)
```
nedbass: `sizeof(consistency_strs)/sizeof(consistency_strs[0])` would be a more general construct. But you may need a shorter array name to fit it all on one line. Alternatively you could have a macro

#define ARRAY_SIZE(x) (sizeof(x) / sizeof((x)[0]))

since you use this construct below as well.

sandrain:
@nedbass Thanks for the detailed review. I will apply your suggestions and will squash/rebase the fixes.

util/unifycr/src/unifycr-cons.c
sandrain: I will update with the first method. Macro can potentially have a name conflict.

util/unifycr/src/unifycr.c
sandrain: I was lazy because it was a debug code :-)  I will fix.

util/unifycr/src/unifycr-rm.c
sandrain: I will comment out.

sandrain:
@nedbass I pushed after squashing the fixes, but the checker is still not triggered. How can I make it work?

util/unifycr/src/unifycr-rm.c
```diff
@@ -0,0 +1,204 @@
...
+        n_nodes++;
+
+    if (ferror(fp)) {
+        ret = -errno;
+        goto out;
+    }
+
+    nodes = calloc(sizeof(char *), n_nodes);
+    if (!nodes) {
+        ret = errno;
```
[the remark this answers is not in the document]
sandrain: will fix it. thanks.

util/unifycr/src/unifycr-rm.c
```diff
@@ -0,0 +1,204 @@
...
+    if (str) {
+        ret = read_resource_lsf(resource);
+        goto out;
+    }
+
+#if 0
+    /* TODO: this options is currently not implemented. */
+    fprintf(stderr, "cannot detect resource manager. "
+                    "please provide node info with -n option\n");
+#endif
```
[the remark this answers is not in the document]
sandrain: will remove this.

sandrain:
@nedbass I applied the final fix based on your comments. What happens now is that because of the recently merged Danielle's patch (18c34e2), it becomes tricky to squash changes into the original commit of this PR. So, do you want me to open a new PR? Otherwise, I can push my commits without squashing, which travis would ignore.

nedbass:
@sandrain I'm not sure I understand. The recent commits from @dsikich were limited to the docs directory so shouldn't conflict with your changes. You should be able to rebase cleanly on origin/dev, squash your cleanup commit, and force update your branch, something like:

```
git fetch origin
git rebase origin/dev
git rebase -i origin/dev # squash
git push -f sandrain iss86-resource-manager-integration
```

sandrain:
@nedbass I rebased correctly again. Pl review it. Thanks.

nedbass:
@sandrain this is looking good. Thanks for addressing the feedback. Can you also remove the unifycr/rm directory structure?

Other than that, this is probably OK to merge and we can address further improvements in new issues and PRs. Although I think the design could be though through a bit more. It would be good to document this utility and why it's needed. From what I can tell, the intent is:

1. Parse unifycr.conf input file and command line arguments for unifycrd run time options
2. Parse environment variables to get a node list
3. Write out a new configuration file for unifycrd to parse out node list and runtime options 
4. Start unifycrd

It's not clear to me why we wouldn't just do all those steps in unifycrd and skip the intermediate configuration file. What does the separate utility buy us?

sandrain:
@nedbass I have just removed the /rm directory. I missed it previously, sorry. About the command-lline utility, my understanding is that it aims to provide a simple(r) mechanism for users to use unifycr in their job scripts. Obviously, launching a file system manually with the mpirun family command plus many cryptic arguments is not ideal for end users. However, it is also true that all those things could be all done inside the server daemon itself. In fact, I was trying to follow what we wrote in the last milestone report, probably without thinking too much about it. Maybe, we can discuss openly about it.

In addition, as you might have already seen, we currently do not have well-defined configuration options and where each option should go, e.g., /etc/unifycr, env, or cli arguments. This also requires further discussions.

PR #107, opened 2018-03-12
Make unifycrd run as a daemon

nedbass:
The unifycrd process currently runs in the foreground. Add code
with minimal set of steps needed to run it in the background as
a daemon.

1. Fork and exit from parent so child runs in the background
2. Set the daemon umask to 0 so file modes passed to open() and
   mkdir() fully control access modes
3. Call setsid() to create a new session and detach from controlling tty
4. Change current working directory to / so daemon doesn't block
   filesystem unmounts
5. close STDIN, STDOUT, and STDERR
6. Fork again to abdicate session leader position to guarantee
   daemon cannot reacquire a controlling TTY

Optional steps not performed in this patch but that might still be
useful or appropriate:

1. Close all open file descriptors
2. Reset signal handlers
3. Sanitize environment variables
4. Create a /var/run/unifycrd.pid file with PID of unifycrd

### DIFF
--- a/server/src/unifycr_init.c
+++ b/server/src/unifycr_init.c
@@ -29,6 +29,7 @@
 
 #include <mpi.h>
 #include <unistd.h>
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


nedbass:
One question for reviewers to consider is when is the right time to call daemonize()? The current patch does it at the top of main(), before all the error checking is done. It might make more sense to put off daemonizing until after all the error checks have passed. I just wasn't sure how fork() affects MPI communicators, so I played it safe and did the fork first. @dsikich @adammoody what do you think?

dsikich:
@nedbass Hmm I'm not sure how fork() affects MPI communicators either. @adammoody do you know?

dsikich:
@nedbass I'm curious how the shutdown of the daemon works? Are we planning on shutting it down when the uesr's allocation ends? I'm not seeing where that happens currently?

Also, I've been thinking about how this would work with MPI more, and talking with Adam. It sounds like the behavior isn't always clear depending on the MPI configuration and resource manager. Although we are planning to remove our MPI dependency in the future. 

I'm also curious did you run this successfully already? 

Thanks!

PR #108, opened 2018-03-14
Global file size fix: store fattr with gfid key

ivadmitry:
Global file size fix: store fattr with gfid key

### Description
Metadata server stores fattr under the wrong key at meta_process_attr_set(). Then meta_process_attr_get() cannot access fattr hence global file size becomes unavailable for non-local nodes.

### Motivation and Context
Global file size is not available for all clients. Cached local (less) value is used instead.

### How Has This Been Tested?
Simul test suite with extra checks for file size.

### Types of changes
Warning: this is a potentially breaking change.

### Checklist:
- [x ] My code follows the UnifyCR code style requirements.
- [x ] All new and existing tests passed.
- [x ] All commit messages are properly formatted.


dsikich:
@ivadmitry I'm just curious if you had something specific in mind when you mentioned, "Warning: this is a potentially breaking change."

ivadmitry:
I was wrong.
meta_process_attr_get() is the only reader of 'fattr' meta db and I've checked the value is correctly retrieved by gfid.

dsikich:
@ivadmitry ok, great. Thanks!

nedbass:
Thanks for the refresh. The new commit message looks good, but your branch needs be cleanly rebased on LLNL/dev and force-updated. There should be just one commit in the PR, with no merge commit.

ivadmitry:
@nedbass No warning now. Thanks!

nedbass:
@ivadmitry I'd be interested in knowing more details about how to reproduce this bug if you're able to share them.

ivadmitry:
The simplest scenario I could suggest is that:
- create a file from node A and write+fsync it to some size;
- open the file from node B and get the size.
In  client/src/unifycr.c:unifycr_fid_open() we get inside fid<0 case at line 1183, overwrite the global fattr record by set_global_file_meta() at line 1214 and exit the open function with zero file length in metadata (unifycr_fattr_t).
With the fix in place the execution would go to line 1154 and the correct "global" file size to be set in the local cache by ins_file_meta():
meta->real_size = ptr_meta->file_attr.st_size;
...
ins_file_meta(&unifycr_fattrs, ...

Thanks!

ivadmitry:
PS. If the file is always opened with O_TRUNC, this issue could not be reproduced. That is the case with ALL tests in client/tests folder.

PR #110, opened 2018-03-28
Implementation of launching unifycrd from unify command line utility.

sandrain:
Lauching the unifycr daemon from the unifycr command line utility.

### Description

Implementation of launching unifycrd from unify command line utility.
- launching the unifycrd with exec mpirun
- adding more environment variables support (the ones in the documentation)
- fix a bug of getting a duplicated node list in PBS

### Motivation and Context

This is a follow-up work of #103 

### Types of changes


util/unifycr/src/unifycr-env.c
```diff
@@ -52,6 +52,26 @@ int unifycr_read_env(unifycr_env_t *env)
...
+        env->unifycr_server_debug_log = strdup(val);
+
+    val = getenv("UNIFYCR_META_SERVER_RATIO");
+    if (val)
+        env->unifycr_meta_server_ratio = strtoul(val, NULL, 0);
+
+    val = getenv("UNIFYCR_CHUNK_MEM");
+    if (val)
+        env->unifycr_chunk_mem = strtoul(val, NULL, 0);
+
```
nedbass: Shouldn't we be setting these in the configuration file now?

util/unifycr/src/unifycr-rm.c
```diff
@@ -169,30 +244,82 @@ static int read_resource_lsf(unifycr_resource_t *resource)
...
+};
+
 int unifycr_read_resource(unifycr_resource_t *resource)
 {
-    int ret = -1;
     char *str = NULL;
 
+    /*
+     * "do not use assignement in if condition"
+     * so, here we use goto instead of if .. else ... any better way?
```
nedbass: Perhaps this construct, otherwise ignore the style warning.

```
char *pbs_jobid = getenv("PBS_JOBID");
char *slurm_jobid = getenv("SLURM_JOBID");
char *lsb_jobid = getenv("LSB_JOBID");

if (pbs_jobid) {
 //
} else if (slurm_jobid) {
 //
} else if (lsb_jobid) {
  //
}
```

util/unifycr/src/unifycr-rm.c
```diff
@@ -169,30 +244,82 @@ static int read_resource_lsf(unifycr_resource_t *resource)
...
+    return resource_managers[resource->rm].read_resource(resource);
+}
+
+/*
+ * TODO: currently it directly uses mpirun command in any cases.  The command
+ * should be:
+ * $ mpirun -n NUMNODES -np NUMNODES unifycrd
+ */
+static char *mpirun_newargv[] = {
+    "mpirun", "-n", NULL, "-np", NULL, BINDIR "/unifycrd", (char *) NULL
```
nedbass: It would be useful to be able to specify the path to unifycrd in the configuration file, and default to the BINDIR location if unspecified, rather than hard coding it here.

util/unifycr/src/unifycr-env.c
sandrain: I will create the configuration file entries, which can be overridden by the environmental variables.

util/unifycr/src/unifycr-rm.c
sandrain: @nedbass agreed. i will make changes accordingly.

util/unifycr/src/unifycr-rm.c
sandrain: I have addressed this as you suggested.

util/unifycr/src/unifycr-rm.c
```diff
@@ -169,30 +244,73 @@ static int read_resource_lsf(unifycr_resource_t *resource)
...
+    mpirun_newargv[2] = nbuf;
+    mpirun_newargv[4] = nbuf;
+    mpirun_newargv[5] = state->unifycrd_path ?
+                        state->unifycrd_path : BINDIR "/unifycrd";
+
+    execvp(mpirun_newargv[0], mpirun_newargv);
+    perror("failed to launch unifycrd (execvp)");
+
+    return -1;
 }
```
dsikich: @sandrain @nedbass Where is this being started from? When I run:
```bash
catalyst159{sikich1}49: build/util/unifycr/src/unifycr start
failed to launch unifycrd (execvp): No such file or directory
```

dsikich:
@sandrain Also, I think something to remember is that /tmp/unifycr doesn't exist by default. Are we going to assume the user will create /tmp/unifycr or are we going to create it for them?

sandrain:
@dsikich I will update to create /tmp/unifycr directory. For the execvp (the path problem), it would probably look for the installation path (make install). I will double-check it.

dsikich:
@sandrain After looking at it under a debugger it was because the launch command was using "mpirun" which is not available using mvapich on the cluster I am using. So, it starts up if I load openmpi first. But, I don't think we can rely on launching with mpirun in every case.

sandrain:
@dsikich thanks for checking that out. actually using the mpirun was the initial implementation, and i was planning to implement other cases after completing the last milestone. I think for now I will update this patch to include the process of creating the directory (/tmp/unifycr). For the launching part, I think we need other work after this patch.

dsikich:
@sandrain yeah, that makes sense to me.

PR #116, opened 2018-04-20
Add note in docs about MPI configuration requirements

dsikich:
The mdhim key-value store requires an MPI with MPI_THREAD_MULTIPLE
support and per-object locking of critical sections (exclues OpenMPI)


dsikich:
@kathrynmohror Sure, I just updated it for the current version of Open MPI.

PR #120, opened 2018-05-10
drop code for BGQ

adammoody:
Drop code for BlueGene/Q persistent memory to simplify client code, since we won't be using that machine.


adammoody:
@nedbass, it seems the style checker is not using the latest commits.  Can you see what I'm doing wrong?

nedbass:
@adammoody I think you need to squash all the commits together. If any individual patch in the stack has style issues it will fail.

adammoody:
Thanks for the review, @dsikich

PR #121, opened 2018-05-10
client: add param_get routines to isolate getenv calls

adammoody:
In preparation for an eventual config file for the client, create initial interface to read parameter values.  This isolates getenv calls to one function, and it allows for a name and default value.  The functions return 1 if the user attempted to set the value, and 0 if the default was used.

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -1417,6 +1417,7 @@ static int unifycr_get_spillblock(size_t size, const char *path)
             /* spillover block exists; attach and return */
             spillblock_fd = __real_open(path, O_RDWR);
         } else {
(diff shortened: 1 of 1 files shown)


adammoody:
This one treads water around some other ongoing work.  If it looks like it will make merging a headache for someone, we can wait to merge this one second.

dsikich:
@adammoody yeah, I think right now this conflicts with PR #124

adammoody:
Let's hold this, and I'll update this PR after we merge PR #124

MichaelBrim:
@adammoody I'm currently working on integrating the client side with the server-generated runstate. From first glance, it looks like my work will subsume everything in this PR, so please hold off on any more updates until I can confirm.

Issue #125, opened 2018-05-22
Support of lseek64(2) syscall


sandrain:
`lseek64(2)` is missing in the current version, and is required to support applications, e.g., HDF5.

sandrain:
The PR (#126) has been merged, and will close this.

Issue #122, opened 2018-05-11
Support of stat family syscalls


sandrain:
The current version of UnifyCR does not fully support `stat` family system calls, which is required to support major applications (#114). The following system calls are expected to be implemented:

```
lstat
fstat
fxstat
```

sandrain:
The merged PR (#126) has addressed this issue, thus closing this.

Issue #114, opened 2018-04-20
HDF5 dataset creation support


sandrain:
### Testing environment
The test has been conducted in our testbed with three x86 nodes:
* Intel Xeon  E5-2603
* 64 GB RAM
* mpi/mpich-3.2-x86_64

The `write_test_static` seems to run fine.

### HDF5 dataset creation test
The basic hdf5 file creation is not working as it is supposed to be. The sequence of hdf5 calls are as follows:

```
H5Fcreate
H5Screate_simple
H5Dcreate2
H5Dclose
H5Sclose
H5Fclose
```

The full testing program was taken from [C Examples In HDF5 Source Code](https://support.hdfgroup.org/ftp/HDF5/examples/src-html/c.html), the create [example](https://support.hdfgroup.org/ftp/HDF5/current/src/unpacked/examples/h5_crtdat.c).
 It has been modified to use MPI as follows:

```
/*
 *  This example illustrates how to create a dataset that is a 4 x 6 
 *  array.  It is used in the HDF5 Tutorial.
 */

#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <fcntl.h>
#include <string.h>
#include <mpi.h>
#include "hdf5.h"
#define FILE "dset.h5"

int main(int argc, char **argv) {

    int rank, rank_num;

    hid_t       file_id, dataset_id, dataspace_id;  /* identifiers */
    hsize_t     dims[2];
    herr_t      status = 0;

    MPI_Init(&argc, &argv);
    MPI_Comm_size(MPI_COMM_WORLD, &rank_num);
    MPI_Comm_rank(MPI_COMM_WORLD, &rank);

    unifycr_mount("/tmp", rank, rank_num, 0, 1);

    MPI_Barrier(MPI_COMM_WORLD);

    if (rank == 0) {
        /* Create a new file using default properties. */
        file_id = H5Fcreate(FILE, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);
        printf("H5Fcreate: %d\n", file_id);

        /* Create the data space for the dataset. */
        dims[0] = 4; 
        dims[1] = 6; 
        dataspace_id = H5Screate_simple(2, dims, NULL);
        printf("H5Screate_simple: %d\n", dataspace_id);

        /* Create the dataset. */
        dataset_id = H5Dcreate2(file_id, "/tmp/dset", H5T_STD_I32BE,
[...]

PR #126, opened 2018-05-22
Implementation of missing syscalls from stat and lseek families

sandrain:
This patch contains an implementation of missing syscalls in `client/src/unifycr-sysio.c`. In specific, it adds `fstat(2)`, `__xstat(2)`, `__fxstat(2)`, `__lxstat(2)` and `lseek64(2)`.

This addresses: #114 #122 #125 

### Types of changes

### DIFF
--- a/client/src/unifycr-sysio.c
+++ b/client/src/unifycr-sysio.c
@@ -325,28 +325,81 @@ int UNIFYCR_WRAP(remove)(const char *path)
 
 int UNIFYCR_WRAP(stat)(const char *path, struct stat *buf)
 {
+    int ret = 0;
+
     DEBUG("stat was called for %s....\n", path);
     if (unifycr_intercept_path(path)) {
+        if (!buf) {
+            errno = EFAULT;
+            return -1;
+        }
+
         int fid = unifycr_get_fid_from_path(path);
         if (fid < 0) {
             errno = ENOENT;
             return -1;
         }
 
-        unifycr_fid_stat(fid, buf);
+        ret = unifycr_fid_stat(fid, buf);
+        if (ret < 0)
+            errno = ENOENT;
 
-        return 0;
+        return ret;
     } else {
         MAP_OR_FAIL(stat);
-        int ret = UNIFYCR_REAL(stat)(path, buf);
+        ret = UNIFYCR_REAL(stat)(path, buf);
         return ret;
     }
 }
 
+int UNIFYCR_WRAP(fstat)(int fd, struct stat *buf)
+{
+    int ret = 0;
+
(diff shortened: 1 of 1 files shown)


client/src/unifycr-sysio.c
```diff
@@ -706,12 +706,19 @@ off_t UNIFYCR_WRAP(lseek)(int fd, off_t offset, int whence)
...
     /* check whether we should intercept this file descriptor */
-    if (unifycr_intercept_fd(&fd)) {
-        /* ERROR: fn not yet supported */
-        fprintf(stderr, "Function not yet supported @ %s:%d\n", __FILE__, __LINE__);
-        errno = EBADF;
-        return (off64_t) (-1);
+    if (unifycr_intercept_fd(&tmpfd)) {
+        if (sizeof(off_t) == sizeof(off64_t))
+            return (off64_t) UNIFYCR_WRAP(lseek) (fd, (off_t) offset, whence);
+        else {
```
adammoody: This looks good to me.  I personally prefer braces around even single lines to prevent future bugs:

If (cond) {
  single line
} else {

However, I think our style checker script forces us to drop them for single-line blocks:

sandrain: Yes, I remember the last time when the checker complained when I put braces around a single statement.

sandrain:
I have put missing syscalls in this single patch. After this is merged, I will submit once more PR which fixes a bug in the `__wrap_write()`.

client/src/unifycr-sysio.c
```diff
@@ -325,28 +325,81 @@ int UNIFYCR_WRAP(remove)(const char *path)
 
 int UNIFYCR_WRAP(stat)(const char *path, struct stat *buf)
 {
+    int ret = 0;
+
     DEBUG("stat was called for %s....\n", path);
     if (unifycr_intercept_path(path)) {
+        if (!buf) {
```
sandrain: If `buf` is not valid, we should set `EFAULT`. I made some changes for this in existing `stat(2)` implementations.

client/src/unifycr-sysio.c
```diff
@@ -356,31 +409,90 @@ int UNIFYCR_WRAP(__xstat)(int vers, const char *path, struct stat *buf)
         }
 
         /* get meta data for this file */
-        unifycr_fid_stat(fid, buf);
+        ret = unifycr_fid_stat(fid, buf);
```
sandrain: `unifycr_fid_stat()` returns `-1` when it fails to find the metadata. Although it is not very likely to happen, still we would need to set `errno` when we return `-1` to caller.

Issue #127, opened 2018-06-06
UNIFYCR_LOG enum constant


sandrain:
`UNIFYCR_LOG` is defined at `client/src/unifycr-internal.h`, together with `UNIFYCR` and `UNIFYCR_STRIPE`. The file system types should be passed as the last parameter when calling `unifycr_mount()`. But, is there any cases that we need to pass something else than `UNIFYCR_LOG`? As far as I remember, I do not think other types are even functional.

In `client/src/unifycr_sysio.c`, `__wrap_pwrite()` does not even check whether the type is `UNIFYCR_LOG` or something else, while `__wrap_write()` does.

```
ssize_t UNIFYCR_WRAP(pwrite)(int fd, const void *buf, size_t count,
                             off_t offset)
{
    /* equivalent to write(), except that it writes into a given
     * position without changing the file pointer */
    /* check whether we should intercept this file descriptor */
    if (unifycr_intercept_fd(&fd)) {
        /* get pointer to file descriptor structure */
        unifycr_fd_t *filedesc = unifycr_get_filedesc_from_fd(fd);
        if (filedesc == NULL) {
            /* ERROR: invalid file descriptor */
            errno = EBADF;
            return (ssize_t) (-1);
        }

        /* write data to file */
        int write_rc = unifycr_fd_write(fd, offset, buf, count);
        if (write_rc != UNIFYCR_SUCCESS) {
            errno = unifycr_err_map_to_errno(write_rc);
            return (ssize_t) (-1); 
        }

        /* return number of bytes read */
        return (ssize_t) count;
    } else {
        MAP_OR_FAIL(pwrite);
        ssize_t ret = UNIFYCR_REAL(pwrite)(fd, buf, count, offset);
        return ret;
    }
}

/* ... */

ssize_t UNIFYCR_WRAP(write)(int fd, const void *buf, size_t count)
{
    ssize_t ret;

    /* check whether we should intercept this file descriptor */
    if (unifycr_intercept_fd(&fd)) {
        /* get pointer to file descriptor structure */
        unifycr_fd_t *filedesc = unifycr_get_filedesc_from_fd(fd);
        if (filedesc == NULL) {
            /* ERROR: invalid file descriptor */
            errno = EBADF;
            return (ssize_t) (-1);
        }

        if (fs_type != UNIFYCR_LOG) {
            /* write data to file */
            int write_rc = unifycr_fd_write(fd, filedesc->pos, buf, count);
[...]

PR #128, opened 2018-06-11
Removing the fs_type_t constants (UNIFYCR_LOG)

sandrain:
This PR is to address #127, and partially #114.

- Removing the enum definition of fs_type_t, because UnifyCR technically
  assumes that it should always be UNIFYCR_LOG
- Removing the last argument (fs_type_t) in unifycr_mount()
- Modifying the test programs accordingly

### Types of changes

### DIFF
--- a/client/src/unifycr-sysio.c
+++ b/client/src/unifycr-sysio.c
@@ -234,6 +234,7 @@ int UNIFYCR_WRAP(truncate)(const char *path, off_t length)
     if (unifycr_intercept_path(path)) {
         /* lookup the fd for the path */
         int fid = unifycr_get_fid_from_path(path);
+
         if (fid < 0) {
             /* ERROR: file does not exist */
             DEBUG("Couldn't find entry for %s in UNIFYCR\n", path);
@@ -382,6 +383,9 @@ int UNIFYCR_WRAP(fstat)(int fd, struct stat *buf)
  * NOTE on __xstat(2), __lxstat(2), and __fxstat(2)
  * The additional parameter vers shall be 3 or the behavior of these functions
  * is undefined. (ISO POSIX(2003))
+ *
+ * from /sys/stat.h, it seems that we need to test if vers being _STAT_VER,
+ * instead of using the absolute value 3.
  */
 
 int UNIFYCR_WRAP(__xstat)(int vers, const char *path, struct stat *buf)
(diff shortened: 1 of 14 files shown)


client/src/unifycr-sysio.c
```diff
@@ -463,12 +467,17 @@ int UNIFYCR_WRAP(__fxstat)(int vers, int fd, struct stat *buf)
...
+
     /* check whether we should intercept this file descriptor */
     if (unifycr_intercept_fd(&fd)) {
+#if 0
+        /* TODO: do we really care? */
         if (vers != 3) {
             errno = EINVAL;
             return -1;
         }
+#endif
```
dsikich: @sandrain I'm not sure the vers check matters either... I wonder can we just delete this unless we can think of a reason it needs to be there?

sandrain: I have fixed this version check. It seems like `vers` should be `_STAT_VER`, according to `bits/stat.h`:

```
/* Versions of the `struct stat' data structure.  */                                                                                                                                                                                                                                                                          
#ifndef __x86_64__                                                                                                                                                                                                                                                                                                            
# define _STAT_VER_LINUX_OLD    1                                                                                                                                                                                                                                                                                             
# define _STAT_VER_KERNEL       1                                                                                                                                                                                                                                                                                             
# define _STAT_VER_SVR4         2                                                                                                                                                                                                                                                                                             
# define _STAT_VER_LINUX        3                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                              
/* i386 versions of the `xmknod' interface.  */                                                                                                                                                                                                                                                                               
# define _MKNOD_VER_LINUX       1                                                                                                                                                                                                                                                                                             
# define _MKNOD_VER_SVR4        2                                                                                                                                                                                                                                                                                             
# define _MKNOD_VER             _MKNOD_VER_LINUX /* The bits defined below.  */                                                                                                                                                                                                                                               
#else                                                                                                                                                                                                                                                                                                                         
# define _STAT_VER_KERNEL       0                                                                                                                                                                                                                                                                                             
# define _STAT_VER_LINUX        1                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                              
/* x86-64 versions of the `xmknod' interface.  */                                                                                                                                                                                                                                                                             
# define _MKNOD_VER_LINUX       0                                                                                                                                                                                                                                                                                             
#endif                                                                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                              
#define _STAT_VER               _STAT_VER_LINUX
```

Currently returns `EINVAL` if `vers` is not `_STAT_VER` (potentially could return `ENOSYS`).

dsikich:
I think this is good to merge.

Issue #72, opened 2018-01-03
Improve error handling if /dev/shm too small


nedbass:
In #59 it was reported that an application crashed with SIGBUS if /dev/shm was configured with insufficient space. On the mailing list it was suggested that this case could be detected and reported more gracefully if we replace the use of `ftruncate()` with `fallocate()`, although that would sacrifice portability since `fallocate()` is Linux-specific. @adammoody also pointed out that we'd need to examine how memory pages are assigned to NUMA banks with `fallocate()`.

Issue #59, opened 2017-12-22
SIGBUS accessing shared memory


nedbass:
@jgmoore-or reported a SIGBUS error accessing shared memory on the unifycr mailing list. Opening an issue here for bug tracking purposes. Email thread quoted below.

```
Adam,

I still can't get past this error:

Program received signal SIGBUS, Bus error.

0x00007ffff05973cb in __memcpy_ssse3_back () from /lib64/libc.so.6

Missing separate debuginfos, use: debuginfo-install glibc-2.17-157.el7.x86_64
libgcc-4.8.5-11.el7.x86_64 libgfortran-4.8.5-11.el7.x86_64
libquadmath-4.8.5-11.el7.x86_64 mpich-3.2-3.2-2.el7.x86_64
numactl-libs-2.0.9-6.el7_2.x86_64 openssl-libs-1.0.1e-60.el7.x86_64
zlib-1.2.7-17.el7.x86_64

(gdb) bt

#0  0x00007ffff05973cb in __memcpy_ssse3_back () from /lib64/libc.so.6
#1  0x0000000000409926 in unifycr_logio_chunk_write (fid=1, pos=49152, meta=0x7fffddcf443c, chunk_id=0, chunk_offset=49152, buf=0x66280c0, count=4096)
    at unifycr-fixed.c:378
#2  0x000000000040a574 in unifycr_fid_store_fixed_write (fid=1, meta=0x7fffddcf443c, pos=49152, buf=0x66280c0, count=4096) at unifycr-fixed.c:699
#3  0x0000000000405280 in unifycr_fid_write (fid=1, pos=49152, buf=0x66280c0, count=4096) at unifycr.c:746
#4  0x000000000040b1aa in unifycr_fd_write (fd=1, pos=49152, buf=0x66280c0, count=4096) at unifycr-sysio.c:541
#5  0x000000000040d45c in __wrap_pwrite (fd=1, buf=0x66280c0, count=4096, offset=49152) at unifycr-sysio.c:1463
#6  0x000000000040bc01 in __wrap_write (fd=1025, buf=0x66280c0, count=4096) at unifycr-sysio.c:865
#7  0x00000000004043a1 in main (argc=13, argv=0x7fffffffe358) at test_write.c:143

Hence I can only successfully write a file up to 12 4096 byte pages.  Anything
larger than this, independent of block/transfer sizes, will fail.

I've checked, and the shared memory for the superblock, chunks, etc. is being
[...]

PR #142, opened 2018-07-18
Use posix_fallocate(3) instead of ftruncate(2) when possible

sandrain:
Use posix_fallocate(3) instead of ftruncate(2) when possible for expanding the shared memory area.

### Description
- Adding the test of posix_fallocate(3) in configure.ac.
- Modifying the client/src/unifycr.c to use posix_fallocate(3) if available.

### Motivation and Context
This patch partially solves #59 and #72. When a system has `posix_fallocate(3)`, unifycr is built with the `posix_fallocate(3)` and gracefully fails if `/dev/shm` space is not sufficient. Currently, unifycr generates a runtime bus error in such a case.

IMO, the ultimate fix for this problem would be
- Use `posix_fallocate(3)` when available.
- Use traditional shm interface instead of posix shm_open/mmap interface, or
- Check the `/dev/shm` size with `statfs(2)` before calling `ftruncate(2)`?

### Types of changes

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -40,6 +40,8 @@
  * Please also read this file LICENSE.CRUISE
  */
 
+#include <config.h>
+
 #define _GNU_SOURCE
 #include <sched.h>
 #include "unifycr-runtime-config.h"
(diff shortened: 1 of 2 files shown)


sandrain:
@dsikich It's `config.h.in`, which is going to be a template to generate `config.h`. I think it was somehow missed in the repository.

sandrain:
@dsikich You are right. I will resubmit this.

PR #143, opened 2018-07-19
add mpiio tests for reading and writing

dsikich:
This adds simple mpiio tests for reading and writing. Read or write
options are specified via the command line input to the test_mpiio
program. The tests are also built for gotcha and static interception
through the build system.


kathrynmohror:
@dsikich was there any copyright text for the tests that need to be included?

dsikich:
@kathrynmohror Good catch, I updated it with UnifyCR copyright.

PR #177, opened 2018-10-04
Add pure POSIX version of sysio-writeread example

MichaelBrim:
### Description

This commit updates the sysio-writeread.c example to allow for a pure POSIX build,
which doesn't include `unifycr.h` and does not call `unifycr_mount` and `unifycr_unmount`.

This pure POSIX version is the baseline used for recent Summitdev experiments.

### Types of changes

### Checklist:

### DIFF
--- a/examples/src/Makefile.am
+++ b/examples/src/Makefile.am
@@ -1,6 +1,6 @@
 noinst_PROGRAMS = sysio-write-gotcha sysio-write-static \
                   sysio-read-gotcha sysio-read-static \
-                  sysio-writeread-gotcha sysio-writeread-static \
+                  sysio-writeread-gotcha sysio-writeread-static sysio-writeread-posix \
                   sysio-writeread2-gotcha sysio-writeread2-static \
                   sysio-dir-gotcha sysio-dir-static \
                   sysio-stat-gotcha sysio-stat-static \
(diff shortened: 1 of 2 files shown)


MichaelBrim:
@sandrain Yes, this is the version of sysio-writeread that I ran on Summitdev, both with and without UnifyCR. For the other examples, the CLI flag isn't enough, we need to avoid linking against libunifycr (and for the static case, not use the linker wrapping).

sandrain:
@MichaelBrim yes, you are correct that the runtime flag does not work if linked statically.

MichaelBrim:
Had to fix a bug, so I pushed a new commit. Then the Travis build timed-out. Anyone know how to force a Travis retry?

CamStan:
Looks like you got it to rebuild. I've noticed several builds have been timing out lately on line 526 of the Travis build log:
```
sudo apt-get install --yes -qq libhdf5-openmpi-dev
```
If nothing gets printed to the terminal for 10 minutes, Travis will time out. Not sure why it takes so long sometimes or what we can do to speed it up. We can increase the amount of time before Travis times out if we want, as shown [here](https://docs.travis-ci.com/user/common-build-problems/#my-builds-are-timing-out),
```
install: travis_wait mvn install
```
or
```
install: travis_wait n mvn install
```
 where `n` is the amount of minutes you want it to wait (default is 20), but they caution not to overuse it.

MichaelBrim:
@CamStan yeah, I figured out that I needed to sign into Travis to be able to see the "Restart Build" button. My timeout was at the same place.

PR #182, opened 2018-10-15
Read bug fix and a file copy example using read(2)/write(2) syscalls.

sandrain:
This patch adds a file copy example program. Also fixes the read bug (#183, #185) and the close issue (#186).

- Fixes bug in __wrap_read (to return the corret bytecount).
- Fixes bug in read/pread to return EOF or EINVAL properly.
- Internally syncs file metadata when `close(2)` is called for files opened for writing.
- Changed the default mount point to /unifycr (from /tmp) in example programs.
- (minor) Correctly default file name in help messages for hdf5 examples.

### Types of changes

### DIFF
--- a/client/src/unifycr-dirops.c
+++ b/client/src/unifycr-dirops.c
@@ -106,7 +106,7 @@ DIR *UNIFYCR_WRAP(opendir)(const char *name)
          */
         meta->size = sb->st_size;
         meta->chunks = sb->st_blocks;
-        meta->real_size = sb->st_size;
+        meta->log_size = 0; /* no need of local storage for dir operations */
     } else {
         fid = unifycr_fid_create_file(name);
         if (fid < 0) {
@@ -118,7 +118,7 @@ DIR *UNIFYCR_WRAP(opendir)(const char *name)
         meta->is_dir = 1;
         meta->size = sb->st_size;
         meta->chunks = sb->st_blocks;
-        meta->real_size = sb->st_size;
+        meta->log_size = 0;
     }
 
     _dirp = unifycr_dirstream_alloc(fid);
--- a/client/src/unifycr-sysio.c
+++ b/client/src/unifycr-sysio.c
@@ -653,7 +653,7 @@ int unifycr_fd_write(int fd, off_t pos, const void *buf, size_t count)
         unifycr_filemeta_t *meta = unifycr_get_meta_from_fid(fid);
         if (write_rc == 0) {
             meta->size = newpos;
-            meta->real_size = pos + count;
+            meta->log_size = pos + count;
         }
     }
     return write_rc;
@@ -917,12 +917,17 @@ ssize_t UNIFYCR_WRAP(read)(int fd, void *buf, size_t count)
     /* check whether we should intercept this file descriptor */
     if (unifycr_intercept_fd(&fd)) {
         /* get pointer to file descriptor structure */
+        unifycr_filemeta_t *meta = unifycr_get_meta_from_fid(fd);
         unifycr_fd_t *filedesc = unifycr_get_filedesc_from_fd(fd);
-        if (filedesc == NULL) {
+
+        if (meta == NULL || filedesc == NULL) {
             /* ERROR: invalid file descriptor */
             errno = EBADF;
             return (ssize_t) (-1);
         }
+
+        if (filedesc->pos >= meta->size)
+            return 0;   /* EOF */
 
         /* read data from file */
         size_t retcount;
@@ -934,12 +939,15 @@ ssize_t UNIFYCR_WRAP(read)(int fd, void *buf, size_t count)
         tmp_req.length = count;
         tmp_req.offset = filedesc->pos;
 
+        /*
+         * this returns error code, which is zero for successful cases.
+         */
         int ret = unifycr_fd_logreadlist(&tmp_req, 1);
 
-        if (!ret)
+        if (ret)
             retcount = 0;
         else
-            retcount = count;
+            retcount = tmp_req.length;
 
         /* update position */
         filedesc->pos += (off_t) retcount;
@@ -1401,6 +1409,21 @@ int unifycr_fd_logreadlist(read_req_t *read_req, int count)
     int *ptr_num = NULL;
 
     /*
+     * Adjust length for fitting the EOF.
+     */
+    for (i = 0; i < count; i++) {
+        read_req_t *req = &read_req[i];
+        unifycr_filemeta_t *meta = NULL;
+
+        meta = unifycr_get_meta_from_fid(req->fid - unifycr_fd_limit);
+        if (!meta)
+            return -1;
+
+        if (req->offset + req->length > meta->size)
+            req->length = meta->size - req->offset;
+    }
+
+    /*
      * Todo: When the number of read requests exceed the
      * request buffer, split list io into multiple bulk
      * sends and transfer in bulks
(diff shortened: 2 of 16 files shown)


sandrain:
@MichaelBrim I've changed to `/unifycr`.

sandrain:
I also included a fix for #185. This allows the following test work:

```
[root@ninja71 _hstest]$ mpirun -n 1 -f nodelist.txt `pwd`/build/examples/src/sysio-cp-static /etc/passwd /unifycr/
[root@ninja71 _hstest]$ mpirun -n 1 -f nodelist.txt `pwd`/build/examples/src/sysio-stat-static --filename=/unifycr/passwd                                      

## [RANK 0] /unifycr/passwd
File type:                regular file
Device containing i-node: major=0   minor=0
I-node number:            373367142
Mode:                     100644
Number of (hard) links:   0
Ownership:                UID=0   GID=0
File size:                1554 bytes
Optimal I/O block size:   4096 bytes
512B blocks allocated:    1
Last file access:         Wed Oct 17 14:28:13 2018
Last file modification:   Wed Oct 17 14:28:13 2018
Last status change:       Wed Oct 17 14:28:13 2018



[root@ninja71 _hstest]$ mpirun -n 1 -f nodelist.txt `pwd`/build/examples/src/sysio-cp-static /unifycr/passwd .                                      
[root@ninja71 _hstest]$ diff /etc/passwd passwd
[root@ninja71 _hstest]$ echo $?
0

```

adammoody:
@sandrain, this PR ready for merge, or are you waiting on more reviews?

sandrain:
@adammoody I would ask your review. The file size might be still buggy (based on our discussion in the mailing list about `size` vs. `real_size`, but it won't affect other works/tests since the current version is more problematic. I will submit another PR for fixing the size problem separately.

client/src/unifycr-dirops.c
```diff
@@ -106,7 +106,7 @@ DIR *UNIFYCR_WRAP(opendir)(const char *name)
          */
         meta->size = sb->st_size;
         meta->chunks = sb->st_blocks;
-        meta->real_size = sb->st_size;
+        meta->size = sb->st_size;
```
adammoody: This meant to be log_size instead?

client/src/unifycr-dirops.c
```diff
@@ -118,7 +118,7 @@ DIR *UNIFYCR_WRAP(opendir)(const char *name)
         meta->is_dir = 1;
         meta->size = sb->st_size;
         meta->chunks = sb->st_blocks;
-        meta->real_size = sb->st_size;
+        meta->size = sb->st_size;
```
adammoody: log_size?

client/src/unifycr-sysio.c
```diff
@@ -1400,6 +1408,21 @@ int unifycr_fd_logreadlist(read_req_t *read_req, int count)
     int *ptr_size = NULL;
     int *ptr_num = NULL;
 
+    /*
+     * Adjust length for fitting the EOF.
+     */
+    for (i = 0; i < count; i++) {
```
adammoody: haven't parsed this code to fully understand it, but can this be done without a for loop?

adammoody: Nevermind, I see now that we're dealing with an array of requests.

client/src/unifycr-sysio.c
```diff
@@ -1523,13 +1546,24 @@ ssize_t UNIFYCR_WRAP(pread)(int fd, void *buf, size_t count, off_t offset)
...
             /* ERROR: invalid file descriptor */
             errno = EBADF;
             return (ssize_t) (-1);
         }
 
+        if (offset == meta->size)
+            return 0;
+
+        /* we should set errno conforming to lseek(2) */
+        if (offset > meta->size) {
```
adammoody: Haven't tested pread in this case, but it looks like lseek allows for seeking past the end of a file:

"The lseek() function allows the file offset to be set beyond the end of the file (but this does not change the size of the file).  If data is later written at this point, subsequent reads of the data in the gap (a "hole") return null bytes ('\0') until data is actually written into the gap."

Perhaps just returning EOF is enough in this case?

client/src/unifycr-dirops.c
dsikich: @sandrain getting ready to merge this, but wanted to check-in about this comment first. Line 107 & 109 are the same so we were assuming you meant to have line 109 be meta->log_size?

client/src/unifycr-sysio.c
sandrain: I will check with the `client/src/unifycr-dirops.c`. I remember I had a build error for that, and probably I have missed to put the fix together.

For the seeking the beyond of the file, `pread(2)` seems to allow reading beyond the length by simply returning 0. I have tested with the following program:

```
[root@ninja82 pread]$ cat test-pread.c
#include <stdio.h>
#include <stdint.h>
#include <stdlib.h>
#include <assert.h>
#include <errno.h>
#include <string.h>
#include <fcntl.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <unistd.h>

static const char *testfile = "passwd";

int main(int argc, char **argv)
{
    int ret = 0;
    int fd = 0;
    off_t offset = 0;
    struct stat sb = { 0, };
    char buf[512] = { 0, };

    fd = open(testfile, O_RDONLY);
    assert(fd > 0);

    ret = fstat(fd, &sb);
    assert(ret == 0);

    printf("size = %lu\n", sb.st_size);
    offset = sb.st_size * 2;

    ret = lseek(fd, SEEK_SET, offset);
    printf("lseek(offset=%lu): ret = %d, errno = %d, %s\n", offset, ret, errno, strerror(errno));

    errno = 0;
    ret = pread(fd, buf, 512, offset);
    printf("pread(offset=%lu): ret = %d, errno = %d, %s\n", offset, ret, errno, strerror(errno));

    close(fd);

    return ret;
}
[root@ninja82 pread]$ ./test-pread 
size = 1554
lseek(offset=3108): ret = -1, errno = 22, Invalid argument
pread(offset=3108): ret = 0, errno = 0, Success
```

I will fix these things and submit again. Thanks!

adammoody: Thanks for the test, @sandrain.  That helps.  It surprised me a bit that lseek() was returning an error here.  I think that works as well, but we've got the argument order mixed in the test code above.  If I change:
```
ret = lseek(fd, SEEK_SET, offset);
```
to:
```
ret = lseek(fd, offset, SEEK_SET);
```
then it outputs:
```
size = 122398
lseek(offset=244796): ret = 244796, errno = 0, Success
pread(offset=244796): ret = 0, errno = 0, Success
```

client/src/unifycr-sysio.c
```diff
@@ -917,13 +917,18 @@ ssize_t UNIFYCR_WRAP(read)(int fd, void *buf, size_t count)
...
         unifycr_fd_t *filedesc = unifycr_get_filedesc_from_fd(fd);
-        if (filedesc == NULL) {
+
+        if (meta == NULL || filedesc == NULL) {
             /* ERROR: invalid file descriptor */
             errno = EBADF;
             return (ssize_t) (-1);
         }
 
+        if (filedesc->pos == meta->size)
```
adammoody: Modified the pread test code to try a normal read() after an lseek with pos past the end of the file, this should return 0 (success) as well, so I think this check should be:

```suggestion
        if (filedesc->pos >= meta->size)
```

client/src/unifycr-sysio.c
```diff
@@ -1523,13 +1546,23 @@ ssize_t UNIFYCR_WRAP(pread)(int fd, void *buf, size_t count, off_t offset)
...
+
+        if (meta == NULL || filedesc == NULL) {
             /* ERROR: invalid file descriptor */
             errno = EBADF;
             return (ssize_t) (-1);
         }
 
+        /*
+         * we should set errno conforming to lseek(2):
+         * for offset being larger than the file size, lseek(2) returns error
```
adammoody: I think it's ok in lseek, too

adammoody:
@sandrain, found a couple more based on the outcome of your handy test case.  Can you update once more, then I think we're there.  Thanks!

client/src/unifycr-sysio.c
sandrain: sorry, my bad :-)

adammoody:
Thanks, @sandrain !

PR #188, opened 2018-10-22
clean up numa m4, fix compile error in client when using numa

adammoody:
This fixes up the NUMA m4 script to set HAVE_LIBNUMA if detected.  It also updates the client/src/unifycr.c so that it uses HAVE_LIBNUMA and adds a missing char* declare for an env variable when using NUMA.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -51,7 +51,7 @@
 #include <mpi.h>
 #include <openssl/md5.h>
 
-#ifdef ENABLE_NUMA_POLICY
+#ifdef HAVE_LIBNUMA
 #include <numa.h>
 #endif
 
@@ -128,9 +128,7 @@ static size_t
 unifycr_spillover_size;  /* number of bytes in spillover to be used for chunk storage */
 long    unifycr_spillover_max_chunks; /* maximum number of chunks that fit in spillover storage */
 
-
-
-#ifdef ENABLE_NUMA_POLICY
+#ifdef HAVE_LIBNUMA
 static char unifycr_numa_policy[10];
 static int unifycr_numa_bank = -1;
 #endif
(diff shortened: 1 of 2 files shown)


adammoody:
Looking through our existing code, I don't think we're actually using libnuma for anything useful right now.  If that's the case, it's probably better to drop linking with libnuma at all.

I know that in CRUISE, this was used to be sure to store data in memory that is in the same NUMA domain as the MPI process.  I suppose that could still be useful if we're writing to blocks in shared memory.

Anyone know whether we're currently using or will need libnuma?

If so, I should update this PR further to only link with -lnuma on the lines where it's really needed (currently adds -lnuma to everything).  If not, we could drop linking -lnuma at all.

adammoody:
@CamStan, can you double check that this PR still links to the libnuma in a non-standard location, e.g., the libnuma installed by spack?

adammoody:
Maybe using it still in mercury branch:
https://github.com/LLNL/UnifyCR/blob/f61c005776dd4cf5b0ba93d6b2fdec1dc0e2f206/client/src/unifycr.c#L1315

CamStan:
@adammoody, yes, libnuma is now getting linked with the spack install as well as when spack is told to use the one already installed.

CamStan:
If it's not getting used by anything at the moment, should we have spack not automatically link libnuma as a dependency and only do so when the user tells spack they want it (i.e., `spack install unifycr+numa`)? Or wait until we know if we want to remove libnuma as a dependency altogether?

adammoody:
I'm thinking libnuma may still be useful once we add mercury -- still need to study the code more to convince myself.  It likely should be an optional dependency, since I think things will still run albeit perhaps more slowly without libnuma.  For now, I'll keep this PR as is.

@CamStan, I would say that you could either drop libnuma in spack or make it an optional dependency/variant.  It doesn't add anything to our current dev branch.  Perhaps it's needed for earlier releases?  You might want to test those.

CamStan:
@adammoody, there are a lot more lines of code using numa in v0.1.1, but still all within `#ifdef` blocks, so it builds just fine without numa. Also, our missing `char *env` is there: https://github.com/LLNL/UnifyCR/blob/6b237f6b1a188c488d6ca653b0c24c3b09500ca9/client/src/unifycr.c#L1734
I'll make it optional and defaulted to without.

PR #199, opened 2018-11-12
simplify and clean up server cmd handler

adammoody:
- Simplify server command handler code, add comments
- Simplify server debug stream code, add comments
- Add comments to client rpc definitions
- Pass app and client ids on thread control struct, drop cli_signature_t struct
- Define functions to allocate and free shared memory regions in server
- Define rm_cmd_exit() function for main thread to call to send command to request manager thread to shut down
- Rename rm_read_remote_data and rm_mread_remote_data functions, to rm_cmd_read and rm_cmd_mread

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr_clientcalls_rpc.h
+++ b/client/src/unifycr_clientcalls_rpc.h
@@ -13,75 +13,97 @@
 #include <mercury.h>
 #include <mercury_proc_string.h>
 
-    MERCURY_GEN_PROC(unifycr_mount_out_t,
-        ((int64_t)(max_recs_per_slice))\
-        ((int32_t)(ret)))
-    MERCURY_GEN_PROC(unifycr_mount_in_t,
-        ((uint32_t)(app_id))\
-        ((uint32_t)(local_rank_idx))\
-        ((uint32_t)(dbg_rank))\
-        ((uint32_t)(num_procs_per_node))\
-        ((uint32_t)(req_buf_sz))\
-        ((uint32_t)(recv_buf_sz))\
-        ((uint64_t)(superblock_sz))\
-        ((uint64_t)(meta_offset))\
-        ((uint64_t)(meta_size))\
-        ((uint64_t)(fmeta_offset))\
-        ((uint64_t)(fmeta_size))\
-        ((uint64_t)(data_offset))\
-        ((uint64_t)(data_size))\
-        ((hg_const_string_t)(external_spill_dir)))
-    DECLARE_MARGO_RPC_HANDLER(unifycr_mount_rpc)
+/* called by client to register with the server, client provides a
+ * structure of values on input, some of which specify global
+ * values across all clients in the app_id, and some of which are
+ * specific to the client process,
+ *
+ * server creates a structure for the given app_id (if needed),
+ * and then fills in a set of values for the particular client,
+ *
+ * server attaches to client shared memory regions, opens files
+ * holding spill over data, and launchers request manager for
+ * client */
+MERCURY_GEN_PROC(unifycr_mount_out_t,
+    ((int64_t)(max_recs_per_slice))
+    ((int32_t)(ret)))
+MERCURY_GEN_PROC(unifycr_mount_in_t,
+    ((uint32_t)(app_id))
+    ((uint32_t)(local_rank_idx))
+    ((uint32_t)(dbg_rank))
+    ((uint32_t)(num_procs_per_node))
+    ((uint32_t)(req_buf_sz))
+    ((uint32_t)(recv_buf_sz))
+    ((uint64_t)(superblock_sz))
+    ((uint64_t)(meta_offset))
+    ((uint64_t)(meta_size))
+    ((uint64_t)(fmeta_offset))
+    ((uint64_t)(fmeta_size))
+    ((uint64_t)(data_offset))
+    ((uint64_t)(data_size))
+    ((hg_const_string_t)(external_spill_dir)))
+DECLARE_MARGO_RPC_HANDLER(unifycr_mount_rpc)
 
+MERCURY_GEN_PROC(unifycr_unmount_out_t, ((int32_t)(ret)))
+MERCURY_GEN_PROC(unifycr_unmount_in_t,
+    ((hg_const_string_t)(external_spill_dir)))
+DECLARE_MARGO_RPC_HANDLER(unifycr_unmount_rpc)
 
-    MERCURY_GEN_PROC(unifycr_metaset_out_t, ((uint32_t)(ret)))
-    MERCURY_GEN_PROC(unifycr_metaset_in_t,
-        ((int32_t)(fid))\
-        ((int32_t)(gfid))\
-        ((hg_const_string_t)(filename)))
-    DECLARE_MARGO_RPC_HANDLER(unifycr_metaset_rpc)
+/* given a global file id and a file name,
+ * record key/value entry for this file */
+MERCURY_GEN_PROC(unifycr_metaset_out_t, ((uint32_t)(ret)))
+MERCURY_GEN_PROC(unifycr_metaset_in_t,
+    ((int32_t)(fid))
+    ((int32_t)(gfid))
+    ((hg_const_string_t)(filename)))
+DECLARE_MARGO_RPC_HANDLER(unifycr_metaset_rpc)
 
-    /*MERCURY_GEN_PROC(unifycr_metaget_out_t,
-                     ((unifycr_file_attr_t)(attr_val)) ((uint32_t)(ret)))*/
-    MERCURY_GEN_PROC(unifycr_metaget_out_t,
-                     ((int64_t)(st_size))\
-                     ((uint32_t)(ret))\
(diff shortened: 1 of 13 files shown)


adammoody:
This is building and it runs when I test by hand.  It's failing on the style checks.  For one, I don't want to squash this into a single commit to get it to pass, since the changes are semantically different.  Also, I'm not in love with some of the required style checks (we can talk about that on a future call).

So I'd like to propose we merge this even though it's failing the style checks.

That ok with folks?

Having said that, I'd like to know if there are certain stylistic changes people want me to do as a compromise.

Anyone have any strong opinions on style fixes I should make?

dsikich:
@adammoody looking at the errors one of them looks like a typo in the request manager code where there is an extra space "! " with a not symbol. And it there was an error saying "initialize" was misspelled. 

I do think we should absolutely have the style checks to enforce consistency across PR's/commits. Given how many errors/warnings this PR has I think we'd have to agree to just get rid of the style checker completely. If this is merged then if someone happens to touch a line that now has an error/warning it will show up in their future PR too. 

I don't know if there is a way to turn off specific style checks that we all agree don't make sense? I think that would be the ideal way to go.

CamStan:
I agree that we should have a should follow a style checker. 

That said, If there is a way to update what is checked for and there are things everyone agrees should be changed, I wouldn't be opposed to changing those. Until then though, I agree that given the number of errors/warnings in this PR, it’d be better to fix these as opposed to ditching the checker all together.

adammoody:
I did fix a few of those in a later commit in the PR, but of course the checker needs to me to squash everything to notice that change, and I don't want to do that.

I'm not entirely opposed to a style-checker, although I do prefer some particular styles that I find to be more readable/safer that our current checker restricts.  We could take some time on a call to discuss which style we want and then try to get the style checker playing nice.  I think the checker does have a lot of flags to customize things, but it's hard to know how closely it will line up with what we want.

A big issue we have right now is that all of our existing code base from cruise/burstfs uses a style that is invalid according to our style checker.  So simply moving a line of code (without changing it) causes the style check to fail.  That then establishes a big disincentive to clean anything up, e.g., by moving lines around.  Once we decide on a style we want to use, it'd be worth running a sweep across the entire code base to eliminate that problem.  For example, that's why we have some lines now like:

```
int foo = function();
if (foo)
```

and others like:

```
int foo = function();

if (foo)
```

cruise/burstfs uses the top style, the checker enforces the bottom.

dsikich:
@adammoody yeah, that is a good point. I've had to deal with similar issues when submitting PR's that follow the old code style. We've been having to deal with this for a while now in terms of getting stuck with fixing large portions of the code you touched because it followed the old code guidelines. It is probably worth a sweep so that it does not continue. Does it work to close and re-open the PR after you've made the changes? I've had issues with the style checker in the past too, and I've generally learned I have to remember to run the style checker locally before submitting the PR to avoid these kind of issues with making fixes and it not reflecting in Travis. 

That sounds promising that we can customize a lot of the flags. I think it is just important that we are consistent. I don't have particularly have strong feelings on `char* name` vs. `char *name.` I just prefer that it is declared the same throughout the code.

adammoody:
To list a few more since this is a good place to do it.

I find the following to be more readable:

```
if (thing1 == thing2 &&
    thing3 == thing4)
{
    then do stuff;
}
```

vs as required by check:
```
if (thing1 == thing2 &&
    thing3 == thing4) {
    then do stuff;
}
```

To me, in the bottom style, the body of the block blends in with the condition, which makes it hard to read.

For another:
```
/* multiline comment
 * describing a code block */
int foo = function();
```
vs as required by check:
```
/* multiline comment
 * describing a code block
 */
int foo = function();
```
I kind of prefer the top in this case, since it reduces line count, but I'm ok with the bottom.  The problem is that all of cruise/burstfs uses the top, so we should fix that if we settle on the bottom.

For safety, I prefer always requiring braces:
```
if (foo) {
  code;
}
```
vs:
```
if (foo)
  code;
```
I'v been bit too many times where another line was added to a block, but the braces were forgotten.  The code will compile and run, but it does the wrong thing... that new line is always exectuted instead of just when the condition is met... hard bug to catch.

Without braces, it also gets kind of ugly if there are various comments that should be inside in the block, e.g.,
```
if (foo) {
  /* comment about stuff */

  /* another comment */
  code;

  /* TODO: need another line after above */
}
```
vs:
```
if (foo)
  /* comment about stuff */

  /* another comment */
  code;

  /* TODO: need another line after above */
```

adammoody:
As far as this particular PR, if we want to enforce checks, I could look at breaking it into a set of PRs, one per commit.  I'd prefer to do that over squashing them into one.

dsikich:
@adammoody All of those make sense to me.

CamStan:
I'm fine with the changes you've suggested thus far @adammoody. I've been in the habit of making my comments like
```
/* 
 * multiline comment
 * describing a code block
 */
int foo = function();
```
which adds two additional lines. However, the checker accepts that. I'm happy to adjust my commenting habits if we want to change this.

Is the style checker the only reason we enforce our PR's being squashed into one commit?

CamStan:
If that's the case, would it work (and would it be acceptable) for this PR until we get out style checker updated to use `$TEST_CHECKPATCH_SKIP_FILES` to have the style checker bypass these files?
https://github.com/LLNL/UnifyCR/blob/13a8a327de9573895f194adff9744a0de70e1389/scripts/checkpatch.sh#L32-L38

adammoody:
> @CamStan it doesn’t require you to squash things into one commit. It just expects that each commit in the PR conforms to the style checker. It is best to run the style checker locally before pushing the commit to GitHub so that you avoid these issues with Travis.
> 
> If it is one commit I generally just commit a ‘fix’ and then rebase it. If you have multiple commits you just have to make sure you run the style checker script before you push to GitHub. But your PR can contain multiple commits as far as I know.

adammoody:
Whoops, hit the wrong button.  Meant to reply to Danielle mentioning that might be another option for me to fix up this PR (fix each commit in turn, squash it, and rebase later commits).

dsikich:
@CamStan Yeah, seems like that would work to have it skip those files if the important errors are resolved. At least until we have worked out which style checks we want to update.

adammoody:
For readability, I also prefer:
```
if (! foo)
```
vs:
```
if (!foo)
```
I find that extra space helps highlight that this condition includes a "not", whereas the bottom style takes a bit more effort to pick that up.  That is, I find that on a glance it's easy to miss the ! here:
```
if (foo->bar->hello->world)
if (!foo->bar->hello->world)
```
whereas these are clearly different:
```
if (foo->bar->hello->world)
if (! foo->bar->hello->world)
```

MichaelBrim:
@adammoody Can we please create a separate issue to track/documennt the desired style checks? It would be nice for somebody to create an example of the alternative styles for each currently enabled check, and then maybe we could just do a doodle poll to decide. FWIW, I generally agree with your positions on most of these (except for brace placement - I'm in the K&R camp for that)

CamStan:
I agree with @MichaelBrim on having a separate issue to list out and take a poll to decide.

However, I haven't done much searching, but at first glance, I'm not sure how customizable `checkpatch.pl` is (the [documentation](http://heim.ifi.uio.no/~knuto/kernel/4.14/dev-tools/run-checkpatch.html) I found made it sound like they'd rather users conform to their checks).

`scripts/linux_kernel_checkpatch/checkpatch.pl --list-types` shows all the message types, but from what it looks like we're doing in `scripts/checkpatch.sh`, you can simply choose to ignore these if you don't want them checked.

That [documentation](http://heim.ifi.uio.no/~knuto/kernel/4.14/dev-tools/run-checkpatch.html) does make it look like maybe some things are customizable (like line length) and the `checkpatch.pl` file has an option for a configuration file https://github.com/LLNL/UnifyCR/blob/13a8a327de9573895f194adff9744a0de70e1389/scripts/linux_kernel_checkpatch/checkpatch.pl#L52 but no luck so far finding docs on how to use it.

We could always edit the `checkpatch.pl` file itself, but that could get messy for everything we'll want to change. https://github.com/LLNL/UnifyCR/blob/13a8a327de9573895f194adff9744a0de70e1389/scripts/linux_kernel_checkpatch/checkpatch.pl#L3073-L3080

I'm not familiar with checkpatch though, so maybe somebody else with more experience with it will know. Just want to make sure we even can choose how styles are checked before we get too far.

adammoody:
Yeah, it looks like it provides and --ignore option where you you eliminate certain checks.  In the case of the BLOCK_COMMENT_STYLE, I like the rest of the check it has, but we could hack the perl to comment out the one that looks for a */ on a trailing line for example.  I'll see what I can dig up.  It might just amount to a few changes, and I can live with some style checks that annoy me.

adammoody:
I'd say the best option would be to update check patch to check what we decide on.  However, that's assuming we can or have the time to figure out the perl foo needed to do that.  As a fallback, I'd say it's fine to just carve out particular checks it's currently enforcing that go against what we decide is ok.

I've got a test commit in there against checkpatch.pl to verify that we can do the latter.

PR #201, opened 2018-11-13
drop pkgconfig check for leveldb

adammoody:
pkgconfig was having trouble finding leveldb.  Switch to use similar approach to other dependencies.


CamStan:
@adammoody, does this mean we'll no longer have a dependency on pkgconfig? Or is it still needed?

.travis.yml
```diff
@@ -44,7 +44,9 @@ before_cache:
 script:
   # Force git to update the shallow clone and include tags so git-describe works
   - git fetch --unshallow --tags
-  - sh autogen.sh && ./configure --with-flatbuffers=$HOME/flatbuffers && make -k && make DISTCHECK_CONFIGURE_FLAGS="--with-flatbuffers=$HOME/flatbuffers" distcheck
+  - sh autogen.sh
+  - ./configure --with-flatbuffers=$HOME/flatbuffers || cat config.log
```
CamStan: The flatcc package got merged into Spack yesterday, so this should be available in their latest develop branch.

adammoody:
We might be able to drop package config.  In our build, we were using pkgconfig to link against leveldb.  That's no longer needed with this patch.

However, I see we're also installing a pkgconfig file as part of the unifycr install.  I suppose that step could still require pkgconfig, but I suspect it might not.  Worth a shot to drop it as a spack dependency and give it a shot.

Good to know about flat buffers.  I'll test that out and update our build steps if it's working.

adammoody:
I've got a PR in the wings to switch to the flatcc spack package.

adammoody:
Ok, updated to use the flatcc spack package, too.  Thanks, @CamStan .

PR #204, opened 2018-11-13
detect and report read errors

adammoody:
This adds detection and reporting of read errors for the case where the service manager thread fails to read data (e.g., backing file was deleted or SSD failed).  This requires:
- adding code to check for errors or short reads (treated as errors) on the service manager
- reporting them through a newly added error field from service manager to request manager
- reporting them through a newly added field from request manager to client in shared memory
- reporting them to the application from the client wrappers.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr-sysio.c
+++ b/client/src/unifycr-sysio.c
@@ -930,27 +930,33 @@ ssize_t UNIFYCR_WRAP(read)(int fd, void *buf, size_t count)
         if (filedesc->pos >= meta->size)
             return 0;   /* EOF */
 
-        /* read data from file */
-        size_t retcount;
+        /* assume we'll succeed in read */
+        size_t retcount = count;
 
         read_req_t tmp_req;
-        tmp_req.fid    = fd + unifycr_fd_limit;
-        tmp_req.offset = filedesc->pos;
-        tmp_req.length = count;
-        tmp_req.buf    = buf;
+        tmp_req.fid     = fd + unifycr_fd_limit;
+        tmp_req.offset  = filedesc->pos;
+        tmp_req.length  = count;
+        tmp_req.errcode = UNIFYCR_SUCCESS;
+        tmp_req.buf     = buf;
 
         /*
          * this returns error code, which is zero for successful cases.
          */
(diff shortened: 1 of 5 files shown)


adammoody:
Depends on https://github.com/LLNL/UnifyCR/pull/199

adammoody:
This is ready for review.

NOTE: We are not correctly reporting read errors in the lio_listio() wrapper, but the aio wrappers in general need some real attention, so the error reporting is deferred to that task.

client/src/unifycr-sysio.c
```diff
@@ -1038,16 +1044,25 @@ int UNIFYCR_WRAP(lio_listio)(int mode, struct aiocb *const aiocb_list[],
...
     }
 
     int ret = unifycr_fd_logreadlist(reqs, nitems);
 
+    for (i = 0; i < nitems; i++) {
+        /* TODO: update fields to record error status
+         * see /usr/include/aio.h,
+         * update __error_code and __return_val fields? */
+    }
+
```
dsikich: @adammoody I'm confused about what this loop does exactly? Looks like it isn't functional right now, but would this be updating the errcode field you added to the structs?

adammoody: Long story on that non-functioning loop... my intent was to set the error codes in each of the aiocb structs based on the errors in the corresponding read requests that were passed to and modified by fd_logreadlist.  However, I realized it was not straight-forward.  With aio calls, one should call aio_error() and then aio_return() to finally get the return code.  The easiest thing would be to save the error code on a field in the aiocb struct, but it's not clear which fields on the aiocb struct are portable for that use.  Because it was going to take a lot more thought, I decided to defer that to a later PR.  I left this loop in there as a reminder as one of the things that needs to be looked at.  I thought I created an issue, too, but I'm not seeing it.  I'll do that if I never did.

adammoody: Created an issue to address aio wrappers in general, which will include this fix: https://github.com/LLNL/UnifyCR/issues/225

adammoody:
Thanks @dsikich

PR #209, opened 2018-11-14
use MPI_Comm_split_type to compute rank within a node

adammoody:
### Description
This replaces the code used to compute the number of ranks on a node and the calling processes rank within its node.  It uses MPI_Comm_split_type() to accomplish the task instead of the gather-to-root, compute on root, scatter approach.

It's simpler and it should be faster.  However, it may be less portable, as this function was not added until MPI-3.0.

If we need both performance, simplicity, and portability, we can use the rankset component from ECP Veloc.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr-internal.h
+++ b/client/src/unifycr-internal.h
@@ -318,7 +318,6 @@ read_req_set_t read_req_set;
 read_req_set_t tmp_read_req_set;
 index_set_t tmp_index_set;
 
-extern int *local_rank_lst;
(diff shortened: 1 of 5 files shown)


adammoody:
This is currently failing because MPI_Comm_split_type() is not defined in the Open MPI rpm that we're installing on travis (it's too old).  We can fix this either by using a newer MPI (install via spack) or by using the rankstr component from veloc.

PR #210, opened 2018-11-15
Add check for return value of unifycr_mount

boehms:
Make sure the test exits if it encounters an error mounting unifycr.


MichaelBrim:
@boehms this has a build failure due to redefinition of 'int ret', check the travis log

PR #213, opened 2018-11-16
add fid field to file descriptor structure, add free fd stack

adammoody:
This adds a fid field to the file descriptor structure to track the underlying local file id that an open file descriptor is associated with.  It also adds a stack to track free fd values, a free fd value is popped from the stack on open() and pushed back on close().

It similarly adds a stack to track free FILE* streams, allowing a process to fopen() the same file more than once.

It accomplishes the following:
- enables a process to have more than one active file descriptor open to the same underlying file
- adds state to a file descriptor to know when it's active or not (e.g., fixes double close bug)
- enables a process to have more than one active FILE* stream open to the same underlying file
- enables a process to have more than one active DIR* stream open to the same underlying directory
- https://github.com/LLNL/UnifyCR/issues/206

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr-dirops.c
+++ b/client/src/unifycr-dirops.c
@@ -15,55 +15,80 @@
 
 #include "unifycr-sysio.h"
 
-/*
- * TODO: the number of open directories clearly won't exceed the number of file
- * descriptors. however, the current UNIFYCR_MAX_FILEDESCS value of 256 will
- * quickly run out. if this value is fixed to be reasonably larger, then we
- * would need a way to dynamically allocate the dirstreams instead of the
- * following fixed size array.
- */
-static unifycr_dirstream_t _dirstreams[UNIFYCR_MAX_FILEDESCS];
-
-/*
- * TODO: currently, we just take the same indexed slot to our fid.
- */
-#if 0
-static pthread_spinlock_t _dirstream_lock;
-static unifycr_dirstream_t *_dirstream_freelist;
-
-static unifycr_dirstream_t *unifycr_dirstream_alloc(void)
-{
-}
-
-static void unifycr_dirstream_free(unifycr_dirstream_t *dirp)
-{
-}
-#endif
-
+/* given a file id corresponding to a directory,
+ * allocate and initialize a directory stream */
 static inline unifycr_dirstream_t *unifycr_dirstream_alloc(int fid)
 {
-    unifycr_dirstream_t *dirp = &_dirstreams[fid];
+    /* allocate a file descriptor for this stream */
+    int fd = unifycr_stack_pop(unifycr_fd_stack);
+    if (fd < 0) {
+        /* exhausted our file descriptors */
+        errno = EMFILE;
+        return NULL;
+    }
 
+    /* allocate a directory stream id */
+    int dirid = unifycr_stack_pop(unifycr_dirstream_stack);
+    if (dirid < 0) {
+        /* exhausted our directory streams,
+         * return our file descriptor and set errno */
+        unifycr_stack_push(unifycr_fd_stack, fd);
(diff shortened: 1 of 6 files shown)


adammoody:
Have some outstanding cases where fd/fid are used interchangeably, which needs to be updated with this PR.  Ongoing...

CamStan:
@adammoody, as far as adding the test you suggested, I created a new issue #215 for a place to suggest future wrapper unit tests since we're going to have a lot to write and they should probably be logically categorized into which call(s) they are testing (I've adjusted our testing framework to hopefully accommodate for this).

adammoody:
Great.  Thanks, @CamStan.  I'll get the tests added back.

I know I need to make another pass or two through the wrapper set to look for new fd/fid bugs that this PR introduces.  I know I've got some in the directory wrappers.

We probably should wrap dup if we can.

adammoody:
We have a set of dirstreams we could extend in the same way in unifycr-dirops.c

client/src/unifycr-dirops.c
```diff
@@ -115,78 +143,124 @@ DIR *UNIFYCR_WRAP(opendir)(const char *name)
...
+    }
 }
 
 int UNIFYCR_WRAP(closedir)(DIR *dirp)
 {
-    fprintf(stderr, "Function not yet supported @ %s:%d\n", __FILE__, __LINE__);
-    errno = ENOSYS;
-
-    return -1;
+    if (unifycr_intercept_dirstream(dirp)) {
```
CamStan: Don't have any tests for these dirops wrappers yet (granted most of them are "not yet supported"). However, as they get implemented, we may end up needing to check if `dirp` is a valid value as calls like `closedir` can return `EBADF` and others like `dirfd`can return `EINVAL`.

client/src/unifycr-sysio.c
```diff
@@ -158,41 +169,49 @@ int UNIFYCR_WRAP(rename)(const char *oldpath, const char *newpath)
...
         /* verify that we don't already have a file by the new name */
-        if (unifycr_get_fid_from_path(newpath) < 0) {
-            /* check that new name is within bounds */
-            size_t newpathlen = strlen(newpath) + 1;
-            if (newpathlen > UNIFYCR_MAX_FILENAME) {
-                errno = ENAMETOOLONG;
+        int newfid = unifycr_get_fid_from_path(newpath);
+        if (newfid >= 0) {
+            /* something exists in newpath, need to delete it */
+            int ret = UNIFYCR_WRAP(unlink)(newpath);
```
CamStan: Not sure if it's an issue, but the compiler is issuing a warning on this:
```
../../../client/src/unifycr-sysio.c:201:13: warning: implicit declaration of function ‘__wrap_unlink’ [-Wimplicit-function-declaration]
             int ret = UNIFYCR_WRAP(unlink)(newpath);
             ^
```

adammoody: Nice catch.  That looks worrisome.  I'll dig into that.

adammoody: Found the problem with UNIFYCR_WRAP here, will fix that in a separate PR.

client/src/unifycr-dirops.c
adammoody: Good point.  The intercept should return if the address the user gives is in range.  We should double check that the structure is still in use after that.

adammoody:
@CamStan you were right about not approving this yet.  Mind doing that when you get a chance.  If the intercept functions are not checking that things are valid, I'll submit another PR or at least open an issue for that.

adammoody:
Thanks @CamStan .  Also fixed up the conflict.

PR #214, opened 2018-11-16
Change the domain socket file name to include uid.

boehms:
If the UnifyCR server does not shut down correctly, the domain socket is
left in /tmp. This can cause issues on shared resources.
Attaching the user id to the file should fix collisions with sockets
left behind by other users.

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -2281,8 +2281,8 @@ static int unifycr_init_socket(int proc_id, int l_num_procs_per_node,
         (l_num_procs_per_node == 1) &&
         (l_num_del_per_node == 1)) {
         // use zero-index domain socket path
-        snprintf(tmp_path, sizeof(tmp_path), "%s0",
-                 SOCKET_PATH);
+        snprintf(tmp_path, sizeof(tmp_path), "%s.%d",
+                 SOCKET_PATH, getuid());
 
 #ifdef HAVE_PMIX_H
         // lookup domain socket path in PMIx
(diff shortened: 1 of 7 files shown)


adammoody:
Thanks @boehms .  It looks like you're also dropping the support for being able to run multiple servers on a node with this change, not sure if you intended that or not.  We could add the uid and still keep the server id, as well.

With burstfs, I seem to recall that we found that it helps to run multiple servers per node if there are a lot of clients.  We used the server id so that multiple servers could run on a node.  However, that math might change with our switch to margo.

MichaelBrim:
> Thanks @boehms . It looks like you're also dropping the support for being able to run multiple servers on a node with this change, not sure if you intended that or not. We could add the uid and still keep the server id, as well.
> 
> With burstfs, I seem to recall that we found that it helps to run multiple servers per node if there are a lot of clients. We used the server id so that multiple servers could run on a node. However, that math might change with our switch to margo.

@adammoody We talked about this a while back on a conference call. The consensus was to make everything work with one server/delegator per node for now, without eliminating the path back to multiple delegators. There are many other bits (e.g., PMIx integration, runstate file generation, ...) that would fail right now if someone tried to run with multiple servers. Maybe we need an issue to track these bits?

dsikich:
@adammoody Joseph and I brought up this issue a while back when we were converting the code to rpc as well. We decided to not worry about supporting multiple servers on a node as a first pass. I think this should be tracked in a separate issue if we want to add support for that later.

PR #217, opened 2018-11-19
hack check patch to be more permissive

adammoody:
Change checkpatch script to allow for styles decided on in https://github.com/LLNL/UnifyCR/issues/212

This does not add support for enforcing these new styles (cause that's hard, and it's in perl), it only relaxes the checks so that the agreed-upon style is allowed.  Because it likely now allows either/any style on some of these checks, we'll just have to look for those in code review.

Most of the voting was unanimous.

The braces always-at-end had more support than the extra line, so I kept that.  I also kept the check that there is no space betwen ! and a variable name.  It had one more vote in favor of allowing the space, but one of those is mine, and I suppose it's kind of unconventional so I'll change my over.

### Types of changes

### Checklist:


adammoody:
@boehms , can you check whether this change enables your PR to pass?  I didn't test that particular case.

PR #219, opened 2018-11-28
Add wrapper support for __open_2

dsikich:
Some versions of glibc rename the open call to __open_2, which
will not allow UnifyCR to intercept the call. This wrapper just
calls the UnifyCR implementation of open we already have.

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -2268,6 +2268,11 @@ static int unifycr_client_rpc_init(char* svr_addr_str,
                        unifycr_mount_in_t,
                        unifycr_mount_out_t, NULL);
 
+    (*unifycr_rpc_context)->unifycr_unmount_rpc_id   =
+        MARGO_REGISTER((*unifycr_rpc_context)->mid, "unifycr_unmount_rpc",
+                       unifycr_unmount_in_t,
+                       unifycr_unmount_out_t, NULL);
+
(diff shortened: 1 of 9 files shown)


dsikich:
@adammoody @CamStan I suppose we can just wait to merge this one until the new changes are reflected in m2 for check patch since we agreed to change the brace rules for one line if/else blocks. This PR isn't urgent unless someone is using a version of glibc that renames open to __open_2.

client/src/unifycr-sysio.c
```diff
@@ -790,6 +790,44 @@ int UNIFYCR_WRAP(open64)(const char *path, int flags, ...)
     return ret;
 }
 
+int UNIFYCR_WRAP(__open_2)(const char *path, int flags, ...)
+{
+    DEBUG("__open_2 was called for fd %d....\n", fd);
```
adammoody: No fd defined here, mean to report the path name?

dsikich: @adammoody thanks. yeah i copied that debug statement from another wrapper and missed that.

CamStan:
@dsikich, #221 should fix the style checks that will allow this to pass.

dsikich:
@CamStan Awesome, thanks!

client/src/unifycr-sysio.c
dsikich: @adammoody I updated/fixed the debug statement here.

Issue #212, opened 2018-11-15
Coding style and style checks


adammoody:
This ticket is to discuss code style options.  We can place each suggestion in its own comment block.  Of particular interesting is deciding on a style that is consistent with the existing code base and/or bring the existing code base into compliance.

CamStan:
Can always use the *thumbs up* or *thumbs down* reactions on the top right of each comment to vote for the first or second suggestion.

CamStan:
Suggested style:
```
int foo = function();
if (foo)
```
Checkpatch style:
```
int foo = function();

if (foo)
```

CamStan:
Suggested style:
```
if (thing1 == thing2 &&
    thing3 == thing4)
{
    then do stuff;
}
```
Checkpatch style:
```
if (thing1 == thing2 &&
    thing3 == thing4) {
    then do stuff;
}
```

CamStan:
Suggested style:
```
/* multiline comment
 * describing a code block */
int foo = function();
```
Checkpatch style:
```
/* multiline comment
 * describing a code block
 */
int foo = function();
```

CamStan:
Suggested style:
```
if (foo) {
  code;
}
```
Checkpatch style:
```
if (foo)
  code;
```

CamStan:
Suggested style:
```
if (! foo)
```
Checkpatch style:
```
if (!foo)
```

MichaelBrim:
Note I interpreted "thumbs-up" as voting for suggested style.

adammoody:
Thanks, @CamStan for listing all of these.

I'd say the best option would be to update check patch to check what we decide on. However, that's assuming we can or have the time to figure out the perl foo needed to do that. As a fallback, I'd say it's fine to just carve out particular checks it's currently enforcing that go against what we decide is ok.

I've got a test commit in there against checkpatch.pl to verify that we can do the latter:

baa024b

This hack does the following:
- allows */ to be at end of comment line, rather than requiring it to be on a separate line
- allows use of a variable to follow its declaration without requiring an empty line in between
- allows a space after unary !
- allow braces to exist around if/while/for blocks even for bodies with a single line

adammoody:
Assuming we're ok with that, I've updated https://github.com/LLNL/UnifyCR/pull/199 to fix the rest of the spots where it was failing style checks.

MichaelBrim:
Suggested style:

```
if (CONSTANT == variable)
```

Checkpatch style:

```
[...]

PR #221, opened 2018-11-29
Change checkpatch to enforce changed styles

CamStan:
### Description
This change enforces pointers to be declared with the * next to the
data type (i.e., char* foo instead of char *foo) and it also
enforces all single statement if and if/else blocks to have braces.

### Motivation and Context
Desired style check changes discussed in issue #212.

Left the constant comparisons as being ignored since we have a lot of the variable-then-constant in the code. Can always changed this to be enforced if we want as well.

### How Has This Been Tested?
Created several cases of what was originally being caught by checkpatch and what wasn't and tested this until they were reversed.

### Types of changes


adammoody:
Looks good to me.  I'd like to get a few PRs ahead of this merged just to save avoid having to redo those.  We can talk about how to sweep through the code and update all stars using astyle or clang format in our next meeting.

CamStan:
@adammoody, sounds good. I think #219 is the only current PR that will need to wait for this change in order to pass.

adammoody:
Thanks @CamStan

PR #223, opened 2018-12-04
replace fprintf to stderr with LOGERR macro

adammoody:
The compiler warned that some fprintf calls were missing their format string.  Rather than add that, I replaced these calls with calls to the new LOGERR macro.  This macro just currently calls fprintf(stderr, ...), however, eventually we can redefine this macro to create a common error message format to include things like version info, file, and line numbers where the error was printed.  It also can add the newline and print to stderr, so something like:

```fprintf(stderr, "ERROR: Something bad happened @ %s:%d\n", __FILE__, __LINE__);```

becomes:

```LOGERR("Something bad happened");```

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -1831,7 +1831,7 @@ static int unifycr_init(int rank)
          * a void* in an fpos_t so check that there's room and at least
          * print a message if this won't work */
         if (sizeof(fpos_t) < sizeof(void *)) {
-            fprintf(stderr, "ERROR: fgetpos/fsetpos will not work correctly.\n");
+            LOGERR("fgetpos/fsetpos will not work correctly");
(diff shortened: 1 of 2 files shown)


dsikich:
@adammoody seems pretty straight forward to me. Although like you said it would be nice to have the file and line number in there too for reporting errors. 

The main thing I was wondering is if there is a reason "LOGERR" is not spelled correctly? If not, we should probably fix that.

adammoody:
Thanks @dsikich .  LOGERR is short for LOGERROR, but I was just saving some key strokes.  We might also add things like LOGWARN and replace DEBUG with LOGDBG.

We can add the file and line numbers to the macro, but I was reserving that for a separate PR.

Issue #226, opened 2018-12-05
update examples to call unmount on all ranks


dsikich:
Some examples currently only call unmount on rank 0, but we need to call unmount on all ranks to properly disconnect that client from the server.

CamStan:
Fixed by #236

PR #227, opened 2018-12-05
Add unmount rpc

dsikich:
Adds unmount rpc that shuts down delegator thread, detaches from
request and read shared memory buffers, and closes the socket
that was opened during mount. This commit also adds a
sanitize_sock_cli function in order to only close the socket
that corresponds to that specific client id. The sock_sanitize
function closes all of the client sockets and the main socket,
which is not what we needed for a client disconnect call.

I will update the examples based on issue #226 in another PR.

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -2239,65 +2239,19 @@ int unifycr_mount(const char prefix[], int rank, size_t size,
 }
 
 /**
- * unmount the mounted file system, triggered
- * by the root process of an application
- * ToDo: add the support for more operations
- * beyond terminating the servers. E.g.
- * data flush for persistence.
+ * unmount the mounted file system
+ * TODO: Add support for unmounting more than
+ * one filesystem.
  * @return success/error code
  */
 int unifycr_unmount(void)
 {
-#if 0
-    int cmd = COMM_UNMOUNT;
-    int bytes_read = 0;
-    int rc;
-    int *response = NULL;
+    int ret;
 
-    memset(cmd_buf, 0, sizeof(cmd_buf));
-    memcpy(cmd_buf, &cmd, sizeof(int));
-
-    rc = __real_write(cmd_fd.fd, cmd_buf, sizeof(cmd_buf));
-    if (rc <= 0)
-        return UNIFYCR_FAILURE;
-
-    cmd_fd.events = POLLIN | POLLPRI;
-    cmd_fd.revents = 0;
-
-    rc = poll(&cmd_fd, 1, -1);
-    if (rc < 0)
-        return UNIFYCR_FAILURE;
-
-    if (cmd_fd.revents != 0 && cmd_fd.revents == POLLIN) {
-        bytes_read = __real_read(cmd_fd.fd, cmd_buf, sizeof(cmd_buf));
-
-        response = (int *) cmd_buf;
-
-        if (bytes_read <= 0 ||
-            response[0] != COMM_UNMOUNT || response[1] != ACK_SUCCESS)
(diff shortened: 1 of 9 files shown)


server/src/unifycr_cmd_handler.c
```diff
@@ -423,6 +423,61 @@ static void unifycr_mount_rpc(hg_handle_t handle)
...
+                     app_config->recv_buf_sz,
+                     &(app_config->shm_recv_bufs[client_id]));
+    app_config->shm_recv_bufs[client_id] = NULL;
+
+    /* destroy the sockets except for the ones for acks */
+    sock_sanitize_cli(client_id);
+
+    /* build output tructure to return to caller */
+    unifycr_mount_out_t out;
+    out.ret = ret;
```
adammoody: Looks like ret is last set from the HG_Get_input call here.  I think we should return something like UNIFYCR_SUCCESS / FAILURE instead.  For now, we could just have it always return SUCCESS, and then add logic later to conditionally return FAILURE or error code if the unmount operation failed for some reason.

If other rpc handlers are similar, let's create an issue and clean this up in another PR.

client/src/unifycr_client.c
```diff
@@ -70,6 +70,40 @@ uint32_t unifycr_client_mount_rpc_invoke(unifycr_client_rpc_context_t** unifycr_
...
+    hret = margo_forward(handle, &in);
+    assert(hret == HG_SUCCESS);
+
+    /* decode response */
+    hret = margo_get_output(handle, &out);
+    assert(hret == HG_SUCCESS);
+
+    margo_free_output(handle, &out);
+    margo_destroy(handle);
+    return out.ret;
```
adammoody: Here, I suppose we should save the out.ret value to a temporary variable that is returned.  It probably works right now, but I'm worried about the sequence here where we free_output(&out) and then later use a field from that struct.  Maybe change to something like:

```
int ret = out.ret;
margo_free_output(handle, &out);
return ret;
```

I see that pattern exists in some other invoke calls too, so we can clean this up in a different PR.

server/src/unifycr_cmd_handler.c
dsikich: @adammoody Makes sense. I updated to return UNIFYCR_SUCCESS.

client/src/unifycr_client.c
dsikich: @adammoody Thanks, this has been updated.

dsikich:
@adammoody I also added the unmount_rpc invocation to the unifycr_mount call.

client/src/unifycr.c
```diff
@@ -2080,47 +2080,19 @@ int unifycr_mount(const char prefix[], int rank, size_t size,
...
-    }
-#endif
+    int ret;
 
-    return UNIFYCR_SUCCESS;
+    /* invoke unmount rpc */
+    printf("calling unmount\n");
+    ret = unifycr_client_unmount_rpc_invoke(&unifycr_rpc_context);
+    return ret;
 }
```
adammoody: Just occurred to me that we should have the client also tear down its end of the connection to the server here, detach from its request and receive shared memory regions and close its socket.

adammoody: @dsikich based on your email, it possible that this logic was never in the client.  In that case, it will take some effort to add it.  We could probably finalize this PR and create an issue to track the fact that we should clean up the connection on the client side.

dsikich:
@adammoody That is fine with me. I can add it to another PR.

dsikich:
@adammoody fixed.

PR #228, opened 2018-12-05
Add __open_2 wrapper

dsikich:
In some versions of glibc the open call gets translated to __open_2.
This adds a wrapper for __open_2 so that we do not miss the open
call in those versions of glibc.


adammoody:
@dsikich the checkpatch PR has now been merged into m2

dsikich:
@adammoody @CamStan Thanks, passing now that the new check patch is merged.

client/src/unifycr-sysio.c
```diff
@@ -789,6 +789,43 @@ int UNIFYCR_WRAP(open64)(const char *path, int flags, ...)
...
+
+    /* check whether we should intercept this path */
+    if (unifycr_intercept_path(path)) {
+        DEBUG("__open_2 was intercepted for path %s....\n", path);
+
+        /* Call open wrapper with LARGEFILE flag set*/
+        if (flags & O_CREAT) {
+            ret = UNIFYCR_WRAP(open)(path, flags | O_LARGEFILE, mode);
+        } else {
+            ret = UNIFYCR_WRAP(open)(path, flags | O_LARGEFILE);
```
adammoody: Checking this one more time, I believe we want to drop the O_LARGEFILE flag here and update the comment about LARGEFILE.  That flag is just for open64, and I think open_2 is a replacement for open() which does not imply that flag.

dsikich: @adammoody yeah good catch. I updated this.

PR #229, opened 2018-12-05
fix: missing declare for wrapper prototype in static build

adammoody:
The compiler did not have the wrapper function prototype in the static build.  As a result, if a wrapper had to call another wrapper, we'd end up with a warning like the following:

```
../../../client/src/unifycr-sysio.c:201:13: warning: implicit declaration of function ‘__wrap_unlink’ [-Wimplicit-function-declaration]
             int ret = UNIFYCR_WRAP(unlink)(newpath);
             ^
```

This is a one line fix that adds that.  Looks like the line might have been dropped along the way, given the continuation character on the live above the fix.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr-internal.h
+++ b/client/src/unifycr-internal.h
@@ -183,9 +183,11 @@ do { \
 
 /* we don't need a variable to record the address of the real function,
  * just declare the existence of __real_open so the compiler knows the
- * prototype of this function (linker will provide it) */
+ * prototype of this function (linker will provide it), also need to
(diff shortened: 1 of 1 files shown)


adammoody:
I also double checked that -DUNIFYCR_GOTCHA is being defined when including the header defining these macros.  It was the static build that was throwing the error, not the gotcha build.  I was tricked because the last file to be listed was a gotcha file due to the parallel build, but that file compiled without error.  It was the static build, further up in the output that died.

PR #231, opened 2018-12-06
style: unifycr_init

adammoody:
Ensure that source files match checkpatch requirements.

### Types of changes

### Checklist:


client/src/unifycr-wrap.h
```diff
@@ -49,8 +49,8 @@
  * POSIX wrappers: file descriptors
  * --------------------------------------- */
 
-UNIFYCR_WRAP_DECL(open, int, (const char *path, int flags, ...));
-UNIFYCR_WRAP_DECL(write, ssize_t, (int fd, const void *buf, size_t count));
+UNIFYCR_WRAP_DECL(open, int, (const char* path, int flags, ...));
```
CamStan: @adammoody, double check me on this, but you might be able to just delete this file. I don't think we have a `UNIFYCR_WRAP_DECL` macro anymore and these functions are already wrapped in client/src/unifycr-sysio.h. I think this file might be left over from something in the initial days of the project as it hasn't been touched in a year.

adammoody: Ok, I'll take a look.  Thanks for looking at that.  I was just on auto-pilot running through the files here.

adammoody:
Ok, have gone through the server and client files, except for client files generated by flatcc and the uthash and utlist files.

Also, the meta/src files have their own style different from ours.  I don't think we should change those, since otherwise it'll be hard to do diffs against MDHIM code.

That leaves the examples and test files.  I'll coordinate with @CamStan on those.

@dsikich this will likely conflict with your unmount PR.  Not sure which will be easiest to merge first.  Will check with you on that.

CamStan:
@adammoody, with #232 getting merged, we're you wanting to hit the files in t/ with this? Or want me to do them in a separate PR?

adammoody:
@CamStan, looking at the t/src files, they look to be pretty clean already.  Since you've largely touched every line with your PR, so we might already declare those as done.  Were there any source files that you think we still need to do?  If so, I could tackle those.

CamStan:
@adammoody, true, I pretty much already go most of those. The only ones I'm seeing that might still be good to do are the files in t/lib.

CamStan:
@adammoody, and potentially the ones in util/ unless you skipped those on purpose. Otherwise that's it.

adammoody:
Looks like t/lib is the original TAP code, so I'll leave that alone.  I missed the util/ directory.  Will get those now.

CamStan:
t/lib/testutil.c is ours, but I may end up adding to it in the future if we end up needing more convenience functions for testing. I can do it then if need be.

Looks great then. You're a machine!

adammoody:
Thanks, @CamStan.  Added t/lib/testutil.{c,h} in there too.

common/src/unifycr_const.h
```diff
@@ -105,14 +105,16 @@
...
 // #define NUM_OF_READ_TASKS 65536
 // #define PORT_STR_LEN 10
 // #define SH_BUF_SIZE 1048576
-// #define SERVICE_MEM_POOL_SIZE 2147483648 /* buffer size on the service manager for read clustering and pipelining */
+// /* buffer size on the service manager for read clustering and pipelining */
+// #define SERVICE_MEM_POOL_SIZE 2147483648
 
 /* ****************** Metadata/MDHIM Default Values ******************** */
-#define META_DEFAULT_DB_PATH /tmp
+#define META_DEFAULT_DB_PATH (/tmp)
```
CamStan: This jumped out at me. Making sure it's intentional.

client/src/unifycr.c
```diff
@@ -147,15 +147,15 @@ int unifycr_initialized = 0;
...
-void *unifycr_superblock = NULL;
-static void *free_fid_stack = NULL;
-void *free_chunk_stack = NULL;
-void *free_spillchunk_stack = NULL;
-unifycr_filename_t *unifycr_filelist    = NULL;
-static unifycr_filemeta_t *unifycr_filemetas   = NULL;
-static unifycr_chunkmeta_t *unifycr_chunkmetas = NULL;
-
-char *unifycr_chunks = NULL;
+void* unifycr_superblock;
```
CamStan: Looks like `NULL` was removed from variables that are declared and initialized at the same time (here and in other places). Not sure if that is a problem or was intended.

common/src/unifycr_const.h
adammoody: Yeah, that was forced by checkpatch.

client/src/unifycr.c
adammoody: checkpatch requires that all "=NULL" initializations for global pointers to be dropped.

common/src/unifycr_const.h
dsikich: https://docs.freebsd.org/info/cpp/cpp.info.Macro_Parentheses.html

adammoody: Well, I should have tested that better.  Adding the parens leads to a runtime error.  Trying quotes instead.

client/src/unifycr.c
adammoody: Well quotes need some help, too...
```
+ /g/g0/moody20/projects/unifycr2.git/install/bin/unifycrd
# accepting RPCs on address "na+sm://146417/0"
completed sock init server
MLOG CRIT Error opening leveldb database, abc..., path is "/tmp"/unifycr_db-0-0
MLOG CRIT MDHIM Rank: 0 - Error while opening database
MLOG CRIT Rank: 0 - Error opening data store for index: 0
MLOG CRIT Error opening leveldb database, abc..., path is "/tmp"/unifycr_db-1-0
MLOG CRIT MDHIM Rank: 0 - Error while opening database
MLOG CRIT Rank: 0 - Error opening data store for index: 1
```

common/src/unifycr_const.h
adammoody: Well quotes need some help, too...
```
+ /g/g0/moody20/projects/unifycr2.git/install/bin/unifycrd
# accepting RPCs on address "na+sm://146417/0"
completed sock init server
MLOG CRIT Error opening leveldb database, abc..., path is "/tmp"/unifycr_db-0-0
MLOG CRIT MDHIM Rank: 0 - Error while opening database
MLOG CRIT Rank: 0 - Error opening data store for index: 0
MLOG CRIT Error opening leveldb database, abc..., path is "/tmp"/unifycr_db-1-0
MLOG CRIT MDHIM Rank: 0 - Error while opening database
MLOG CRIT Rank: 0 - Error opening data store for index: 1
```

CamStan: Looking at that link @dsikich posted, I wonder if checkpatch thought the "/" was a mathematical operation and not a path.

client/src/unifycr.c
dsikich: @adammoody do we have to use macros for these paths?

common/src/unifycr_const.h
adammoody: This particular MACRO that defines the DB path is only used in common/src/unifycr_configurator.h.  We could likely avoid this problem by hardcoding the string in the UNIFYCR_CFG statement in that file.  I'll test that.

CamStan: Makes me wonder what Travis CI is doing differently (or wrong) that allowed it to run unifycrd.

PR #232, opened 2018-12-07
Adjust testing framework and add additional tests

CamStan:
### Description

This adjusts the structure of our testing framework to allow for tests to be organized into categories, or test suites, and will also make adding tests for future wrappers easier as well.

The first commit updates this framework structure. Currently if new tests for a different wrapper are wanted, it needs a new .c file, a new .t file to wrap the test (to source the test environment and unifycr settings) and eight new lines added to the Makefile. With this change, those steps only are needed for a new test suite. For each new wrapper being tested now only a .c file is needed and one line added to the Makefile. Then the test suite simply needs to call the new test function. Subsequent commits show examples of this.

New tests are added for creat, creat64, close, open, open64, mkdir, and rmdir.
- Notice our wrapper for creat64 hasn't been implemented yet, so these tests just skip to visually show that.
- Some mkdir tests have identified bugs in our code (at least in regard to how the normal file system behaves). These tests have been marked as XFAIL as they fail when they run but are expected to pass once the bugs are fixed. If it is decided these failures are not bugs, but are how UnifyCR is desired to behave, these tests can then be changed to different tests that show that UnifyCR functions as is desired.
- I'll add additional documentation for how to handle the process of PASS/FAIL/XFAIL/XPASS

A test suite for stdio wrappers has been added, primarily as an example, with some fopen and fclose tests. These may need to be adjusted once we start diving into testing the stdio wrappers.

I've tried to set these commits up as examples of how to add additional tests yourself, but am happy to help anyone learn the process. I'll add more documentation as well, but please point out anything that is unclear.

There's a lot here. Hopefully the multiple commits make it easier to look over.

### Motivation and Context

TESTING!!!

### How Has This Been Tested?

Do the tests count as testing for the tests...?

### Types of changes

### Checklist:

### DIFF
--- a/t/std/stdio_suite.c
+++ b/t/std/stdio_suite.c
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2018, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2018, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyCR.
+ * For details, see https://github.com/LLNL/UnifyCR.
+ * Please read https://github.com/LLNL/UnifyCR/LICENSE for full license text.
+ */
+
+#include <string.h>
+#include <mpi.h>
+#include <unifycr.h>
+#include "t/lib/tap.h"
+#include "t/lib/testutil.h"
+
+#include "stdio_suite.h"
+
+/* The test suite for stdio wrappers found in client/src/unifycr-stdio.c.
+ *
+ *
+ * To add new tests to existing stdio tests:
+ * 1. Simply add the tests (order matters) to the appropriate
+ *    <stdio_function_name>.c file.
+ *
+ *
+ * When a new wrapper in unifycr-stdio.c needs to be tested:
+ * 1. Create a <stdio_function_name>.c file with a function called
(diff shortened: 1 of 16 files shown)


t/std/stdio_suite.c
```diff
@@ -0,0 +1,92 @@
...
+
+    /* 7 tests */
+    fopen_fclose_test(unifycr_root);
+
+    done_testing();
+
+    MPI_Finalize();
+
+    return 0;
+}
```
dsikich: @CamStan This is all really excellent and well documented. Is the plan(8) coming from the 7 tests in fopen_fclost_test + the one test in the main function in stdio_suite.c? Is there a way to have a counter for each test that is run so that we can pass the counter to the plan(<counter_variable>) to it instead of manually updating?

CamStan: @dsikich, yes, that is where the 8 is coming from. 

You make a great point; manually updating them really will be a pain. The tests are not run through any kind of iterative process that would make a counter automatic, but there still might be a better way to do it. Let me think though that. If not, it might be better to revert back to the `plan(NO_PLAN)` option than to have something we need to manually update.

CamStan:
@dsikich, @adammoody, I made the changes to keep from having to manually update `plan(N)`. We can always implement that feature if it turns out it becomes a problem.

This should be ready now.

adammoody:
Thanks, @CamStan

PR #235, opened 2018-12-11
Finalize

adammoody:
Add finalize function to clean up resources allocated during unifycr_init().  Since init is called during mount(), finalize is called during unmount().

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -799,12 +799,11 @@ int unifycr_get_global_file_meta(int fid, int gfid, unifycr_file_attr_t* gfattr)
         return -EINVAL;
     }
 
-    unifycr_file_attr_t* fmeta = NULL;
+    unifycr_file_attr_t fmeta;
     int ret = unifycr_client_metaget_rpc_invoke(
         &unifycr_rpc_context, &fmeta, fid, gfid);
     if (ret == UNIFYCR_SUCCESS) {
-        *gfattr = *fmeta;
-        free(fmeta);
+        *gfattr = fmeta;
     }
 
     return ret;
@@ -2307,9 +2306,6 @@ static int unifycr_finalize(void)
         unifycr_fd_stack = NULL;
     }
 
-    /* free resources allocated in client_rpc_init */
-    unifycr_client_rpc_finalize(&unifycr_rpc_context);
-
     /* clean up configuration */
     int tmp_rc = unifycr_config_fini(&client_cfg);
     if (tmp_rc) {
@@ -2333,6 +2329,9 @@ int unifycr_unmount(void)
     /* invoke unmount rpc */
     printf("calling unmount\n");
     int ret = unifycr_client_unmount_rpc_invoke(&unifycr_rpc_context);
+
+    /* free resources allocated in client_rpc_init */
+    unifycr_client_rpc_finalize(&unifycr_rpc_context);
 
     unifycr_finalize();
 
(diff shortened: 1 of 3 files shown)


adammoody:
This is branched from the style branch, so most of these commits will go away once the style PR is merged.  The last commit is the one of interest here.

adammoody:
@MichaelBrim , running some tests under a memory debugger, I can see some leaks showing up from strings allocated during unifycr_config_set_defaults().  I'm guessing we're overwriting those strings with custom values via command line, config file, or environment variables without free the default value first.  What do you think?

MichaelBrim:
@adammoody I think all the string values used will leak, regardless of default vs user-specified. Looks like we're missing a call to `unifycr_config_fini`

MichaelBrim:
@adammoody Oh nevermind last comment, I see you have it in this PR. Will look deeper

MichaelBrim:
@adammoody yes you're right, anytime we overwrite a non-NULL string setting it will leak. I'll put together a fix PR for that.

adammoody:
Thanks, @MichaelBrim.

@MichaelBrim and @dsikich , would you please review the finalize logic I've added here?  Of course, look for any bugs I've got.  However, also please try to think of other things that we should be cleaning up that I might be missing.

client/src/unifycr.c
```diff
@@ -2266,9 +2307,42 @@ int unifycr_unmount(void)
...
+
+    unifycr_finalize();
+
+    /* free memory tracking our mount prefix string */
+    if (unifycr_mount_prefix != NULL) {
+        free(unifycr_mount_prefix);
+        unifycr_mount_prefix = NULL;
+        unifycr_mount_prefixlen = 0;
+    }
+
```
dsikich: @adammoody I wonder if we should put this in the unmount rpc function?

dsikich:
@adammoody the build is failing because of the macro defining the /tmp path. Are you merging the PR that fixes that one into this?

client/src/unifycr.c
adammoody: I placed things to match the clean up code so that things allocated in the left function were freed in the corresponding right function:

init --> finalize
mount --> unmount
rpc_init --> rpc_finalize

It makes sense to me for us to free this prefix string in unmount.

If we do that, we should probably also move the strdup from init to mount.

adammoody:
Yeah, we should merge the /tmp macro PR before this one.

client/src/unifycr.c
dsikich: @adammoody it might be nice to have all of the corresponding mount functionality in the mount/unmount rpc functions? Will that make it easier to organize?

adammoody:
Thanks @MichaelBrim .  I agree the init/finalize calls are kind of a scramble right now.  @dsikich  also pointed out that we should clean this up.  I'd like to save that for another PR, because we should reorganize and simplify both the mount and unmount calls as part of that cleanup.

For this PR, I just wanted to assemble the finalize logic so that we have it all.  I tried to match free calls to their corresponding allocate as described above.  I also wanted to tear things down in reverse order as they were created, for clean software layering.  The mount call flows like:
```
mount()
    strdup prefix
    init()
    rpc_init()
    mount_rpc_invoke()
```
so here my goal was to have finalize run in reverse:

```
unmount()
    unmount_rpc_invoke()
    rpc_finalize()
    finalize()
    free prefix
```

I'll create a new PR to try to simplify and cleanup the mount and unmount calls.

PR #236, opened 2018-12-11
call unmount for every MPI rank in the job

dsikich:
After adding the unmount RPC function it now
expects unmount to be called from every rank,
not just rank 0. This commit updates the example
programs.

### DIFF
--- a/examples/src/sysio-dir.c
+++ b/examples/src/sysio-dir.c
@@ -338,7 +338,7 @@ int main(int argc, char** argv)
     MPI_Barrier(MPI_COMM_WORLD);
 
 out_unmount:
-    if (!standard && unmount && rank == 0) {
+    if (!standard && unmount) {
         unifycr_unmount();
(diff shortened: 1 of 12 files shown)


examples/src/sysio-dir.c
```diff
@@ -336,7 +336,7 @@ int main(int argc, char **argv)
     MPI_Barrier(MPI_COMM_WORLD);
 
 out_unmount:
-    if (!standard && unmount && rank == 0)
+    if (!standard)
```
adammoody: Did you intend to delete the unmount test here?

dsikich: @adammoody nope :-) Fixed.

adammoody:
Sorry, @dsikich .  The merged style PR trampled on this one a bit.

dsikich:
@adammoody It's ok. I can update.

PR #237, opened 2018-12-11
config: drop UNIFYCR_META_DB_PATH macro

adammoody:
config: drop UNIFYCR_META_DB_PATH macro, define path directly in configurator instead

The checkpatch script does like paths as bare macros.  It requires parens or quotes around them.  However, adding those characters interferes with our configurator macros.  So the solution opted for here is to drop the macro that defines the path and hardcode that path in the configurator file directly.

### Types of changes

### Checklist:


dsikich:
@adammoody doing it this way seems fine to me.

PR #241, opened 2018-12-12
Rework init / finalize flow in client

adammoody:
Simplify init/finalize code.
 - merge unifycrfs_mount into unifycr_mount
 - drop unused functions to expose data region and chunks to external tools (left over from CRUISE)
 - ensure code works when unifycr_use_memfs is disabled
 - detach (unmap and unlink) from shared memory segments
 - close socket to server
 - fix mmap/msync/munmap to call real functions until wrappers fully supported
 - style edits

Depends on Finalize PR being merged first.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -92,7 +92,7 @@ int local_rank_cnt;
 int local_rank_idx;
 
 int local_del_cnt = 1;
-int client_sockfd;
+int client_sockfd = -1;
 struct pollfd cmd_fd;
 
 /* shared memory buffer to transfer read requests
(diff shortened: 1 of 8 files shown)


dsikich:
@adammoody I see a lot of commits but the only differences that are showing up is for the wrappers you added? Looks like it is out of sync with m2?

client/src/unifycr.c
```diff
@@ -2914,29 +2732,151 @@ int unifycrfs_mount(const char prefix[], size_t size, int rank)
...
+            LOGERR("Failed to close socket to server rc=%d (%s)",
+                errno, strerror(errno));
+        }
+    }
+
+    /* invoke unmount rpc to tell server we're disconnecting */
+    printf("calling unmount\n");
+    int ret = unifycr_client_unmount_rpc_invoke(&unifycr_rpc_context);
+
+    /* free resources allocated in client_rpc_init */
```
dsikich: @adammoody so why is there a rpc finalize and a unifycr_finalize that is not an rpc? I see that is true for init too. Is there a reason they shouldn't be one function?

adammoody: We can look at merging the rpc_init into the unifycr_init too, but I'll do that in a new PR since this is ready to merge.

PR #246, opened 2018-12-14
server: fix read bug introduced with errcode field

adammoody:
This fixes a bug introduced with the errcode field in the read reply headers.  This particular code was using an int, along side the other long types in the struct.  The problem arose because the service manager was packing each field individually, while the request manager was reading them using the struct.  Due to the field-size round up used by C, this led to an offset problem, where the header was packed into fewer bytes than the request manager was expecting.

Like one of @MichaelBrim 's recent changes to dev, this patch uses the struct to also pack data on the service manager, so the two are consistent.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifycr_service_manager.c
+++ b/server/src/unifycr_service_manager.c
@@ -87,11 +87,8 @@
  * requesting delegator, these are appened in an ack list
  * which records a set of read replies before being sent */
 typedef struct {
-    long src_fid;    /* global file id of file */
-    long src_offset; /* offset in file */
-    long length;     /* length of data */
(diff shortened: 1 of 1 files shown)


MichaelBrim:
@adammoody @CamStan seems this one is failing some of the new tests. may be worth a look.

CamStan:
@MichaelBrim, yeah, I've been debugging that. I know what it's doing that's causing it to fail, but why it's doing it suddenly on the m2 branch but not on the PR that introduced it or on my own machine weird. I'm meeting with @adammoody in a few minutes to try to work through it.

CamStan:
This is passing now.

PR #253, opened 2018-12-19
Use margo API calls to get input struct for rpcs

dsikich:
Mercury and Margo both have their own respective API calls
to retrieve the rpc input struct. We want to use Margo API
calls when possible as that is the intended interface for
Mercury.

### DIFF
--- a/server/src/unifycr_cmd_handler.c
+++ b/server/src/unifycr_cmd_handler.c
@@ -318,7 +318,7 @@ static void unifycr_mount_rpc(hg_handle_t handle)
 
     /* get input params */
     unifycr_mount_in_t in;
-    int ret = HG_Get_input(handle, &in);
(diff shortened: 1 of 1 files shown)


adammoody:
Does margo also have its own return codes that we can use, or does it just reuse the mercury return codes?

adammoody:
Can we also replace the mercury references in unifycr_clientcalls_rpc.h?

dsikich:
@adammoody which mercury references do you mean? The header files or the types? I don't think the hg_size_t and hg_bulk_t have equivalent margo calls based on their current example code.

adammoody:
I was just wondering if we could fully drop all references to mercury, which would let us then drop it from spack.  Sounds like we still need it for a few things.

dsikich:
@adammoody Yeah, I believe we still need it for a few things.

PR #261, opened 2019-02-05
add rpc handler to compute file size

adammoody:
This adds an rpc handler in the server to compute file size.  It accomplishes this by querying for all key/value pairs for a file using a starting offset of 0 and ending offset of "max file size" which is the max value of a signed long.  It then iterates over each returned key/value pair to find the maximum offset.  This is clearly inefficient, especially for files that have lots of key/value pairs.  However, it should return a correct answer until we get something better.

I don't fully understand the limits of bulk get in MDHIM, so this could be problematic, for example, if there a file has more key/value pairs than can be returned in a single bulk get call. 

Second, it provides a unifycr_gfid_stat() call that returns a stat structure given a global file id.  Up until now, we only had a unifycr_fid_stat() that requires a local file id, but that means a process that does not have a reference to a particular file in its local data structure could not query for the file's stat info.  This also rewrites the stat wrappers to call this new function that uses just the global file id.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifycr-internal.h
+++ b/client/src/unifycr-internal.h
@@ -483,6 +483,9 @@ int unifycr_fid_is_dir_empty(const char* path);
 /* return current size of given file id */
 off_t unifycr_fid_size(int fid);
 
+/* fill in limited amount of stat information for global file id */
+int unifycr_gfid_stat(int gfid, struct stat* buf);
+
 /* fill in limited amount of stat information */
 int unifycr_fid_stat(int fid, struct stat* buf);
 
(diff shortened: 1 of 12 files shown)


dsikich:
@adammoody you've got a check patch warning in this build

CamStan:
Hmm. It's asking for braces on the preprocessor if statement: `#if 0`. Didn't think to test that case when we changed checkpatch. I'll play with that trying to ignore that case.

dsikich:
@CamStan yeah that's weird. @adammoody did you test the file size function at all? I think it would be good to add a test verifying we can get the correct file size now.

adammoody:
@dsikich , I did test the file size manually.  But you bring up a good point.  This is a nice contained case where I could add a test case for it.  I'll work with @CamStan to do that.

adammoody:
Nice catch @MichaelBrim .  I think I updated it to use size_t / hg_size_t to be consistent with the other functions.  Can you please review it again?

PR #268, opened 2019-02-12
Verify correct file size in sysio-write example

adammoody:
This adds a check to verify that the correct file size was created in the sysio-write example.  It should help catch bugs if the internal file size computation goes bad.

We can add similar tests to our travis suite after some write tests are available.

It also changes test_print() to always print a newline and call fflush, along with changing calls to this function to drop newlines.

### Types of changes

### Checklist:

### DIFF
--- a/examples/src/sysio-write.c
+++ b/examples/src/sysio-write.c
@@ -146,7 +146,7 @@ static void report_result(void)
                     "I/O request size:          %llu B\n"
                     "Aggregate write bandwidth: %lf MB/s\n"
                     "Min. write bandwidth:      %lf MB/s\n"
-                    "Total Write time:          %lf sec. (%lf for fsync)\n\n",
+                    "Total Write time:          %lf sec. (%lf for fsync)\n",
                     total_ranks,
                     1.0 * blocksize * nblocks / (1 << 20),
                     1.0 * total_ranks * blocksize * nblocks / (1 << 20),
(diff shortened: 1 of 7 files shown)


examples/src/sysio-write.c
```diff
@@ -345,16 +345,32 @@ int main(int argc, char** argv)
...
+                test_print(rank, "%s size incorrect got %llu, expected %llu",
+                    targetfile, (unsigned long long)sbuf.st_size,
+                    (unsigned long long)expected_size);
+                ret = 1;
+            }
         } else {
-            test_print(rank, "stat(%s) failed: errno=%d (%s)",
+            /* our call to stat failed */
+            test_print(rank, "stat(%s) failed:",
                 targetfile, errno, strerror(errno));
```
CamStan: @adammoody, looks like you might have dropped some format tags on line 371. There's one tag but three arguments on line 372.

dsikich: yeah the "failed:" portion is missing an input

adammoody: Nice catch!  Forgot to drop those arguments.  Turns out that test_print adds them if errno is non-zero, so dropped the formatting flags but forgot the args.

CamStan: That works too!

CamStan:
It building twice threw me for a sec. Realized you pushed a new branch.

PR #270, opened 2019-02-15
fix bugs associated with attaching to an existing superblock

adammoody:
This addresses two bugs uncovered when debugging @CamStan parallel test failures.
- when attaching to a superblock left over from a previous run, we should not reinitialize its internal state
- it should not be a fatal error if mount point already exists in superblock in unifycr_mount

Without this, a second job would overwrite the state from a previous job, so creating any new files would have silently clobbered files created by the run before.

### Types of changes

### Checklist:


MichaelBrim:
re: Travis failures - who knew checkpatch was a spell checker too? also, can we disable that 'SYMBOLIC_PERMS' check?

PR #271, opened 2019-02-15
Add infrastructure for invoking rpcs on the server

dsikich:
infrastructure now allows adding an rpc from the server, which
will be executed on the client

    - initialize margo client as SERVER to allow rpc's
      to be serviced from the client
    - add function to record client address in file encoded
      with app_id and cliend_id
    - add function so that server can lookup client address
    - add client_addr field to app_config struct so that
      client address is looked up and recorded for each app_id
      and client_id during mount call
    - add unifycr_server.h, unifycr_servercalls_rpc.h for later
      adding functions and storing rpc id's
    - add common/src/unifycr_util.c and common/src/unifycr_util.h
      for publish/lookup functions on client and server

### DIFF
--- a/client/src/unifycr.c
+++ b/client/src/unifycr.c
@@ -63,6 +63,9 @@
 
 #include "unifycr_client.h"
 #include "unifycr_clientcalls_rpc.h"
+#include "../../server/src/unifycr_server.h"
+#include "../../server/src/unifycr_servercalls_rpc.h"
+#include "../../common/src/unifycr_util.h"
 
 #include "unifycr_log.h"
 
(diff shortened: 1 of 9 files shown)


dsikich:
@adammoody guessing you closed this because of the rename? 
@MichaelBrim Thanks for the suggestions. I will make these edits and submit a new request. I also thought about moving all of the rpc infrastructure code to common. I think that is a good idea.

adammoody:
@dsikich , yeah every open PR against m2 or dev got whacked when I deleted those branches on GitHub.  Perhaps GitHub has an option to rename a branch to prevent that, but the directions I was following just said to delete the remote branch and push.

adammoody:
@dsikich , sorry :-)

PR #274, opened 2019-02-22
Add infrastructure for invoking rpcs on the server

dsikich:
infrastructure now allows adding an rpc from the server, which
will be executed on the client

    - initialize margo client as SERVER to allow rpc's
      to be serviced from the client
    - add function to record client address in file encoded
      with app_id and cliend_id
    - add function so that server can lookup client address
    - add client_addr field to app_config struct so that
      client address is looked up and recorded for each app_id
      and client_id during mount call
    - add unifycr_server.h, unifycr_servercalls_rpc.h for later
      adding functions and storing rpc id's
    - add common/src/unifycr_util.c and common/src/unifycr_util.h
      for publish/lookup functions on client and server
    - move client/server headers for rpcs used in both to common library


dsikich:
@MichaelBrim @adammoody this is the same PR #271 

But I also added your suggestions to this one.

dsikich:
@MichaelBrim hmm yeah already actually added a field to the app_config_t on the client which holds the margo address (hg_addr_t* client_addr[MAX_NUM_CLIENTS]). And this gets filled in during the mount_rpc call executed on the server. I ended up adding this since I needed the app_id and client_id encoded in the file name, and the app_id wasn't defined until the filesystem is mounted. 

I suppose we can actually remove/change the publish/lookup calls for the server looking up the client address, and just read from that field in the app_config struct instead, since it is already being passed to the server on mount.

dsikich:
@MichaelBrim I've changed this PR send the client address through the input struct in the mount_rpc instead of creating a file for each client_id. I kept the file functions around in common though. 

@MichaelBrim @adammoody Can you take a fresh look at it when you get a chance?

Thanks.

dsikich:
@MichaelBrim yeah, i removed them from the rpc_util files in common. i also renamed the sir_id file to unifycrd_id

common/src/unifycr_rpc_util.c
```diff
@@ -0,0 +1,67 @@
...
+    char* str = NULL;
+
+    /* TODO: support other lookup methods here like PMIX */
+
+    /* read server address string from well-known file name in ramdisk */
+    FILE* fp = fopen("/dev/shm/unifycrd_id", "r");
+    if (fp != NULL) {
+        /* opened the file, now read the address string */
+        char addr_string[50];
+        int rc = fscanf(fp, "%s", addr_string);
```
adammoody: While we're at it, I'm worried about possible buffer overflow here.

Let's bump the address array from 50 to 256.  This in case Mercury decides to use address strings longer than 50 bytes, which feels a bit on the short side.

Also, let's add a field width value in the format like fscan(fp, "%255s", ...) to prevent a really long string from overrunning our addr_string buffer.  I picked one less than the buffer size to leave room for the trailing NULL.

common/src/unifycr_rpc_util.h
```diff
@@ -0,0 +1,25 @@
...
+ * Please read https://github.com/LLNL/UnifyCR/LICENSE for full license text.
+ */
+
+#ifndef UNIFYCR_UTIL_H
+#define UNIFYCR_UTIL_H
+
+/* publish the address of the server */
+void rpc_publish_server_addr(const char* addr);
+
+/* lookup address of server */
```
adammoody: `/* lookup address of server, returns NULL if server address is not found,
 * otherwise returns server address in newly allocated string that caller must free */`

dsikich: @adammoody thanks for the review. i've updated the code.

dsikich:
@adammoody this build is failing for the same reason as PR #725 https://travis-ci.org/LLNL/UnifyCR/builds/499556940

That one got merged in somehow but the travis build is broken on dev

adammoody:
Yeah, looks like our git clone to grab the margo spack package is failing.

```0.11s$ git clone https://xgitlab.cels.anl.gov/sds/sds-repo.git $HOME/sds-repo.git
Cloning into '/home/travis/sds-repo.git'...
fatal: unable to access 'https://xgitlab.cels.anl.gov/sds/sds-repo.git/': gnutls_handshake() failed: Handshake failed
The command "git clone https://xgitlab.cels.anl.gov/sds/sds-repo.git $HOME/sds-repo.git" failed and exited with 128 during .
Your build has been stopped.
```

Perhaps deleting the travis cache will work, or otherwise it looks like travis changed a setting that is blocking the git clone command from working.

CamStan:
Clearing the cache had no luck. At present I'm leaning towards the idea that travis changed a setting that's blocking that clone. I haven't been able to find much regarding this particular issue. I've tried anything I could find on my forked repo's travis and nothing has worked thus far. Got one or two other things I can try but after that I'm at a loss.

adammoody:
We may just have to check in the spack files from the sds repo into our own repo to avoid the git clone.  Another thing might be to try the SSH git URL instead of the https version.

I love it when we spend as much time debugging travis as we do our own code (sigh)

dsikich:
@CamStan yeah using the ssh version instead of https might fix it.

CamStan:
Our build has three clones but only this one fails. Possibly because it's not on GitHub, which travis seems to favor. 

Here are the results of the attempt to change to the ssh version:
```
$ git config --global http.sslVerify false
$ echo -e "Host xgitlab.cels.anl.gov\n\tStrictHostKeyChecking no\n" >> $HOME/.ssh/config
$ git clone git@xgitlab.cels.anl.gov:sds/sds-repo.git $HOME/sds-repo.git
Cloning into '/home/travis/sds-repo.git'...
Warning: Permanently added 'xgitlab.cels.anl.gov,140.221.6.150' (RSA) to the list of known hosts.
Password:
```
Need to use a RSA key to get around this and apparently you can only use those with travis if you're using the payed-for version. I'm no expert on this process though, so maybe there's still a way to get this work.

The other thing I tried was to change the [virtualization environment](https://docs.travis-ci.com/user/reference/overview/#deprecated-virtualization-environments) that our travis build runs in. It apparently defaults to Ubuntu 14.04 (I've seen others with similar issues related to this version but no idea why ours suddenly started to fail). I changed this to Ubuntu 16.04 and the build **DID** succeed. However, I don't know if this will eventually cause other problems. My hope is that it won't since Ubuntu 16.04 will most likely eventually become travis' default.

However, this wouldn't be a problem if margo was in spack's main repo.

Should I just push the virtualization environment change and call it good for now?

dsikich:
@CamStan Awesome you figured that out! I would push that and call it good for now.

CamStan:
Pushed in #279

PR #275, opened 2019-02-25
fix libmdhim_a_LDFLAGS warning

MichaelBrim:
### Description

Running autogen.sh has been producing a warning along the lines of 
"no matching target for libmdhim_a_LDFLAGS". The message is a bit
misleading, but what it really means is that "target_LDFLAGS" is not a
valid automake construct for library archive targets, since the linker
is not used to make them. If we ever need to pass flags, we should 
use target_ARFLAGS.

### Types of changes

### DIFF
--- a/meta/src/Makefile.am
+++ b/meta/src/Makefile.am
@@ -27,7 +27,8 @@ libmdhim_a_SOURCES = Mlog2/mlog2.c \
                      mdhim.h \
                      uthash/uthash.h
 
-libmdhim_a_LDFLAGS = $(AM_LDFLAGS) $(LEVELDB_LDFLAGS) $(LEVELDB_LIBS) $(MPI_CLDFLAGS)
+# target_LDFLAGS is not valid for library archive targets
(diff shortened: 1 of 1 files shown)


MichaelBrim:
unifycr_unmount() test is failing in Travis. I'm actually seeing unmount failures on Summitdev too for other branches based on dev. I'm pretty sure commenting out libmdhim_a_LDFLAGS is not the cause :) 

I'll continue investigating unmount to see if I can figure out what's going wrong.

CamStan:
We noticed something was happening right before going home last Friday but didn't have time to look into. I did rebuild the dev branch from scratch on Travis (which I believe is the first time it's been done since the branches were renamed) and the unifycr_unmount() test is failing there too. Potentially a bug snuck in that a Travis was missing because of caching things but not sure. Unfortunately that test isn't the most informative at the moment; it's only telling us unmount is returning a very large number instead of zero.

MichaelBrim:
Looks like the server is using the wrong out struct in the unmount rpc handler. Will submit a separate PR to fix.

adammoody:
Thanks @MichaelBrim and @CamStan !

Issue #300, opened 2019-04-12
Intel Fortran compiler errors out on unifycr file descriptors


adammoody:
With the Intel Fortran compiler, I get the following:

```
+ test_write
forrtl: Bad file descriptor
forrtl: severe (108): cannot stat file, unit 0, file /unifycr/file.ckpt
Image              PC                Routine            Line        Source             
test_write         000000000040B0CE  for__io_return        Unknown  Unknown
test_write         000000000041A0C5  for_open              Unknown  Unknown
test_write         0000000000408A51  Unknown               Unknown  Unknown
test_write         000000000040866E  Unknown               Unknown  Unknown
libc-2.17.so       00002AAAAC09D3D5  __libc_start_main     Unknown  Unknown
test_write         0000000000408569  Unknown               Unknown  Unknown
```

I was able to see that this was in fact running through the UnifyCR I/O wrappers. The for_open was eventually calling into the open64 wrapper. It's possible that the Intel Fortran I/O library may not like the out-of-range file descriptor the wrapper is returning.

### System information

Type                   | Version/Name
---                    |     ---
Operating System       |
OS Version             |
Architecture           |
UnifyCR Version        |

PR #288, opened 2019-03-09
add fortran interface for mount and unmount

adammoody:
New files unifycrf.{c,h} define Fortran wrappers to call unifycr_mount/unmount calls.

### Description

I have some simple test cases not included in this current PR.  I can see that UNIFYCR_MOUNT from Fortran calls unifycr_mount just fine.  However, my Fortran call to open() is not intercepted by our open wrapper as I was hoping for.  Will need to figure out what's going on there.

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Performance enhancement (non-breaking change which improves efficiency)
- [ ] Code cleanup (non-breaking change which makes code smaller or more readable)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (a change to man pages or other documentation)

### Checklist:

- [ ] My code follows the UnifyCR code style requirements.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] All commit messages are properly formatted.

PR #302, opened 2019-04-16
Add note about GFortran usability to docs

CamStan:
This adds a note about compatibility with GFortran to the docs and lets users know about issue #300.

This was accomplished with #288  

### Types of changes

### DIFF
--- a/docs/api-mount.rst
+++ b/docs/api-mount.rst
@@ -4,9 +4,31 @@ Mounting UnifyCR
 
 In this section, we describe how to use the UnifyCR API in an application.
 
+.. Attention:: **Fortran Compatibility**
+
+   ``unifycr_mount`` and ``unifycr_unmount`` are now usable  with GFortran.
+   There is a known issue_ with the Intel Fortran compiler. Other Fortran
+   compilers are currently unknown.
+
 ---------------------------
 Mounting 
 ---------------------------
+
+In ``C`` applications, include *unifycr.h*. See writeread.c_ for a full
+example.
+
+.. code-block:: C
+    :caption: C
+
+        #include <unifycr.h>
+
+In ``Fortran`` applications, include *unifycrf.h*. See writeread.f90_ for a
+full example.
+
+.. code-block:: Fortran
+    :caption: Fortran
+
+        include 'unifycrf.h'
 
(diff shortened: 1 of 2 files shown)


docs/api-mount.rst
```diff
@@ -4,6 +4,12 @@ Mounting UnifyCR
 
 In this section, we describe how to use the UnifyCR API in an application.
 
+.. Note:: **Fortran Compatibility**
+
+   ``unifycr_mount`` and ``unifycr_unmount`` are now usable  with GFortran.
```
adammoody: Thanks @CamStan .  Just because this looks strange to me, let's use the name in all lower case as "gfortran".  That's how I see it most of the time.

adammoody: Well, seems to use lots of variants out on the web:  "GFortran", "Gfortran", "gfortran", so I take that back.  Good how you have it.

adammoody: Oh, we should mention that a Fortran application needs to link with:
```
-lunifycrf -lunifycr_gotcha
```

adammoody:
And let's mention that the Fortran source needs to include unifycrf.h:
```
include 'unifycrf.h'
```

As an example for how to integrate the UNIFYCR_MOUNT/UNMOUNT calls, we could point users to: https://github.com/LLNL/UnifyCR/blob/dev/examples/src/writeread.f90

CamStan:
Updated.

docs/build-intercept.rst
```diff
@@ -208,8 +208,19 @@ Steps for dynamic linking using gotcha:
...
         -I<unifycr>/include -L<unifycy>/lib -lunifycr_gotcha \
         -L<gotcha>/lib64 -lgotcha
+
+Fortran
+^^^^^^^^^^^^^^
+
+.. code-block:: Bash
+
+    $ mpif90 -o test_write test_write.F \
+        -I<unifycr>/include -L<unifycy>/lib -lunifycrf -lunifycr_gotcha \
```
CamStan: @adammoody, I want to double check that this is correct.

CamStan: Aside from the extra `\` on the end...

adammoody:
Nice!  Thanks, @CamStan

PR #307, opened 2019-04-30
code cleanup before server-server margo

MichaelBrim:
- Fixes build warnings for gcc and XL compilers
- Moves files into/out of common where appropriate
- Cleans up #includes for client and server
- Remove --enable-debug configure option

Note on '--enable-debug': other than modifying user CFLAGS, which
is considered bad practice, we were not doing anything useful when
this option was given. The recommended method to support debug
builds is to have the user pass the appropriate CFLAGS on the
configure command.

### Types of changes

### Checklist:


CamStan:
No idea if this is the case, but if since Travis is building off of the cache, any changes to the way the build goes can make things go weird. This environment might now have mercury, mercury@develop, margo, and margo@develop. Might need to source the one you want since you changed it in the travis.yml, and maybe clearing the cache will help.

MichaelBrim:
I should note that on Summitdev, with the most recent mercury spackage, I had to install:
```
mercury@develop+sm
margo@develop ^mercury@develop+sm
```

CamStan:
> I should note that on Summitdev, with the most recent mercury spackage, I had to install:
> 
> ```
> mercury@develop+sm
> margo@develop ^mercury@develop+sm
> ```
> 
> This is because the new spackage turns off the shared-memory plugin by default, but we use it in Unify.

I imagine we'll need to adjust our Spack package for this then?

MichaelBrim:
@CamStan yes, since the updated sds-repo spackages got upstreamed 12 days ago.

MichaelBrim:
@CamStan we may also want to eliminate using the sds-repo in the Travis build since the packages are all in upstream

CamStan:
Oh, nice! I'll get on it.

configure.ac
```diff
@@ -87,20 +87,6 @@ else
 AM_CONDITIONAL([USE_PMIX], [false])
 fi
 
-# Debug build
-AC_ARG_ENABLE([debug],
-              AC_HELP_STRING([--enable-debug],
-                             [Enable debug build options.]))
```
CamStan: I'm not clear on why the removal of this option. Currently updating our spackage which has a variant for this.

MichaelBrim: It wasn't doing anything useful - see note in the issue description. If someone wants a debug build, they should use `configure <config-opts> "CFLAGS=-g -O0"` or similar

MichaelBrim: The actual CFLAGS to pass are compiler dependent.

PR #308, opened 2019-05-03
Add lua plugin for SLURM integration

dsikich:
A SLURM SPANK lua plugin that adds an option to
SLURM called unifycr. Using this option starts
and stops the unifycr daemon as the user at job
startup and termination.

This also makes the assumption currently that 'unifycrd' is installed in a default system path. The SPANK functions such as task_init and task_exit only run as the user running the job in a remote context.

Plugin still needs to be tested (in progress).

### DIFF
--- a/util/plugins/slurm/README.md
+++ b/util/plugins/slurm/README.md
@@ -1,0 +1,26 @@
+### Integration with SLURM Resource Manager
+
+To enable job-level integration of UnifyCR, where the UnifyCR daemon
+is automatically launched on each node during job startup and
+terminate once the job completes, a SPANK lua plugin must be added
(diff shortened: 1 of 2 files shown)


MichaelBrim:
@dsikich I noticed `-alloc_flags` was still used in the README. Does srun/salloc really have that flag too? I thought it was specific to LSF's bsub.

dsikich:
@MichaelBrim I don't think so, I started with your README for lsf, and missed deleting that.

dsikich:
@MichaelBrim Thanks for catching the `alloc_flags` in the README.md.. it's been updated/removed.

CamStan:
Creating an issue to pick this up in the future.

PR #309, opened 2019-05-03
Add RPC documentation to readthedocs

dsikich:
This adds a page that describes how to
add an RPC function with the Margo API.


dsikich:
@CamStan Thanks, added it to index.rst.

Issue #299, opened 2019-04-01
Data transfer API (between unifycr and other file system)


sandrain:
I am opening a discussion about how unifycr will support the data transfer from/to other file systems.

The initial interface to support data transfers is (based on our milestone document M3Y1):

```
unifycr_transfer(IN source_fname, IN dest_fname, IN flags, OUT status_handle);

unifycr_transfer_async(IN source_fname, IN dest_fname, IN flags, OUT req_handle);
```

These functions will perform data transfers in both directions, i.e., from unifycr to an external file system and from an external file system to unifycr. The direction is supposed to be specified using the argument `flag`, possibly with other available control options. The API is quite intuitive, so I think it can be directly adopted. In C, the above interface would look:

```

enum {
    UNIFYCR_TRANSFER_TO_UNIFYCR = 0,      /* from external to unifycr */
    UNIFYCR_TRANSFER_FROM_UNIFYCR = 1,    /* from unifycr to external */
    /* more control options ... */
    UNIFYCR_TRANSFER_DIR = 2,    /* essentially, cp -r (useful for N-N)*/
};

/* the asynchronous transfer function */
int unifycr_transfer_async(const char *src, const char *dst, int flags, int handle);

enum {
    UNIFYCR_TRANSFER_ERROR = -1,
    UNIFYCR_TRANSFER_IN_PROGRESS = 0,
    UNIFYCR_TRANSFER_COMPLETE = 1,
};

/* check the current status */
int unifycr_transfer_status(int handle);

/* cancel the on-going transfer */
int unifycr_transfer_cancel(int handle);

/* wait until the transfer is complete */
int unifycr_transfer_wait(int handle);

/* synchronous wrappers */
int unifycr_transfer(const char *src, const char *dst, int flags);

int unifycr_transfer_in(const char *src, const char *dst);

int unifycr_transfer_out(const char *src, const  char *dst);
```

For the transfer direction, we can also infer from the source and destination pathnames (by observing which pathname has `"/unifycr"`). In such a case, we can eliminate some constants and functions.

Internally, the transfer functions should be able to exploit any possible parallelism. For instance, when a file in unifycr has to be copied to a parallel file system, it is not ideal for a single client is collecting all data and then writing in a single stream. Instead, it would be better if all participant compute nodes transfer the target file in parallel by each node handling locally stored portion of the file. To do so, it would require:

[...]

PR #310, opened 2019-05-08
unifycr_transfer_file, a simple version of data stage operation

sandrain:
- client/src/unifycr.h: added the function prototype
- client/src/unifycr.c: the implementation of the client-side transfer
- client/src/unifycr-sysio.c: change the read(2) to return success on EOF
- examples/src/transfer.c: testing unifycr_transfer_file api

addressing #299 .

### Types of changes

### DIFF
--- a/client/src/unifycr-sysio.c
+++ b/client/src/unifycr-sysio.c
@@ -1023,14 +1023,18 @@ ssize_t UNIFYCR_WRAP(read)(int fd, void* buf, size_t count)
          */
         int ret = unifycr_fd_logreadlist(&tmp_req, 1);
 
+        /*
+         * FIXME: when we can get the global file size correctly, the following
+         * should be rewritten. currently, we cannot detect EOF reliably.
+         */
         if (ret != UNIFYCR_SUCCESS) {
-            /* error reading data */
-            errno = EIO;
-            retcount = -1;
-        } else if (tmp_req.errcode != UNIFYCR_SUCCESS) {
-            /* error reading data */
-            errno = EIO;
-            retcount = -1;
+            if (tmp_req.errcode != UNIFYCR_SUCCESS) {
+                /* error reading data */
+                errno = EIO;
+                retcount = -1;
+            } else {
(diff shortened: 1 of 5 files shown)


adammoody:
Thanks @sandrain .  As you discovered, we currently don't handle reporting EOF when reading to the end of a file.  However, I think calling stat should correctly return the file size.  For now, I think you can get something function if you call stat to get the file size, and then manually track reading up to that file size but not beyond.

sandrain:
Hold the merging until I complete the test of creating a file with multiple clients (N-1).

sandrain:
I've tested with four nodes, and test test was successful. I've used the following script:


```
#!/bin/bash

mpirun="mpirun -f nodelist.txt -np 4"

rm -f /tmp/tt*

# stage out
for i in `seq 1 4`; do
    $mpirun `pwd`/prefix/libexec/sysio-write-static -n8 -f testfile$i;
    $mpirun `pwd`/prefix/libexec/sysio-read-static -n8 -f testfile$i;
    $mpirun `pwd`/prefix/libexec/transfer-static /unifycr/testfile$i /tmp/tt$i;
done

# stage in
for i in `seq 1 4`; do
    $mpirun `pwd`/prefix/libexec/transfer-static /tmp/tt$i /unifycr/tt$i;
    $mpirun `pwd`/prefix/libexec/sysio-read-static -n8 -f tt$i
done

```

For stage in, because we only use a single process for transferring, the file data is all written in a single shmem space of the process.

adammoody:
Thanks @sandrain !  This PR is good, although we'll likely need to come back and update the error handling in unifycr_fd_logreadlist().

Your work here led me to uncover some inconsistencies in how we're reading data that I need to look into.  Running this under totalview, I can see the server returns data for a regional of the file that is just before the requested region which is for a block of data beyond the end of the file.  The implementation of unifycr_fd_logreadlist() then detects the mismatch and returns an error, which is what it should do.  I need to debug things further.

PR #316, opened 2019-05-14
initialize and publish margo remote RPC address

MichaelBrim:
### Description

This commit switches to using "bmi+tcp" Mercury provider
for remote (i.e., server-server) RPCs since the ofi provider
interferes with MPI on test systems that have Infiniband.

The local and remote server addresses are published using PMIx.

The 'Adding RPCs' documentation has been updated to
reflect the recent code restructuring.

This also fixes a case where using the LOG() macro
before unifycr_log_open() was called would lead to a
segfault, by making sure the log stream is never NULL.

### Types of changes

### Checklist:


.travis.yml
```diff
@@ -54,7 +54,7 @@ install:
   - spack install leveldb
   - spack install gotcha@0.0.2
   - spack install flatcc
-  - spack install margo^mercury~boostsys
+  - spack install margo^mercury+bmi~boostsys
```
CamStan: Is this just an option we need to document for users to do when building on a system with Infiniband? Or do we need to always require this? 
We could:
1. Just document it and let the user designate when to use it or not: `spack install margo^mercury+bmi`
2. Make it a variant in our spackage: 
    ```
    variant('infiniband', default='False', description="Enable InfiniBand support. Uses Mercury BMI plugin."
    ⋮
    depends_on('mercury+bmi', when='+infiniband')
    ```
3. Always require it in our spackage: `depends_on('mercury+bmi')`, which means we can't build without it when using spack

If the second option, do we use both libfabric and BMI at the same time, or is it a one-or-the-other option? If the latter, we could omit `ofi` from Mercury as well: `depends_on('mercury+bmi~ofi', when='+infiniband')`.

MichaelBrim: Right now, the '+bmi' is required for server-server margo, regardless of the physical network type. I'm using 'bmi+tcp' as the mercury transport, which just uses TCP sockets for interprocess communication. So, if we try running with a mercury that doesn't support the bmi variant, server initialization will fail.

I didn't turn off the ofi variant since it only seems problematic on IBM SpectrumMPI systems like Summit/Sierra. We should document the problem for those systems and tell users to disable ofi. We need to test on other Infiniband clusters to see if it interferes with their MPI too.

CamStan: So it sounds like we'll want `depends_on('mercury+bmi')` in our spackage so users get a heads up before server initialization fails.

If that's the case, we could also put 
```
conflicts('^spectrum-mpi, when='^mercury+ofi' msg='Install with ^mercury~ofi or use a different mpi library')
```
in our spackage.

MichaelBrim: Yes, I think we should depend on mercury+bmi in spack for now. I'm not sure if mercury+ofi conflicts with all SpectrumMPI systems, or just the infiniband ones like Summit/Sierra.

CamStan: Ah, okay. Thank you. I'll make an note about that but won't include a conflicts.

Lastly, we also conflict if the shared-memory plugin for Mercury is disabled (`mercury~sm`), correct?

MichaelBrim: Yes, we always need +sm.

MichaelBrim:
@CamStan @adammoody I rebased this on current dev - ready for (re)review

CamStan:
@MichaelBrim, we've run into some weird issue with spack on our systems using LSF here when trying to install bmi. I'm assuming you didn't have any issues when doing this.

Spack seems to do some magic with autotools when it comes to their `config.guess` file, but it's not liking something with our configuration here:
```
==> Executing phase: 'configure'
==> Error: ProcessError: Command exited with status 1:
    'prefix/spack/var/spack/stage/bmi-develop-grn74jiucrdr3w73anyxbs7uujs73746/bmi/configure' '--prefix=prefix/spack/opt/spack/linux-rhel7-ppc64le/gcc-4.9.3/bmi-develop-grn74jiucrdr3w73anyxbs7uujs73746' '--enable-shared' '--enable-bmi-only'

1 error found in build log:
     130    /usr/convex/getsysinfo =
      ⋮
  >> 136    configure: error: cannot guess build type; you must specify one
```
@becker33, correct me if I'm wrong.

Don't know that this should prevent us from merging though.

MichaelBrim:
@CamStan I had already forgotten about that. Yes, I couldn't build bmi using spack on the IBM machines, since its old config.guess doesn't know about ppc64le systems. I was going to ask Phil Carns to update the bmi spackage and source tarball, but never did so. For my testing, I ended up building bmi, mercury, and margo from source. Once bmi is updated, we should be able to get around this problem.

CamStan:
@MichaelBrim I do recall you saying something about that now, but didn't connect the dots. I can send Phil a note, unless you already did. Or we could just tag him in on this conversation I suppose.

MichaelBrim:
@CamStan I'll contact him.

MichaelBrim:
@CamStan Phil updated the bmi git repo to use the updated config.* files. I did a spack install on Summit TDS and it worked. I've cleared the Travis cache and restarted the build to make sure nothing else goes wrong.

adammoody:
Whoops, I see you guys are discussing mercury items.  Sorry if I jumped the gun on that merge.

MichaelBrim:
@adammoody I think this was good to go now that Phil Carns updated bmi to work on ppcle64

PR #320, opened 2019-05-21
fix: skip data lookup when no keys returned

adammoody:
We have a bug in the case that no keys are returned in which we still send a request to lookup data to a request delegator.  It tends to work if a previous valid request has been issued, since it reuses the info for the previous valid read, which returns data out of range for the current request(s), which is then caught as an error on the client -- all by luck.

This partially fixes that to skip the lookup.

We still need logic to handle the case of holes.

### Types of changes

### Checklist:


adammoody:
Addresses https://github.com/LLNL/UnifyCR/issues/319

adammoody:
I'll leave fixing the holes to a separate PR, since it may be non trivial.  This at least fixes one bug.

PR #322, opened 2019-05-24
move from pmix to generic keyval interface

MichaelBrim:
### Description

This transitions from unifycr_pmix.[ch] to unifycr_keyval.[ch]
and implements key-value support for PMI2 and shared file
systems. The latter is the default behavior when
--enable-pmix or --enable-pmi is not selected at configure
time. The shared file system kvstore support depends on
the user setting a new configuration setting cfg.sharedfs_dir
(with corresponding env var UNIFYCR_SHAREDFS_DIR) to a path
on a file system visible to all servers/clients.

### How Has This Been Tested?

I tested on Summit TDS to make sure PMIx and file system kvstore functionality
was working. I still need to test on a PMI2 system (e.g., SLURM) to make sure 
the new support is working. 

### Types of changes

### Checklist:


common/src/unifycr_keyval.c
```diff
@@ -0,0 +1,673 @@
...
+static char pmi_errstr[MPI_MAX_ERROR_STRING];
+
+// initialize PMI2
+static int unifycr_pmi2_init(void)
+{
+    int rc, version, nprocs, rank;
+
+    rc = PMI2_Init(NULL, &nprocs, &rank, NULL, &version, NULL);
+    if (rc != MPI_SUCCESS) {
+        MPI_Error_string(rc, pmi_errstr, &pmi_errstr_len);
```
adammoody: Do you mean to use MPI to report a PMI error string?  I don't see an PMI2_Error_string in the pmi2.h, but I'm guessing MPI may not do this portably across different MPI libs.

adammoody:
Thanks, @MichaelBrim .  I like the abstraction here.  This will give us a lot of flexibility.  If PMI is broken on a system, then the file system will be a good fall back.

For PMI2, I seem to remember that the KVS_Put/Get are only meant for a group of procs that start up together, like the unify servers.  Those procs share a common key-value space.  Then one is supposed to use the Publish/Lookup methods for exposing info across process groups, like between the unify servers and clients.  However, I also seem to remember that the publish/lookup implementation in SLURM PMI2 is not complete?

common/src/unifycr_keyval.c
MichaelBrim: According to PMI2 docs I found, you are supposed to use MPI_Error_string() to get the string version of the PMI error.

MichaelBrim:
@adammoody I still need to test the PMI2 support, so we'll see if I run into the process group restrictions you mentioned. I had a discussion with the SchedMD lead developer back in April and he suggested their PMI2 support was complete.

common/src/unifycr_keyval.c
MichaelBrim: I ended up removing the use of MPI_Error_string(), and coding up my own function that uses the PMI2 error definitions.

MichaelBrim:
@adammoody I have this building ok on LC Quartz with Slurm's pmi2.h and libpmi2, but PMI2_Init() is failing. I'm not sure what's going wrong. Do you know of a particular LC system where PMI2 is known to be functioning with the default MPI?

adammoody:
I believe we are using PMI2 to start up MVAPICH2, but I would have to double check.  I think the SLURM PMI2 plugin also supports PMI apps, so MVAPICH2 is really using PMI.

Do you get any clues from the return code / error message?

MichaelBrim:
@adammoody This is ready now. It's working with SLURM PMI2 on LC Quartz. I also took your suggestion and moved to a rank-based remote lookup, rather than hostname-based.

adammoody:
Thanks, @MichaelBrim

Issue #324, opened 2019-06-15
Multi-rank local file size bug


CamStan:
### Describe the problem you're observing
When running the [write](https://github.com/LLNL/UnifyCR/blob/dev/examples/src/write.c) and [writeread](https://github.com/LLNL/UnifyCR/blob/dev/examples/src/writeread.c) examples (mostly tested with gotcha run mode, but static does this as well) with more than one process, I get an error that looks like I simply ran out of spillover space.

```
unifycr_chunk_alloc: spill-over device out of space (15)
[1] write() failed (errno=28, No space left on device)
[1] write req 0 failed (errno=28, No space left on device)
[1] write() failed (errno=28, No space left on device)
[1] write req 1 failed (errno=28, No space left on device)
⋮
```
However, not always on the first attempt (but mostly) and always every attempt after, this occurs when writing a file that would only be 64MB in size (each process writing two chunks of 16MB a piece) and when there is multiple GB of spillover space available.

Notice this error was only thrown by `rank 1` and never by `rank 0`.

### Describe how to reproduce the problem
Set the following environment variables before running the server:
```
export UNIFYCR_DAEMONIZE=off
export UNIFYCR_SPILLOVER_ENABLED=yes
export UNIFYCR_DEBUG=on
export UNIFYCR_LOG_VERBOSITY=5
export UNIFYCR_SPILLOVER_SIZE=1073741824
export UNIFYCR_SPILLOVER_META_DIR=/path/to/ssd
export UNIFYCR_SPILLOVER_DATA_DIR=/path/to/ssd

# start the server
/path/to/bin/unifycr start -m /unifycr -e /path/to/bin/unifycrd &
```
While debugging, I used the `write-gotcha` example with either 1 node and 2 processes or 2 nodes and 1 process on each:
```
jsrun -n2 -a1 -c18 -r1 /path/to/libexec/write-gotcha -p n1 -n 2 -c 4096 -b 16384 -v -f testfile
```

### Bug 1
I'm not familiar with how all this should work as of yet, so here is what I observed. After stepping through the code multiple times, I believe I've narrowed down the cause and potentially another related bug. 

[...]

PR #325, opened 2019-06-15
Add missing fattrs to input struct in client metaset rpc

CamStan:
### Description
This adds fattrs to the input struct in the client metaset rpc so that when other ranks aside from rank 0 get the global file size, it is accurate.

See #324 for full description.

### How Has This Been Tested?
By running the examples in multiple ways as described in issue #324.

### Types of changes

### DIFF
--- a/client/src/margo_client.c
+++ b/client/src/margo_client.c
@@ -260,8 +260,13 @@ int invoke_client_metaset_rpc(unifycr_file_attr_t* f_meta)
     in.fid      = f_meta->fid;
     in.gfid     = f_meta->gfid;
     in.filename = f_meta->filename;
-    /* TODO: unifycr_metaset_in_t is missing struct stat info
-     * in->file_attr = f_meta->file_attr; */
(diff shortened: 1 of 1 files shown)


CamStan:
As mentioned in #324, I dove into some deep holes on this one and although feels I have a better understanding, I'm not positive on how the bigger picture of all these parts are suppose to behave (plus I've been staring at it way too long). However, I'm not sure where this assignment of the global file's size to the non-zero rank's local file size comes into play, or if anything should be addressed with it in this PR as well:
https://github.com/LLNL/UnifyCR/blob/52c2d7d9f36df1d3e8247d6ba0ef77802bb9313d/client/src/unifycr.c#L1319

@adammoody, I believe you also mentioned we might have some bugs in relation to global file size vs local file sizes.

MichaelBrim:
@CamStan The initial size in the metadata set during file creation will be 0, and that's what other clients who open the newly created file will see until somebody fsync()s their writes.

PR #336, opened 2019-06-28
Fix testutil create file bug

CamStan:
### Description

When running examples with more than one process and either without UnifyCR or when using file-per-process, this makes sure each process creates the file so that it can be found when opening.

### Motivation and Context

When running the `writeread` example with more than one process and doing file-per-process, only process 0 creates the file and the other processes then cannot find the file when going to open it:
```
jsrun -n2 -a1 -c18 -r1 /prefix/UnifyCR/install/libexec/writeread-static -p nn -n2 -c 4096 -b 16384 -v -f testfile
```
results in:
```
DEBUG: generating write requests
DEBUG: creating target file /unifycr/testfile-0
DEBUG: starting write requests
../../../client/src/unifycr.c:1354 in unifycr_fid_open: /unifycr/-1 does not exist (O_CREAT not given)
```
Not using UnifyCR:

**Update:** this one is fine. See comment [below](https://github.com/LLNL/UnifyCR/pull/336#issuecomment-506843838).
```
jsrun -n2 -a1 -c18 -r1 /prefix/UnifyCR/install/libexec/writeread-posix -p n1 -n2 -c 4096 -b 16384 -v -U -m /dev/shm -f testfile2
```
results in:
```
DEBUG: creating target file /dev/shm/testfile2
DEBUG: generating write requests
DEBUG: starting write requests
[1] ERROR: open(/dev/shm/testfile2) failed (errno=2, No such file or directory)
```
and not using UnifyCR while using file-per-process:
```
jsrun -n2 -a1 -c18 -r1 /prefix/UnifyCR/install/libexec/writeread-posix -p nn -n2 -c 4096 -b 16384 -v -U -m /dev/shm -f testfile3
```
results in:
```
DEBUG: creating target file /dev/shm/testfile3-0
DEBUG: generating write requests
DEBUG: starting write requests
[1] ERROR: open(/dev/shm/testfile3-1) failed (errno=2, No such file or directory)
```

### Types of changes

### DIFF
--- a/examples/src/testutil.h
+++ b/examples/src/testutil.h
@@ -872,8 +872,8 @@ int test_create_file(test_cfg* cfg, const char* filepath, int access)
         fmode = test_access_to_stdio_mode(access);
     }
 
-    // rank 0 creates
-    if (cfg->rank == 0) {
+    // rank 0 creates or all ranks create if using file-per-process
+    if (cfg->rank == 0 || cfg->io_pattern == IO_PATTERN_NN) {
         if (cfg->use_stdio) {
             fp = fopen(filepath, fmode);
             if (NULL == fp) {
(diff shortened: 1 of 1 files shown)


CamStan:
@MichaelBrim, ended up duplicating some code for this since using `-p n1` while using UnifyCR still requires the `test_barrier(cfg)` call. Possibly a better way to fix this.

MichaelBrim:
@CamStan For writeread-posix in single-shared file (-p n1) mode, the mountpoint directory *must* be on a shared file system (/dev/shm is node-local). It should work that way with no changes.

CamStan:
@MichaelBrim, nice catch, thank you. Adjusted this so it only creates for all ranks if file-per-process and the barrier only used if single-shared file.

adammoody:
Thanks @CamStan and @MichaelBrim

PR #341, opened 2019-07-12
Add bootstrap.sh, fix build errors on GCC 9

tonyhutter:
### Description
- Add bootstrap.sh to automatically download and build depedencies
- Fix some build errors on GCC 9
- Fail in configure when MPI is not detected

Use `bootstrap.sh` to download and build all UnifyCR's dependencies.  The dependencies are built in `UnifyCR/deps`, and their headers, libs, and binaries are installed to `UnifyCR/install`.   After the dependencies are built, the script prints something like this:
```
*******************************************************************************
Dependencies are all built.  You can now build Unify with:

  export PKG_CONFIG_PATH=/home/hutter/UnifyCR/install/lib/pkgconfig
  ./configure --with-leveldb=/home/hutter/UnifyCR/install --with-gotcha=/home/hutter/UnifyCR/install --with-flatcc=/home/hutter/UnifyCR/install
  make

*******************************************************************************
```

### Motivation and Context
Easily get UnifyCR built on a spack-less system.  Fix GCC 9 build errors.  This doesn't fix the build warnings though, since they're numerous and can be addressed in a separate PR.

### How Has This Been Tested?
Ran `bootstrap.sh` and built Unify on Fedora 30.

### Types of changes

### Checklist:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,106 @@
+#!/bin/bash
+#
+# This is an easy-bake script to download and build all UnifyCR's dependencies.
+#
+
+ROOT="$(pwd)"
+
+mkdir -p deps
+mkdir -p install
+INSTALL_DIR=$ROOT/install
+
+cd deps
+
+repos=(	git://git.mcs.anl.gov/bmi
+	https://github.com/google/leveldb.git
+	https://github.com/LLNL/GOTCHA.git
+	https://github.com/pmodels/argobots.git
+	https://github.com/mercury-hpc/mercury.git
+	https://xgitlab.cels.anl.gov/sds/margo.git
+	https://github.com/dvidelabs/flatcc.git
+)
+
+for i in "${repos[@]}" ; do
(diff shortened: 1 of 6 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,106 @@
...
+			git clone $i
+		fi
+	fi
+done
+
+echo "### building bmi ###"
+cd bmi
+./prepare && ./configure --enable-shared --enable-bmi-only \
+	--prefix="$INSTALL_DIR"
+make -j $(nproc) && make install
```
adammoody: Does $nproc get set in here, or should the user set that before calling the script?

tonyhutter: It's not a variable; we're invoking the `nproc` command here to return the number of CPUs:
```
$ nproc
12
```

adammoody: Oh, gotcha.  Thanks.

adammoody:
Thanks @tonyhutter

PR #347, opened 2019-08-01
Rename UnifyCR to UnifyFS

CamStan:
Rename of all instances of UnifyCR/UNIFYCR/unifycr -> UnifyFS/UNIFYFS/unifyfs within the repo.

### Description

This rename was done with a script that changed all the directory names, then the file names, and then the lines in the files. Each unifycr became unifyfs, UnifyCR became UnifyFS, and UNIFYCR became UNIFYFS. 

The only one not changed was the email address in configure.ac. 

Once merged, links to the docs in the READMEs will break, and links in the docs to files in the repo will break. Changing the url on the Read The Docs admin page will fix the first issue, and renaming the GitHub repo will fix the latter.

Rename took longer than expected as several checkpatch issues showed up that needed to be fixed. Some of these were dealing with comments, which is why there are more lines deleted than added. After fixing them, forced checkpatch to skip the rest since they were skipped before when updating checkpatch standards and we'd decided not to enforce our style rules there (such as in meta/).

### Motivation and Context

Avoid confusion since this is not a checkpoint/restart library.

### How Has This Been Tested?

* Manually built and ran unit tests
* Manually ran continuous integration test suite
* Incrementally started the servers and ran several examples that were already passing.


adammoody:
Thank you, @CamStan !

PR #352, opened 2019-08-12
Fix size bugs, add lamination with chmod/fchmod, misc fixes

tonyhutter:
### Description
This patch does a number of things:

- Wrap `chmod()` and `fchmod()` functions.

- Add in functionality to laminate files via clearing the write bits in chmod.  Add `is_laminated` file meta attribute.  A file that is not laminated will show 0 hard links (`stat.st_nlink`), whereas a laminated file will show 1.  This makes it easy to see if a file is laminated with `stat()`.

- Report final file size and apparent size in `stat()` if file is laminated.   If file is not laminated, return a zero file size.

- Correctly gitignore the gotcha,posix,static files in examples/src/

- Add `size.c` and `chmod.c` test cases.  Add utility functions to `testutil.c`.

- Start `UNIFYFS_ERROR_ENUMERATOR` error numbers after the standard `errno.h` ones.

- Allow for zero length files.

### Motivation and Context
- Provide a way to laminate files
- Make sure size is correctly reported after lamination

### How Has This Been Tested?
Added `chmod.c` and `size.c` test cases

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -278,7 +278,7 @@ int unifyfs_err_map_to_errno(int rc)
     default:
         break;
     }
-    return EIO;
+    return ec;
 }
 
 /* given an errno error code, return corresponding UnifyFS error code */
@@ -649,10 +649,8 @@ static int unifyfs_fid_store_free(int fid)
 int unifyfs_fid_is_dir(int fid)
 {
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-    if (meta) {
-        /* found a file with that id, return value of directory flag */
-        int rc = meta->is_dir;
-        return rc;
+    if (meta && meta->mode & S_IFDIR) {
+        return 1;
     } else {
         /* if it doesn't exist, then it's not a directory? */
         return 0;
@@ -694,6 +692,16 @@ static int unifyfs_gfid_from_fid(const int fid)
     return -EINVAL;
 }
 
+/* Given a fid, return the path.  */
+const char* unifyfs_path_from_fid(int fid)
+{
+    unifyfs_filename_t* fname = &unifyfs_filelist[fid];
+    if (fname->in_use) {
+            return fname->filename;
+    }
+    return NULL;
+}
+
 /* checks to see if a directory is empty
  * assumes that check for is_dir has already been made
  * only checks for full path matches, does not check relative paths,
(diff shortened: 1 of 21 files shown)


tonyhutter:
Oops, forgot to run checkpatch.sh, let me fix those...

client/src/unifyfs.c
```diff
@@ -2431,16 +2455,6 @@ int unifyfs_mount(const char prefix[], int rank, size_t size,
     bool b;
     char* cfgval;
 
-    if (-1 != unifyfs_mounted) {
```
adammoody: do we need to drop this check?

client/src/unifyfs.c
```diff
@@ -858,15 +858,25 @@ int unifyfs_gfid_stat(int gfid, struct stat* buf)
     size_t filesize;
     ret = invoke_client_filesize_rpc(gfid, &filesize);
     if (ret != UNIFYFS_SUCCESS) {
-        return ret;
```
adammoody: There could be a case where the file has been laminated, but the filesize_rpc failed with as error.  For that case, let's still have this return an error.  Can we use the laminated field on the metadata to distinguish instead?

client/src/unifyfs.c
```diff
@@ -858,15 +858,25 @@ int unifyfs_gfid_stat(int gfid, struct stat* buf)
...
+         * That happens when we have a local file that has not been laminated
+         * yet.  In that case, return filesize = 0.
+         */
+        filesize = 0;
     }
 
+    /*
+     * set the file size
+     *
+     * TODO: Why doesn't invoke_client_metaget_rpc() fill in the filesize to
```
adammoody: The answer here is that currently it is still be expensive to look up the file size, because we are fetching *all* key/value pairs and iterating over the full set to compute the max write offset.  However, we may need to get the file meta frequently when we don't care about the file size, so they are separated into two different rpcs.

Once a file has been laminated, we could also record the file size in the metadata, so that if we see the laminated bit set, we also know we can trust the file size field and avoid calling the filesize_rpc.

tonyhutter: Ok, I'm my next push, I'll cache the size when the file is laminated.

client/src/unifyfs.c
tonyhutter: In my next push, I'll cache the size on lamination so this won't be an issue.

client/src/unifyfs.c
tonyhutter: Oops, I must have removed this while debugging something.  I'll add it back in

tonyhutter:
@adammoody I've fixed all your comments and the chechpatch errors in my latest push

adammoody:
This will be very helpful going forward.  Thanks, @tonyhutter .

PR #354, opened 2019-08-23
Add laminate to write example for file size check

CamStan:
### Description

Recent changes to how the file size is calculated (#352) make stat return a size of 0 until the file is laminated. This was causing an error in the write example when file size was checked:

```
[0] ERROR - file size check failed - actual size is 0 B, expected 268435456 B
```

Laminating the file in the write example allows for the size to be correct when stat is called.

### Motivation and Context

Error was causing integration tests to fail.

### How Has This Been Tested?

* Running write example(s) before and after change.
* Running integration tests

### Types of changes

### DIFF
--- a/examples/src/write.c
+++ b/examples/src/write.c
@@ -190,6 +190,13 @@ int main(int argc, char* argv[])
     test_barrier(cfg);
     test_print_verbose_once(cfg, "DEBUG: finished sync");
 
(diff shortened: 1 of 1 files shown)


CamStan:
@tonyhutter, for some reason it won't let me select you as a reviewer, but this is using what you did in #352. Feel free to check this as well.

PR #373, opened 2019-10-04
remove MPI-based remote read code

MichaelBrim:
### Description

This PR removes the code for requesting and servicing remote reads
that was based on MPI. That code was deprecated with the move to
using Margo RPCs. 

The service manager is greatly simplified as a result, and its 
dedicated thread now has a single responsibility, issuing read
response RPCs. The Margo RPC handler threads actually
service the read requests. For now, I have disabled the bursty
data calculation that previously controlled the time to sleep
between servicing read requests, as it's not clear that behavior
is still useful. 

This PR also has some additional code cleanup:
- makes the client use the default log level
- disables a couple server-server RPCs that we don't currently use
- removes socket code deprecated by use of Margo

### Types of changes

### Checklist:


adammoody:
Awesome!  Thanks, @MichaelBrim !

Issue #374, opened 2019-10-08
Design: should fflush() sync extents (it currently doesn't)


tonyhutter:
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | all
OS Version             | all
Architecture           | all
UnifyFS Version        | all

### Describe the problem you're observing
I naively used `fflush()` instead of `fsync()` to sync the extents in a test program.  It didn't sync the extents.  After laminating the file, I could not read the contents.  `fsync()` works though.

Should we make it so `fflush()` behaves the same as `fsync()`?  They're not supposed to behave the same in POSIX (sync to OS buffers, vs sync to disk).  Do we want to cheat and make `fflush()` sync out the extents too?  The benefit would be that if you're using stream IO, you already have a file pointer, so it's easier to sync with `fflush()` than `fsync()`.  However, you can always use `fileno()` to get the file descriptor from the file pointer, which you can then pass to `fsync()`.

### Describe how to reproduce the problem
This works:
```C
    fp = fopen(path, "w");
    rc = fwrite("hello world", 12, 1, fp);
    rc = fclose(fp);
    /* Sync the file with fsync */
    fd = open(path, O_RDWR);
    rc = fsync(fd);
    close(fd);

    /* Laminate it */
    rc = chmod(path, 0444);

    fp = fopen(path, "r");
    fread(buf, sizeof(buf), 1, fp);
    fclose(fp);

    rc = strcmp("hello world", buf);
    ok(rc == 0, "syncing a file with fsync works '%s': %s", buf, strerror(errno));
```

This does not:
```C
    fp = fopen(path, "w");
    rc = fwrite("hello world", 12, 1, fp);
    rc = fflush(fp);
    rc = fclose(fp);

    /* Laminate it */
    rc = chmod(path, 0444);

    fp = fopen(path, "r");
    fread(buf, sizeof(buf), 1, fp);
    fclose(fp);

    rc = strcmp("hello world", buf);
    ok(rc == 0, "syncing a file with fflush works '%s': %s", buf, strerror(errno));
```

MichaelBrim:
I don't see a problem with adding this, as long as we verify that multiple metadata flushes works correctly. I can easily see a use case where people iteratively fwrite and fflush.

adammoody:
Yes, I also think flushing extents on fflush() would be fine.

[...]

PR #375, opened 2019-10-09
Implement fread(), add more read/write test cases

### LINKED ISSUE #367: fread() not implemented for FILE_STORAGE_LOGIO
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | all
OS Version             | all
Architecture           | all
UnifyFS Version        | all

`fread()` needs to be implemented.  More specifically, it needs to be implemented for the `meta->storage == FILE_STORAGE_LOGIO` case, which is the default right now.  An `fread()` currently returns an error. 

### Describe how to reproduce the problem
Do an `fread()`

### Include any warning or errors or relevant debugging data
The `fread()` call path looks like this:

```C
UNIFYFS_WRAP(fread)
	unifyfs_stream_read()
		unifyfs_fd_read()
			unifyfs_fid_read():

			/* determine storage type to read file data */
    			if (meta->storage == FILE_STORAGE_FIXED_CHUNK) {
			        /* file stored in fixed-size chunks */
				rc = unifyfs_fid_store_fixed_read(fid, meta, pos, buf, count);
			} else {
			        /* unknown storage type */
			        rc = (int)UNIFYFS_ERROR_IO;
			}
```

tonyhutter:
### Description
- Implement `fread()` for the `FILE_STORAGE_LOGIO` case

- Add a test case for `fread()`/`fwrite()`.  It also tests some other stream IO calls like fseek()/rewind()/fgets().
 
- Add a test case for `read()`/`write()`/`lseek().

- Add a test case for `fflush()` for #374.  It's currently set to "skip" while we decide what the `fflush()` behaviour should be.

- Add additional `LOGDBG()` logging

These test cases will clear the way for fixing the various file size bugs.  I didn't want to include the file size changes in this PR, since it's already a really big patch.

### Motivation and Context
- Make `fread()` work
- Add test cases.
- Fixes #367 

### How Has This Been Tested?
Added test cases

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-stdio.c
+++ b/client/src/unifyfs-stdio.c
@@ -513,6 +513,7 @@ static int unifyfs_stream_read(
         /* ERROR: invalid file descriptor */
         s->err = 1;
         errno = EBADF;
+        LOGDBG("Invalid file descriptor");
         return UNIFYFS_ERROR_BADF;
     }
 
@@ -520,6 +521,8 @@ static int unifyfs_stream_read(
     if (!filedesc->read) {
         s->err = 1;
         errno = EBADF;
+        LOGDBG("Stream not open for reading");
+
         return UNIFYFS_ERROR_BADF;
     }
 
@@ -531,12 +534,14 @@ static int unifyfs_stream_read(
             /* ERROR: failed to associate buffer */
             s->err = 1;
             errno = unifyfs_err_map_to_errno(setvbuf_rc);
+            LOGDBG("Couldn't setvbuf");
             return setvbuf_rc;
         }
     }
 
     /* don't attempt read if end-of-file indicator is set */
     if (s->eof) {
+        LOGDBG("Stop read, at EOF");
         return UNIFYFS_FAILURE;
     }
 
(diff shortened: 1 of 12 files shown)


client/src/unifyfs-stdio.c
```diff
@@ -594,17 +599,20 @@ static int unifyfs_stream_read(
 
             /* read data from file into buffer */
             size_t bufcount;
-            int read_rc = unifyfs_fd_read(s->fd, current, s->buf, s->bufsize, &bufcount);
-            if (read_rc != UNIFYFS_SUCCESS) {
-                /* ERROR: read error, set error indicator and errno */
+            size_t read_rc = unifyfs_fd_read(s->fd, current, s->buf,
+                s->bufsize);
+            if (read_rc  == -1) {
```
adammoody: I think you mean for this check to be <0 rather than ==-1.  However, I don't think we want that either, because I don't think our UNIFYFS_ERROR_* values are negative.

adammoody: Oh, I see later you updated the return values from fd_read, as well.  Let me keep reviewing.

adammoody:
This looks correct to me.  However, there are a set of unifyfs_fd_* functions that all operate on file descriptors.  We should go back through these and make them consistent.  Now we have:

unifyfs_fd_init() -- operates on a file descriptor, returns UNIFYFS_SUCCESS
unifyfs_fd_write() -- operates on a file descriptor, returns a UNIFYFS_ERROR value, and does not change errno
unifyfs_fd_read() -- operates on a file descriptor, returns bytes read or -1, updates errno
unifyfs_fd_logreadlist() -- this operates on file ids (not file descriptors), and returns UNIFYFS_ERROR values

The unifyfs_fd_logreadlist should probably be renamed to unifyfs_fid_logreadlist or something, since it works with file ids and does not deal with file descriptors at all.

It would be nice to have unifyfs_fd_read and unifyfs_fd_write behave similarly.

MichaelBrim:
@tonyhutter Only wrappers for POSIX should use/return errno values. The eventual client library API will not use errno values for return codes.

adammoody:
Although we're not using them now, the value of having custom error codes is that we can eventually return more useful info about why something failed internally, e.g., the remote service manager timed out / died rather than just returning EIO up through the stack.

MichaelBrim:
@adammoody I'll see if I can rationalize the fd vs fid functions as I start work on the client API. In the meantime, I think it's ok to approve and merge Tony's current code.

adammoody:
Thanks @tonyhutter and @MichaelBrim .

PR #377, opened 2019-10-16
Fix log_size bugs, cleanup *stat(), add local append

tonyhutter:
### Description
- Separate `meta->size` into `meta->local_size` and `meta->global_size`
- Hijack `stat.st_rdev` to store our local_size for debugging
- Cleanup the `stat()` functions.  Add `fstat()` back in.
- Make `fseek(SEEK_END)` and file append work on local files
- Add append mode for both `fopen()` and `open()`.
- Add/update tests to include global and local file size checking, and test append mode.

Note: this depends on #375 being in (I've included that commit here so Travis runs).  Please review #375 before reviewing this.

### Motivation and Context
- Fix the log_size bug
- Separate out global_size and local_size
- Make local append work

### How Has This Been Tested?
Added/updated test cases 

### Types of changes

### Checklist:


client/src/unifyfs-sysio.c
```diff
@@ -487,74 +468,60 @@ int UNIFYFS_WRAP(__fxstat)(int vers, int fd, struct stat* buf)
...
- * or end of file is reached */
-int unifyfs_fd_read(int fd, off_t pos, void* buf, size_t count,
-                    size_t* retcount)
+/*
+ * Read 'count' bytes info 'buf' from file starting at offset 'pos'.
+ *
+ * Returns number of bytes actually read, or -1 on error, in which
+ * case errno will be set.
+ */
+size_t unifyfs_fd_read(int fd, off_t pos, void* buf, size_t count)
```
MichaelBrim: Perhaps change return value to `ssize_t`, like POSIX read, since it supports -1 values without a cast.

client/src/unifyfs-sysio.c
```diff
@@ -333,71 +335,83 @@ int UNIFYFS_WRAP(remove)(const char* path)
...
+    unifyfs_file_attr_t fattr;
+    int ret;
 
-    if (unifyfs_intercept_path(path)) {
-        /* check that caller gave us a buffer to write to */
-        if (!buf) {
-            errno = EFAULT;
-            return -1;
-        }
+    gfid = unifyfs_generate_gfid(path);
```
MichaelBrim: why not use `gfid = unifys_gfid_from_fid(fid)` after getting fid from path?

client/src/unifyfs-sysio.c
```diff
@@ -451,34 +448,18 @@ int UNIFYFS_WRAP(__lxstat)(int vers, const char* path, struct stat* buf)
...
-        }
-
-        /* success */
-        return 0;
+        return __stat(path, buf);
     } else {
-        MAP_OR_FAIL(__lxstat);
-        int ret = UNIFYFS_REAL(__lxstat)(vers, path, buf);
-        return ret;
+        MAP_OR_FAIL(__xstat);
```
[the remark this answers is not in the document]
tonyhutter: Good catch, yes it should be

client/src/unifyfs-sysio.c
tonyhutter: Will do

client/src/unifyfs-sysio.c
tonyhutter: IIRC, it was because the file may have a gfid but not a fid.  Like if you're stating '/unifycr'

adammoody:
I still feel queasy allowing open(O_APPEND) or lseek(SEEK_END) when we don't know the file size.

In this PR, what do we expect for a file with pseudo code like?

```
fd = open(file, O_WRONLY)
lseek(fd, 10, SEEK_SET)
write(fd, 1, 'A')
close(fd)

fd = open(file, O_WRONLY | O_APPEND)
write(fd, 1, 'B')
close(fd)
```

Under POSIX, one would expect to get:
byte offsets: [0-1) hole
byte offsets: [1-2] = "AB"

If we're using local_size to set the position and if local_size = log size, then we'd get:
byte offsets: [0-1) hole
byte offsets: [1-2) "B"
byte offsets: [2-10): hole
byte offsets: [10-11) = "A"

The "B" gets misplaced because the log_size of the file is currently 1 after writing a single char (A).

We can support lseek(SEEK_END) after lamination, and after lamination open(O_APPEND) is invalid since a user can't write to a laminated file.  Before lamination, perhaps we could get fetch the current global size and use that to set the file pointer instead of the local/log size?

adammoody:
I see other parts of the code were modified, so the example of the file contents above is probably not accurate.

Anyway, one size value that we need to track is the current "log size", and I thought that's what local_size was doing here since the log_size field was dropped.  Each write operation just appends data to this running log.  The log size records a pointer to tell us where to append the data for the next write.  It's also needed to know when to allocate another physical chunk of storage to assign to the file in case the log exhausts its current capacity, so that ```log_size <= num_chunks * chunk_size```.

Are we still keeping track of the log size here?

tonyhutter:
> I still feel queasy allowing open(O_APPEND) or lseek(SEEK_END) when we don't know the file size.

You can only O_APPEND on a local file, in which case we do know the local_size.

> In this PR, what do we expect for a file with pseudo code like?

It should look the same as the POSIX case.

> The "B" gets misplaced because the log_size of the file is currently 1 after writing a single char (A).

I don't understand, log_size is removed in this patch

> We can support lseek(SEEK_END) after lamination, and after lamination open(O_APPEND) is invalid since a user can't write to a laminated file.

Correct

> Before lamination, perhaps we could fetch the current global size and use that to
> set the file pointer instead of the local/log size?

The global size is going to be totally in flux and unreliable before lamination.  By the time the global size RPC returns to the client, the global size may have already changed, especially if you have hundreds of nodes writing to the same shared file.  That's why the user should expect all pre-lamination file operations to be relative to their local files only.

client/src/unifyfs-sysio.c
```diff
@@ -636,45 +601,23 @@ int unifyfs_fd_write(int fd, off_t pos, const void* buf, size_t count)
...
     /* finally write specified data to file */
     int write_rc = unifyfs_fid_write(fid, pos, buf, count);
     if (write_rc == 0) {
         meta->needs_sync = 1;
-        if (meta->storage == FILE_STORAGE_LOGIO) {
-            meta->size = newpos;
-            meta->log_size = pos + count;
-        }
+        meta->local_size = MAX(meta->local_size, pos + count);
     }
```
adammoody: The logic that was tracking log size on writes was obfuscated in the old code here.  For UnifyFS, the meta->size field was actually recording the size of the log, not the size of the file.  The unifyfs_fid_size(fid) line here is returning the current size of the log and storing that into "filesize".  So then filesize holds the size of the log, not the file.

We then drop into the "else if storage==LOGIO" branch and compute the size that the log will be after we append the data associated with the current write operation, which we store in newpos ```newpos = filesize + count```.  The newpos value is then passed to unifyfs_fid_extend.  That function determines whether the current set of chunks assigned to the file are sufficient to hold the new data that we're about to append, and if not, it allocates more chunks to the log as needed.

Finally data is appended to the log in fid_write and then the meta->size field is updated to record the new size of the log ```meta->size = newpos```.

The meta->log_size variable was never fully implemented as a replacement, so it was essentially a NOP.

Obviously, all of that logic is hard to make sense of due to the variable names.

tonyhutter:
@adammoody regarding https://github.com/LLNL/UnifyFS/pull/377#issuecomment-543398459, let me look into that.  It sounds like I'm incorrectly using local_size for those chunk allocations when I should resurrect log_size for doing that different size accounting.

client/src/unifyfs-sysio.c
adammoody: Had the old code use better names, this would have been something like:
```
logsize = unifyfs_fid_logsize(fid) // returns meta->log_size
newlogsize = logsize + count // size of log after appending new data
unifyfs_fid_extend(fid, newlogsize) // allocate new chunks for log, if needed
unifyfs_fid_write(fid, pos, but, count) // copy data into log
meta->log_size = newlogsize; // update size of log
```

tonyhutter: Ok, I think I get it.  Log_size is basically the sum of all 'counts' when we do a write(..., count).  It makes sense that that calculation would be used for allocating more chunks.  I'll add `log_size` back in and use it for the allocation.

MichaelBrim: @adammoody thanks for catching this.

tonyhutter:
Ok, I added log_size back in and updated the test cases to print out the log_size.  I also added added in the fixes that @MichaelBrim broght up.

client/src/unifyfs-fixed.c
```diff
@@ -629,8 +629,8 @@ int unifyfs_fid_store_fixed_write(int fid, unifyfs_filemeta_t* meta, off_t pos,
         chunk_id = pos >> unifyfs_chunk_bits;
         chunk_offset = pos & unifyfs_chunk_mask;
     } else if (meta->storage == FILE_STORAGE_LOGIO) {
-        chunk_id = meta->size >> unifyfs_chunk_bits;
-        chunk_offset = meta->size & unifyfs_chunk_mask;
+        chunk_id = meta->local_size >> unifyfs_chunk_bits;
+        chunk_offset = meta->local_size & unifyfs_chunk_mask;
```
adammoody: I think this should be log_size here.

client/src/unifyfs.c
```diff
@@ -1129,8 +1153,8 @@ int unifyfs_fid_extend(int fid, off_t length)
     /* TODO: move this statement elsewhere */
     /* increase file size up to length */
     if (meta->storage == FILE_STORAGE_FIXED_CHUNK) {
-        if (length > meta->size) {
-            meta->size = length;
+        if (length > meta->local_size) {
+            meta->local_size = length;
         }
     }
```
adammoody: I think we could drop this logic, so long as we're increasing the log_size elsewhere.

client/src/unifyfs.c
```diff
@@ -1333,8 +1361,9 @@ int unifyfs_fid_open(const char* path, int flags, mode_t mode, int* outfid,
         }
 
         if (flags & O_APPEND) {
-            unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-            pos = meta->size;
+            meta = unifyfs_get_meta_from_fid(fid);
+            LOGDBG("opening append, meta is %p\n", meta);
+            pos = meta->local_size;
```
adammoody: Let's add a comment here on something to the effect that it's ok to use local_size vs global_size, since we assume the global_size only matters after lamination and someone can't write to a laminated file.

client/src/unifyfs-sysio.c
```diff
@@ -1057,8 +1005,19 @@ ssize_t UNIFYFS_WRAP(write)(int fd, const void* buf, size_t count)
             return (ssize_t)(-1);
         }
 
+        if (filedesc->append) {
+            /*
+             * With O_APPEND we always write to the end, despite the current
+             * file position.
+             */
+            int fid = unifyfs_get_fid_from_fd(fd);
+            pos = unifyfs_fid_local_size(fid);
```
adammoody: Do we need to update filedesc->pos to point to the end of the file?

tonyhutter: We do a few lines down:
```C
      if (filedesc->append) {
            /*
             * With O_APPEND we always write to the end, despite the current
             * file position.
             */
            int fid = unifyfs_get_fid_from_fd(fd);
            pos = unifyfs_fid_local_size(fid);
        } else {
            pos = filedesc->pos;
        }

        /* write data to file */
        int write_rc = unifyfs_fd_write(fd, pos, buf, count);
        if (write_rc != UNIFYFS_SUCCESS) {
            errno = unifyfs_err_map_to_errno(write_rc);
            return (ssize_t)(-1);
        }
        ret = count;

        /* update file position */
        filedesc->pos += ret;
```

adammoody: Is that the correct value, though?  Can someone do something like this?
```
fd = open(path, O_APPEND) // consider that file exists and is already 200 bytes
lseek(fd, 100, SEEK_SET) // can this move filedesc->pos to 100?
write(fd, 1) // write one byte at the end (offset=200)
```
However then ```filedesc->pos += 1``` so that now ```filedesc->pos = 101```.

Should it be 201 instead?  I don't know.

client/src/unifyfs.c
tonyhutter: I agree, but let's git rid of all the FILE_STORAGE_FIXED_CHUNK code in a separate PR.  This patch is getting pretty big..

client/src/unifyfs-fixed.c
tonyhutter: Ok

client/src/unifyfs.c
tonyhutter: Will do

client/src/unifyfs.c
adammoody: sure, sounds good

client/src/unifyfs-sysio.c
tonyhutter: You're correct, for append it should be `filedesc->pos = filesize + count`:
```
O_APPEND The file is opened in append mode. Before each write(2), the file offset is positioned at the end of the file.
```

tonyhutter:
@adammoody I included your new fixes in my latest push.

client/src/unifyfs-sysio.c
adammoody: Thanks @tonyhutter , one more subtle detail.

What happens in the case of a write error?  Based on the man page, it sounds like the position is updated, and then then write is attempted.  Does that mean we need to move the position to the end, regardless of whether the write is successful?  If so, I'm thinking we should do a

```
int fid = unifyfs_get_fid_from_fd(fd);
pos = unifyfs_fid_local_size(fid);
filedesc->pos = pos
```
then later, we could keep the line as:
```
filedesc->pos += ret;
```

client/src/unifyfs-stdio.c
```diff
@@ -704,7 +704,7 @@ static int unifyfs_stream_write(
             errno = EBADF;
             return UNIFYFS_ERROR_BADF;
         }
-        current = unifyfs_fid_size(fid);
+        current = unifyfs_fid_local_size(fid);
```
adammoody: Let's change this to your logical_size() call to future proof a bit.  I think there's a good chance that we'll end up having to deal with the global file size before lamination at some point and the logical_size() call is a good abstraction for that.

client/src/unifyfs-sysio.c
tonyhutter: That make sense

client/src/unifyfs-stdio.c
tonyhutter: ok

tonyhutter:
@adammoody I've included your fixes in my latest push.

adammoody:
My 50/50 guess about updating the offset on a failed write looks to be wrong:
http://man7.org/linux/man-pages/man2/open.2.html
```
The file is opened in append mode.  Before each write(2), the
file offset is positioned at the end of the file, as if with
lseek(2).  The modification of the file offset and the write
operation are performed as a single atomic step.
```
So I think the original code you had was correct.

adammoody:
Thanks @tonyhutter !

tonyhutter:
My latest push restores the old append behaviour

PR #385, opened 2019-10-24
Clean up and portability

boehms:
- remove AC_PROG_RANLIB
- check for dependent libraries early on
- use commoncrypto on MacOS instead of OpenSSL (newer versions do not
  ship with OpenSSL headers)
- Add check for presence of functions to configure (some of the
  functions we wrap do not exist on MacOS and potentially other OS)
- Make Gothca an option (this disables building Gotcha libs and tests)

TODO:
  --wrap is not supported on MacOS, still need to check if the linker
  supports this option and disable static linking if this is not the
  case.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/Makefile.am
+++ b/client/src/Makefile.am
@@ -1,11 +1,14 @@
-lib_LTLIBRARIES = libunifyfs.la libunifyfs_gotcha.la
+lib_LTLIBRARIES = libunifyfs.la
+libunifyfsdir = $(includedir)
+
+if HAVE_GOTCHA
+lib_LTLIBRARIES += libunifyfs_gotcha.la
+libunifyfs_gotchadir = $(includedir)
+endif
 
 if HAVE_FORTRAN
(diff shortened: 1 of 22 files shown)


MichaelBrim:
I like the idea of the configure checks for functions to wrap. Is there any reason not to do the check for all the functions we try to wrap? Would like to see that as a separate PR.

I also am in favor of making the gotcha support optional at configure time. Would like to see that as a separate PR.

The extern/const cleanup and removal of AC_PROG_RANLIB is also good.

I oppose all of the APPLE/MACOS portability changes, since we're targeting Linux supercomputers.

boehms:
I was mainly focused to get it to compile on my Mac. I am looking into making a separate function to simplify the checks for the wrap function. I will add that as soon as I have something that works.

I will add the configure cleanup in a separate PR and we can go from there.

MichaelBrim:
@boehms  can we close this draft now?

Issue #386, opened 2019-10-24
Function to split index into offset ranges hard-coded with 4096 limit


adammoody:
The function here splits a single write index entry into multiple index entries based on slice boundaries (that default to 1MB).
https://github.com/LLNL/UnifyFS/blob/6cc407bd32b0d155e7032cd008681294e0bed7ac/client/src/unifyfs-fixed.c#L404

It has a hard-coded limit of 4096 entries that can be created.

https://github.com/LLNL/UnifyFS/blob/6cc407bd32b0d155e7032cd008681294e0bed7ac/client/src/unifyfs-internal.h#L318
https://github.com/LLNL/UnifyFS/blob/6cc407bd32b0d155e7032cd008681294e0bed7ac/common/src/unifyfs_const.h#L92

There is no check to make sure we don't overflow this limit of 4096.  We need check for overflow and throw an error.

Issue #384, opened 2019-10-24
Handle errors and short writes when writing to spillover


adammoody:
We use pwrite() when copying data to the spillover file, however, we don't properly handle short writes or errors on this call:

https://github.com/LLNL/UnifyFS/blob/6cc407bd32b0d155e7032cd008681294e0bed7ac/client/src/unifyfs-fixed.c#L371

Thanks to @craigsteffen for pointing this out.

PR #388, opened 2019-10-24
client: cleanup index slice code

adammoody:
This simplifies the unifyfs_split_index() code with shorter variable names and adds more comments.

It also adds a check that we don't overflow the output array when splitting a write index (resolves https://github.com/LLNL/UnifyFS/issues/386).  The split routine also now writes index values directly to the shared memory region, rather than a temporary array.  The index_set_t struct that defined the temporary array has been deleted.

A unifyfs_coalesce_index function is defined to attempt to merge the first index in a write with any existing index.  This packages up and simplifies some existing code that was previously inlined in the chunk write function.  Includes an additional (missing) check that log positions are consecutive.  This may have been implied with the fid and file_pos checks, but this feels safer.

Rename mem_pos to log_pos in unifyfs_index_t structure for clarity, since some log offsets fall in the spill over file, which is not memory.

Handle short writes and errors in pwrite to spillover file (resolves https://github.com/LLNL/UnifyFS/issues/384).

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-fixed.c
+++ b/client/src/unifyfs-fixed.c
@@ -365,15 +365,6 @@ static int unifyfs_split_index(
     /* initialize count of output index entries */
     off_t count = 0;
 
-    /* no room to write index values */
-    if (count >= maxcount) {
-        /* no room to write more index values,
-         * and we have at least one more,
-         * record number we wrote and return with error */
-        *used_count = 0;
-        return UNIFYFS_FAILURE;
-    }
-
     /* define new index entries in index_set by splitting write index
      * at slice boundaries */
     if (idx_end <= slice_end) {
@@ -559,9 +550,6 @@ static int unifyfs_logio_chunk_write(
     /* lookup number of existing index entries */
     off_t num_entries = *(unifyfs_indices.ptr_num_entries);
 
-    /* remaining entries we can fit in the shared memory region */
-    off_t remaining_entries = unifyfs_max_index_entries - num_entries;
-
     /* get pointer to index array */
     unifyfs_index_t* idxs = unifyfs_indices.index_entry;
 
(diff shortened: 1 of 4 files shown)


adammoody:
Ready for review on this one.

client/src/unifyfs-fixed.c
```diff
@@ -264,74 +264,209 @@ static int unifyfs_chunk_read(
...
+    long slice_end = slice_start + slice_range - 1;
 
+    /* get pointer to first output index structure */
+    unifyfs_index_t* set = index_set;
+
+    /* initialize count of output index entries */
+    off_t count = 0;
+
+    /* no room to write index values */
+    if (count >= maxcount) {
```
MichaelBrim: How can this condition ever be true? Seems only when `maxcount == 0`, in which case we shouldn't call this function in the first place.

client/src/unifyfs-fixed.c
```diff
@@ -397,83 +553,53 @@ static int unifyfs_logio_chunk_write(
...
-                     * last index in list */
-                    prev_idx->length  += next_idx->length;
-
-                    /* advance to next index in temp list */
-                    i++;
-                }
-            }
-        }
+    /* remaining entries we can fit in the shared memory region */
+    off_t remaining_entries = unifyfs_max_index_entries - num_entries;
```
MichaelBrim: Re: prior comment. Should bail out with error here if `remaining_entries == 0`.

adammoody: There is one corner case in which the write index can be fully coalesced with the last index, so that it doesn't require adding an additional entry.  We need to throw an error if ```remaining_entries==0``` and we need to call the split routine, though.  Currently that logic is inside of the split function, but I'll pull it out here.

MichaelBrim: ok, sounds good

adammoody:
Pushed a new commit with those changes.

By the way, I'll look at modifying this to fsync when we exhaust the entries buffer.  That will turn all of this more into a while loop, so something like:
```
while (cur_idx.length > 0) {
  int exhausted_flag = 0
  split_index(..., &exhausted_flag)
  if (exhausted_flag) {
    fsync with server
    num_entries=0
    remaining_entries = max_entries
  }
}
```

adammoody:
@MichaelBrim , I think I changed the comments.  Nice catch.

adammoody:
Thanks for reviewing, @MichaelBrim !

PR #392, opened 2019-10-25
server: clean up fsync rpc handler

adammoody:
Cleanup server code that inserts file extents into MDHIM with shorter variable names and more comments.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_request_manager.c
+++ b/server/src/unifyfs_request_manager.c
@@ -944,7 +944,7 @@ int rm_cmd_fsync(int app_id, int client_side_id, int gfid)
 
         /* MDHIM needs to know the byte size of each key and value */
         fattr_key_lens[i] = sizeof(fattr_key_t);
-        fattr_val_lens[i] = sizeof(fattr_val_t);
+        fattr_val_lens[i] = sizeof(unifyfs_file_attr_t);
     }
 
     /* batch insert file attribute key/values into MDHIM */
--- a/server/src/unifyfs_request_manager.c
+++ b/server/src/unifyfs_request_manager.c
@@ -816,121 +816,149 @@ int rm_cmd_exit(reqmgr_thrd_t* thrd_ctrl)
  */
 int rm_cmd_fsync(int app_id, int client_side_id, int gfid)
 {
-    int ret = 0;
+    size_t i;
 
-    unifyfs_key_t** unifyfs_keys;
-    unifyfs_val_t** unifyfs_vals;
-    int* unifyfs_key_lens = NULL;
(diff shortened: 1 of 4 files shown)


adammoody:
While reading this function to refresh my mind and clean it up in passing, I found that we're storing the address to metadata info.  I'm pretty sure this is buggy and needs to be fixed.  I suspect we could just drop dealing with inserting file metadata info in this step.  I need to dig further.

server/src/unifyfs_request_manager.c
```diff
@@ -816,121 +816,151 @@ int rm_cmd_exit(reqmgr_thrd_t* thrd_ctrl)
...
         goto rm_cmd_fsync_exit;
     }
 
+    /* create file attribute key/values for insertion into MDHIM */
     for (i = 0; i < attr_num_entries; i++) {
+        /* for a key, we use the global file id */
         *fattr_keys[i] = attr_payload[i].gfid;
+
+        /* for the value, we use address of file attribute? */
+        /* TODO: this means each server must have a local copy for each file,
```
MichaelBrim: MDHIM doesn't store the pointer as the val, it copies the data at the given pointer with the given size. There is a bug here though, since the ptr is to a unifyfs_file_attr_t struct, but we give the size of fattr_val_t struct.

server/src/unifyfs_request_manager.c
```diff
@@ -816,121 +816,151 @@ int rm_cmd_exit(reqmgr_thrd_t* thrd_ctrl)
...
+         * and they must be consistent across servers */
         fattr_vals[i] = &(attr_payload[i]);
+
+        /* MDHIM needs to know the byte size of each key and value */
         fattr_key_lens[i] = sizeof(fattr_key_t);
         fattr_val_lens[i] = sizeof(fattr_val_t);
```
MichaelBrim: wrong struct size here

server/src/unifyfs_request_manager.c
adammoody: Oh, yeah I just realized that's what MDHIM was doing.

Let me check the difference on structs.

MichaelBrim: AFAICT, fattr_val_t is no longer used for anything, and should be removed.

server/src/unifyfs_request_manager.c
adammoody: Yep, nice catch!  I'll fix that up, too.

server/src/unifyfs_request_manager.c
adammoody: Yes, I concur.  Doing that now.

Issue #387, opened 2019-10-24
Support flushing of write index entries when buffer is full


adammoody:
While appending write index entries into the shared memory buffer on the client, there is a check to make sure we don't add more entries than it can hold (currently around 655,360).  We print an error, but do not throw an error once this buffer is exhausted.

https://github.com/LLNL/UnifyFS/blob/6cc407bd32b0d155e7032cd008681294e0bed7ac/client/src/unifyfs-fixed.c#L471

To allow for more write entries, we could fsync_rpc to the server, wait for the sync to complete to know that the server has copied the contents out of the buffer, and then clear the buffer to write additional entries at the beginning.

PR #393, opened 2019-10-25
server: flush extent buffer to server when full

adammoody:
This will flush the file extent index buffer to the server whenever it fills up (resolves https://github.com/LLNL/UnifyFS/issues/387).

Currently, this is using fsync(gfid) for the current file.  That so happens to flush all extents, regardless of gfid.  If fsync is made to adhere to only flushing data for the given file, this needs to be changed to flush all extents in the buffer or otherwise each file should have space to store its own extents.

This uses the unifyfs_gfid_from_fid(fid) function to lookup global file id values everywhere, eliminating get_gfid() and calls to bsearch() to search the file attributes in the shared memory segment (resolves https://github.com/LLNL/UnifyFS/issues/394).

This changes the server to avoid setting file attributes on fsync, and it drops the shared memory segment used to transfer file attributes from the client to the server during fsync.  A while ago, we switched to a mode where file attributes are set by the client via rpc when a file is created.  The shared memory buffer is no longer needed.

### Types of changes

### Checklist:


MichaelBrim:
Not really a comment on your changes, but something I noticed while looking at them. There are a lot of places where we assume the pointer returned from `unifyfs_get_meta_from_fid(fid)` is non-NULL, even though it could be NULL in theory if someone passed it a bad fid. Maybe we should change the `return NULL` case in that function to be an assert() to avoid segfaults.

adammoody:
Yeah, adding a assert there is a good idea.

adammoody:
Good point, I'll keep those as int32_t.  Thanks.

adammoody:
Thanks for your review and suggestions, @MichaelBrim .

PR #390, opened 2019-10-25
Clean up configure.ac

boehms:
- Add check for -wrap in linker
- Add checks for wrapped functions
- make gotcha optional
- remove leveldb and flatcc from the optional dependencies

### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Performance enhancement (non-breaking change which improves efficiency)
- [x] Code cleanup (non-breaking change which makes code smaller or more readable)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation (a change to man pages or other documentation)

### Checklist:

- [ ] My code follows the UnifyFS code style requirements.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] All commit messages are properly formatted.

PR #398, opened 2019-10-30
Put local and log sizes in st_rdev, update bootstrap.sh

### LINKED ISSUE #396: Don't use stat.st_dev to store log_size for test cases
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | all
OS Version             | all
Architecture           | all
UnifyFS Version        | dev

My commit #377 hijacks `stat.st_dev` to store the log size for the file:

https://github.com/LLNL/UnifyFS/commit/49790fb7fd53ba5cdc293c08944e945f166957d5#diff-6fee1f4292bfeb1841b685854ce3f389R372

We use this to check the correct `log_size` in our test cases.  However, `st_dev` should be the device number for the underlying device for the file.  We now know that HDF5 is expecting `st_dev` to be sane, so we're going to have to change this.  The quick fix would be to assume stat_t is the 64-bit variant (which is probably a safe assumption), and use the upper 32-bits of `st_rdev` to store the log size, and the lower 32-bit to store the local_size.  That way we can still get those size values to the test cases.  Long term, we may want to consider using xattrs for local size and log size.

tonyhutter:
### Description
Previously we stored the local size in `stat.st_rdev` and the log size in `stat.st_dev`.  However, HDF5 expects sane values in `st_dev`, so move log size into the upper 32-bits of `st_rdev`.

Also update bootstrap.sh to export `LEVELDB_ROOT` and `FLATCC_ROOT` instead of having them as `--with-leveldb` and `--with-flatcc`, in accordance with #390.  Along with that, add `--prefix` line to point to install dir.

### Motivation and Context
- Fixes #396
- Makes boostrap.sh work again.

### How Has This Been Tested?
Tested bootstrap.sh
Passed 'make check;

### Types of changes

### Checklist:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -104,11 +104,12 @@ cd ..
 cd "$ROOT"
 
 echo "*************************************************************************"
-echo "Dependencies are all built.  You can now build Unify with:"
+echo "Dependencies are all built.  You can now build UnifyFS with:"
 echo ""
-echo "  export PKG_CONFIG_PATH=$INSTALL_DIR/lib/pkgconfig"
-echo "  ./autogen.sh && ./configure --with-leveldb=$INSTALL_DIR" \
-	"--with-gotcha=$INSTALL_DIR --with-flatcc=$INSTALL_DIR"
+echo -n "  export PKG_CONFIG_PATH=$INSTALL_DIR/lib/pkgconfig && "
+echo "export LEVELDB_ROOT=$INSTALL_DIR && export FLATCC_ROOT=$INSTALL_DIR"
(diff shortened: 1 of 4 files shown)


MichaelBrim:
@tonyhutter should we add a configure check to make sure `sizeof(st_rdev)` is 64?

tonyhutter:
@MichaelBrim that's a good idea.  Or maybe just a general check for a 64-bit build environment.  I doubt anyone has tested Unify on a 32-bit machine.

adammoody:
Thanks, @tonyhutter , @MichaelBrim and @CamStan .  I'll get this in, and we can add the configure check later.

PR #408, opened 2019-11-12
fix EOF for remote reads after laminate

MichaelBrim:
### Description
This fixes the writeread example in io_shuffle mode, where each process reads different data than it wrote. Previously, we were seeing EOFs on reads with offsets that exceeded the locally cached file size, which was based on what was written. The cached file size was being used because processes other than rank 0 did not mark the file as laminated. The fix was to add a `stat()` call for the non-zero ranks, and to make sure that the associated call to `unifyfs_get_global_file_meta()` caused the local metadata to be correctly updated, which wasn't the case before.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-internal.h
+++ b/client/src/unifyfs-internal.h
@@ -500,11 +500,8 @@ off_t unifyfs_fid_log_size(int fid);
  */
 off_t unifyfs_fid_logical_size(int fid);
 
-/* fill in limited amount of stat information for global file id */
-int unifyfs_gfid_stat(int gfid, struct stat* buf);
-
-/* fill in limited amount of stat information */
-int unifyfs_fid_stat(int fid, struct stat* buf);
+/* Update local metadata for file from global metadata */
+int unifyfs_fid_update_file_meta(int fid, unifyfs_file_attr_t* gfattr);
 
 /* allocate a file id slot for a new file
  * return the fid or -1 on error */
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -359,7 +359,9 @@ static int __stat(const char* path, struct stat* buf)
 
     /* update local file metadata (if applicable) */
     fid = unifyfs_get_fid_from_path(path);
-    unifyfs_fid_update_file_meta(fid, &fattr);
+    if (fid != -1) {
+        unifyfs_fid_update_file_meta(fid, &fattr);
+    }
 
     /* copy attributes to stat struct */
     unifyfs_file_attr_to_stat(&fattr, buf);
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -338,50 +338,48 @@ int UNIFYFS_WRAP(remove)(const char* path)
 /* The main stat call for all the *stat() functions */
 static int __stat(const char* path, struct stat* buf)
 {
-    int gfid, fid;
+    int gfid, fid, ret;
     unifyfs_file_attr_t fattr;
-    int ret;
-
-    gfid = unifyfs_generate_gfid(path);
-    fid = unifyfs_get_fid_from_path(path);
 
     /* check that caller gave us a buffer to write to */
     if (!buf) {
         errno = EFAULT;
         return -1;
     }
+    memset(buf, 0, sizeof(*buf));
 
     /* lookup stat data for global file id */
-    ret = invoke_client_metaget_rpc(gfid, &fattr);
+    gfid = unifyfs_generate_gfid(path);
+    ret = unifyfs_get_global_file_meta(gfid, &fattr);
     if (ret != UNIFYFS_SUCCESS) {
         LOGDBG("metaget failed");
-        return ret;
+        errno = ENOENT;
+        return -1;
     }
 
-    memset(buf, 0, sizeof(*buf));
+    /* update local file metadata (if applicable) */
+    fid = unifyfs_get_fid_from_path(path);
+    unifyfs_fid_update_file_meta(fid, &fattr);
 
     /* copy attributes to stat struct */
     unifyfs_file_attr_to_stat(&fattr, buf);
 
-    /*
-     * For debugging and testing purposes, we hijack st_rdev to store our
-     * local size and log size.  We also assume the stat struct is
-     * the 64-bit variant.  The values are stored as:
-     *
-     * st_rdev = log_size << 32 | local_size;
-     *
-     */
-    buf->st_rdev = fid;
     if (fid >= 0) { /* If we have a local file */
-        buf->st_rdev = (unifyfs_fid_log_size(fid) << 32) |
-            (unifyfs_fid_local_size(fid) & 0xFFFFFFFF);
+        /*
+         * For debugging and testing purposes, we hijack st_rdev to store our
+         * local size and log size.  We also assume the stat struct is
+         * the 64-bit variant.  The values are stored as:
+         *
+         * st_rdev = log_size << 32 | local_size;
+         *
+         */
+        buf->st_rdev  = (unifyfs_fid_log_size(fid) << 32);
+        buf->st_rdev |= (unifyfs_fid_local_size(fid) & 0xFFFFFFFF);
     }
 
+    /* global filesize is zero for non-laminated files */
     if (!fattr.is_laminated) {
-        /*
-         * It was decided that all non-laminated files would report a global
-         * filesize of zero.
-         */
+        LOGDBG("file is NOT laminated")
         buf->st_size = 0;
     }
 
@@ -544,6 +542,7 @@ ssize_t unifyfs_fd_read(int fd, off_t pos, void* buf, size_t count)
 
     /* if we don't read any bytes, return success */
     if (count == 0) {
+        LOGDBG("returning EOF");
         return 0;
     }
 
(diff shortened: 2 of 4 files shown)


client/src/unifyfs-sysio.c
```diff
@@ -338,50 +338,45 @@ int UNIFYFS_WRAP(remove)(const char* path)
...
     if (!buf) {
         errno = EFAULT;
         return -1;
     }
+    memset(buf, 0, sizeof(*buf));
 
     /* lookup stat data for global file id */
-    ret = invoke_client_metaget_rpc(gfid, &fattr);
+    gfid = unifyfs_generate_gfid(path);
+    fid = unifyfs_get_fid_from_path(path);
```
adammoody: This PR does not introduce the issue, but what happens in the case that a process stats a file that it has never opened?  I think it would not have a valid fid, so this will likely return -1.  I think that process may then see an invalid global file size.

client/src/unifyfs-internal.h
```diff
@@ -567,6 +561,7 @@ int unifyfs_set_global_file_meta(int gfid,
                                  unifyfs_file_attr_t* gfattr);
 
 int unifyfs_get_global_file_meta(int gfid,
+                                 int fid,
```
adammoody: Can we create a wrapper around unifyfs_get_global_file_meta that takes the fid?

To detangle the code, I am hoping we can keep unifyfs_set/get_global_meta clean in that they only know about gfid and a related input/output file_attr struct.

client/src/unifyfs-sysio.c
adammoody: And an issue the writeread case brings up which this is trying to fix.  I guess we haven't defined or documented when or how a process can detect that a file it already has open has been laminated.

client/src/unifyfs-internal.h
MichaelBrim: yes, we could use a wrapper to separate the implementation concerns. we would still need to call the wrapper in all cases to make sure the local metadata gets updated.

client/src/unifyfs-sysio.c
MichaelBrim: if the fid is -1 (i.e., no local metadata), then we don't try to update the local metadata.

there were a couple of other places that I fixed where we were assuming a good fid was being passed in a call to get the local metadata, and the resultant (possibly NULL) pointer was being dereferenced.

MichaelBrim: you're right about the undefined nature of discovering lamination status. with current code, you have to laminate (chmod from rank0), barrier, then stat() from other ranks to ensure a consistent view, which is what I modified our testutil `write_laminate()` to do.

I suppose we could punt for now and say only the rank that does the laminate knows the status, and it's up to application-level logic to ensure a consistent view. the alternative requires us to push the new status to processes that have it open, which I can't think of a reasonable way to accomplish.

MichaelBrim:
looks like travis failing due to the spack environment modules issue.

@adammoody can you merge #405? kathryn was going to but I guess she forgot.

client/src/unifyfs-internal.h
MichaelBrim: @adammoody I pushed a new version that creates a helper method to update local metadata from global metadata. This made more sense to me than a wrapper, since it allows us to be more precise in deciding when to do the local update.

client/src/unifyfs-sysio.c
adammoody: Sounds good.  Thanks, @MichaelBrim .  That also pairs up well as the reverse of this function which sets the global file metadata based on the local metadata for a fid:
https://github.com/LLNL/UnifyFS/blob/66448e0d791362bc303a31a07537d03c5a1c3635/client/src/unifyfs-internal.h#L564

client/src/unifyfs.c
```diff
@@ -740,15 +740,21 @@ off_t unifyfs_fid_global_size(int fid)
 {
     /* get meta data for this file */
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-    return meta->global_size;
+    if (NULL != meta) {
+        return meta->global_size;
+    }
+    return (off_t)-1;
```
tonyhutter: I'd replace with:
```C
assert(meta)
return meta->global_size
```
If `meta` is NULL then it's a hard bug.

client/src/unifyfs.c
```diff
@@ -740,15 +740,21 @@ off_t unifyfs_fid_global_size(int fid)
...
 off_t unifyfs_fid_local_size(int fid)
 {
     /* get meta data for this file */
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-    return meta->local_size;
+    if (NULL != meta) {
+        return meta->local_size;
+    }
+    return (off_t)-1;
 }
```
tonyhutter: assert(meta)
return meta->global_size

client/src/unifyfs.c
```diff
@@ -1288,8 +1262,11 @@ int unifyfs_fid_open(const char* path, int flags, mode_t mode, int* outfid,
         }
 
         /* initialize global size of file from global metadata */
-        unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-        meta->global_size = gfattr.size;
+        ret = unifyfs_fid_update_file_meta(fid, &gfattr);
+        if (ret != UNIFYFS_SUCCESS) {
+            LOGERR("failed to update file metadata from global");
+            /* this is not a fatal error, continue */
```
tonyhutter: Why wouldn't this be a fatal error?  We need to know if the file is laminated on open().

client/src/unifyfs-sysio.c
```diff
@@ -338,50 +338,48 @@ int UNIFYFS_WRAP(remove)(const char* path)
...
+        buf->st_rdev  = (unifyfs_fid_log_size(fid) << 32);
+        buf->st_rdev |= (unifyfs_fid_local_size(fid) & 0xFFFFFFFF);
     }
 
+    /* global filesize is zero for non-laminated files */
     if (!fattr.is_laminated) {
-        /*
-         * It was decided that all non-laminated files would report a global
-         * filesize of zero.
-         */
```
tonyhutter: Please keep this comment in.

client/src/unifyfs-sysio.c
```diff
@@ -338,50 +338,48 @@ int UNIFYFS_WRAP(remove)(const char* path)
...
+        LOGDBG("file is NOT laminated")
```
tonyhutter: Is this LOGDBG still necessary?

client/src/unifyfs.c
MichaelBrim: We can't assume the fid that is passed is valid. My opinion for libraries linked into client applications is that they should never assert(). Rather, they should always return with an error.

client/src/unifyfs.c
MichaelBrim: Because fid will only be valid if we already have the file open locally.

client/src/unifyfs-sysio.c
MichaelBrim: necessary, no. but still useful for future debugging.

client/src/unifyfs.c
MichaelBrim: the short answer is that the return code check really isn't needed, because we already know we have a valid fid and non-null pointer, which are the only reasons the function could return an error. I actually thought about leaving out the check, but left it because it was more consistent with other usage.

client/src/unifyfs.c
MichaelBrim: see above comment

client/src/unifyfs-sysio.c
MichaelBrim: I kept the comment, just shortened it and moved it above the if.
"global filesize is zero for non-laminated files"

client/src/unifyfs.c
tonyhutter: I would say that in general, we should be checking the fid in the higher level functions that the user calls (like `fstat()`) and returning the error there.  That way we can return the error as early as possible.  You could add a comment above the unifyfs_fid_*_size() functions saying  "we assume fid has been verified" and keep the assert in for sanity.  That way we're not unnecessarily re-checking fid at every level of the code.  

That said, I'm ok with the way you have it.

client/src/unifyfs.c
tonyhutter: Ok, then I would just add a comment above `unifyfs_fid_update_file_meta()` saying "we assume fid and gfattr have already been verified", and remove the check and the return code.  That simplifies the function, and removes the need for callers to verify the `unifyfs_fid_update_file_meta()` return code.

client/src/unifyfs-sysio.c
tonyhutter: Ok

client/src/unifyfs-sysio.c
tonyhutter: Ok

client/src/unifyfs.c
MichaelBrim: I agree that the higher level functions should verify the fid, and fid checking at multiple levels is redundant. If we want to treat the `unifyfs_fid_xxx_size()` functions as simple getters, my suggestion would be to pass in the meta pointer rather than the fid.

client/src/unifyfs.c
MichaelBrim: For this particular case we know the parameters are valid. There are other cases where the function may be called with `fid == -1`, so we can't just assume all uses will be good. I can remove the check for this use though.

Issue #411, opened 2019-12-03
Corruption while doing many writes to multiple open files


tonyhutter:
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | Linux
OS Version             | TOSS (modified RHEL 7.5)
Architecture           | x86_64
UnifyFS Version        | dev

### Describe the problem you're observing
While testing out my "flattening writes" code, I decided to write a test case to do a lot of random writes to a bunch of open files and verify the data was correct.  When I ran the test, I noticed what appears to be corrupted data when running the test with 20 open files and ~10k writes.  I don't seem to get the corruption as much or at all if I lower the number of open files to ~10 or so.  Below is the result of 6 runs, with 5 of them showing corruption, and one showing success.   Just looking at the data, it appears the bad writes may be going to the wrong offsets (or they're being read back wrong).  The reproducer is here https://github.com/tonyhutter/UnifyFS/blob/corrupt-data/examples/src/corrupt.c
```
/unifyfs/tmp.q3q1N8X0K8: read 1046736 bytes
/unifyfs/tmp.q3q1N8X0K8 failed at offset 5204 (tmpbuf['I'] != bigbuf['E'])
Comparing last 10 bytes before/after:
expected: UVWXYZABCD|E|FGHIJKLMNO
got:                |I|JKLMNOPQ  
file 7/20 failed


/unifyfs/tmp.iMVbmI3GP4: read 1045839 bytes
/unifyfs/tmp.iMVbmI3GP4 failed at offset 1607 (tmpbuf['I'] != bigbuf['V'])
Comparing last 10 bytes before/after:
expected: LMNOPQRSTU|V|WXYZABCDEF
got:                |I|JKLMNOPQRS
file 1/20 failed


/unifyfs/tmp.yQTL6L7eFT: read 1048150 bytes
/unifyfs/tmp.yQTL6L7eFT failed at offset 834 (tmpbuf['I'] != bigbuf['C'])
Comparing last 10 bytes before/after:
expected: STUVWXYZAB|C|DEFGHIJKLM
got:                |I|JKLMNOPQRS
file 1/20 failed


/unifyfs/tmp.2jweY5gncF: read 1047859 bytes
/unifyfs/tmp.2jweY5gncF failed at offset 2366 (tmpbuf['I'] != bigbuf['A'])
Comparing last 10 bytes before/after:
expected: QRSTUVWXYZ|A|BCDEFGHIJK
got:                |I|JKLMNOPQRS
[...]

PR #414, opened 2019-12-05
Add write flattening

tonyhutter:
### Description

This flattens all the writes on an fsync to reduce the number of extents transferred.  

### Motivation and Context
Flattenting writes reduces the number of extents when you're doing lots of overwrites.  For example, let's say you did five writes:
```
[w1---------------]
         [w2-----------]
                [w3-------------]
                [w4-------------]
    [w5----------------]
```
After flattening:
```
[w1][w5----------------][w4-----]
```
This also means that reads are faster, since there's fewer extents to iterate over.  You can disable write flattening via the UNIFYFS_CLIENT_FLATTEN_WRITES config value (enabled by default).

Other notes:

- The red-black tree implementation (tree.h) included in this commit was taken from OpenBSD:
    
http://cvsweb.openbsd.org/cgi-bin/cvsweb/~checkout~/src/sys/sys/tree.h
    
tree.h's macros are valid but checkpatch still flags them.  It's exempted via the commit message:
TEST_CHECKPATCH_SKIP_FILES=common/src/tree.h
    
- Cameron's travis.yml fix is included.

### Motivation and Context
Better performance on overwrite heavy workloads

### How Has This Been Tested?
Tested via manually running the added `examples/src/multi-write.c` example code.  Due to hitting an unrelated corruption bug (https://github.com/LLNL/UnifyFS/issues/411), I've left `multi-write.c` as example code, not a test case.  Once the bug is fixed, we should consider converting it to a test case.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-fixed.c
+++ b/client/src/unifyfs-fixed.c
@@ -44,6 +44,7 @@
 #include "unifyfs-fixed.h"
 #include "unifyfs_log.h"
 #include "margo_client.h"
+#include "seg_tree.h"
 
 static inline
 unifyfs_chunkmeta_t* filemeta_get_chunkmeta(const unifyfs_filemeta_t* meta,
@@ -336,12 +337,22 @@ static int unifyfs_coalesce_index(
     return UNIFYFS_SUCCESS;
 }
 
-/* given an index, split it into multiple indices whose range is equal or
- * smaller than slice_range size
- * @param  cur_idx:     the index to split
- * @param  slice_range: the slice size of the key-value store
- * @return index_set:   the set of split indices
- * @param  maxcount:     number of entries in output array */
+/*
+ * Given an index, split it into multiple indices whose range is equal or
+ * smaller than slice_range size.  For example, if you passed a cur_index
+ * for a 3.5MB write, and the slice size was 1MB, it would split it into
+ * four indexes, and update cur_index.length to be zero.  This also takes
+ * in a 'maxcount' field, so you can limit the number of indexes you
+ * create.  Using our above example, if 'maxcount=2', then this would
+ * create two indexes, and update cur_index.length to 1.5MB (for the remaining
+ * data).
+ *
+ * @param  cur_idx:     The index to split
+ * @param  slice_range: The slice size of the key-value store
+ * @return index_set:   The set of split indices
+ * @param  maxcount:    Number of entries in output array
+ * @param  used_count:  Number of entries we actually added in the split
+ */
 static int unifyfs_split_index(
     unifyfs_index_t* cur_idx,   /* write index to split (offset and length) */
     long slice_range,           /* number of bytes in each slice */
@@ -471,6 +482,152 @@ static int unifyfs_split_index(
     return UNIFYFS_SUCCESS;
 }
 
+/*
+ * Clear all entries in the log index.  This only clears the metadata,
+ * not the data itself.
+ */
+static void
+unifyfs_clear_index(void)
+{
+    *unifyfs_indices.ptr_num_entries = 0;
+}
+
+/*
+ * Sync all the extents to the server.  Clears the metadata index afterwards.
+ *
+ * Returns 0 on success, nonzero otherwise.
+ */
+int
+unifyfs_sync(int gfid)
+{
+    if (unifyfs_flatten_writes) {
+        unifyfs_rewrite_index_from_seg_tree();
+    }
+
+    int ret = invoke_client_fsync_rpc(gfid);
+    if (ret != UNIFYFS_SUCCESS) {
+        /* something went wrong when trying to flush key/values */
+        LOGERR("failed to flush key/value index to server");
+        return UNIFYFS_ERROR_IO;
+    }
+
+    /*
+     * flushed, clear buffer and refresh number of entries
+     * and number remaining
+     */
+    unifyfs_clear_index();
+
+    return UNIFYFS_SUCCESS;
+}
+
+void
+unifyfs_add_index_entry_to_seg_tree(unifyfs_filemeta_t* meta,
+    unifyfs_index_t* index)
+{
+    if (!unifyfs_flatten_writes) {
+        /* We're not flattening writes.  Nothing to do */
+        return;
+    }
+
+    /*
+     * Store the write in our segment tree.  We will later use this for
+     * flattening writes.
+     */
+    seg_tree_add(&meta->seg_tree, index->file_pos,
+        index->file_pos + index->length - 1,
+        index->log_pos);
+
+    /*
+     * We want to make sure the next write following this wont overflow the
+     * max number of index entries (if it were synced).  A write can at most
+     * create two new nodes in the seg_tree.  If we're close to potentially
+     * filling up the index, sync it out.
+     */
+    if (seg_tree_count(&meta->seg_tree) >= (unifyfs_max_index_entries - 2)) {
+        unifyfs_sync(meta->gfid);
+    }
+}
+
+/* Add the metadata for a single write to the index */
+static int unifyfs_logio_add_write_meta_to_index(unifyfs_filemeta_t* meta,
+    off_t file_pos, off_t log_pos, size_t length)
+{
+    /* define an new index entry for this write operation */
+    int gfid = meta->gfid;
+    unifyfs_index_t cur_idx;
+    cur_idx.gfid     = gfid;
+    cur_idx.file_pos = file_pos;
+    cur_idx.log_pos  = log_pos;
+    cur_idx.length   = length;
+
+    /* lookup number of existing index entries */
+    off_t num_entries = *(unifyfs_indices.ptr_num_entries);
+
+    /* get pointer to index array */
+    unifyfs_index_t* idxs = unifyfs_indices.index_entry;
+
+    /* attempt to coalesce contiguous index entries if we
+     * have an existing index in the buffer */
+    if (num_entries > 0) {
+        /* get pointer to last element in index array */
+        unifyfs_index_t* prev_idx = &idxs[num_entries - 1];
+
+        /* attempt to coalesce current index with last index,
+         * updates fields in last index and current index
+         * accordingly */
+        unifyfs_coalesce_index(prev_idx, &cur_idx,
+            unifyfs_key_slice_range);
+        if (cur_idx.length == 0) {
+            /* We were able to coalesce this write into prev_idx */
+            unifyfs_add_index_entry_to_seg_tree(meta, prev_idx);
+        }
+    }
+
+    /* add new index entries if needed */
+    while (cur_idx.length > 0) {
+        /* remaining entries we can fit in the shared memory region */
+        off_t remaining_entries = unifyfs_max_index_entries - num_entries;
+
+        /* if we have filled the key/value buffer, flush it to server */
+        if (0 == remaining_entries) {
+            /* index buffer is full, flush it */
+            int ret = unifyfs_sync(cur_idx.gfid);
+            if (ret != UNIFYFS_SUCCESS) {
+                /* something went wrong when trying to flush key/values */
+                LOGERR("failed to flush key/value index to server");
+                return UNIFYFS_ERROR_IO;
+            }
+        }
+
+        /* split any remaining write index at boundaries of
+         * unifyfs_key_slice_range */
+        off_t used_entries = 0;
+        int split_rc = unifyfs_split_index(&cur_idx,
+            unifyfs_key_slice_range, &idxs[num_entries],
+            remaining_entries, &used_entries);
+        if (split_rc != UNIFYFS_SUCCESS) {
+            /* in this case, we have copied data to the log,
+             * but we failed to generate index entries,
+             * we're returning with an error and leaving the data
+             * in the log */
+            LOGERR("failed to split write index");
+            return UNIFYFS_ERROR_IO;
+        }
+
+        /* Add our split index entries to our seg_tree */
+        for (int i = 0; i < used_entries; i++) {
+            unifyfs_add_index_entry_to_seg_tree(meta, &idxs[num_entries + i]);
+        }
+
+        /* account for entries we just added */
+        num_entries += used_entries;
+
+        /* update number of entries in index array */
+        (*unifyfs_indices.ptr_num_entries) = num_entries;
+    }
+    return UNIFYFS_SUCCESS;
+}
+
 /* write count bytes from user buffer into specified chunk id at chunk offset,
  * count should fit within chunk starting from specified offset */
 static int unifyfs_logio_chunk_write(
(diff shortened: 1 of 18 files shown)


adammoody:
Thanks @tonyhutter .  I know this was discussed on a call that I missed, but are we allowed to add LGPL source code to a BSD project?

https://opensource.stackexchange.com/questions/6698/can-i-use-code-from-an-lgpl-v2-licenced-project-in-a-bsd-3-clause-licenced-proje 

If that's a problem, do we have AVL code under BSD?

client/src/unifyfs-sysio.c
```diff
@@ -2009,6 +2005,12 @@ int UNIFYFS_WRAP(fsync)(int fd)
 {
     /* check whether we should intercept this file descriptor */
     if (unifyfs_intercept_fd(&fd)) {
+
```
adammoody: style: kill this blank line after if()

client/src/unifyfs-sysio.c
```diff
@@ -998,12 +997,13 @@ ssize_t UNIFYFS_WRAP(read)(int fd, void* buf, size_t count)
...
 ssize_t UNIFYFS_WRAP(write)(int fd, const void* buf, size_t count)
 {
-    LOGDBG("write was called for fd %d", fd);
     size_t ret;
     off_t pos;
 
     /* check whether we should intercept this file descriptor */
     if (unifyfs_intercept_fd(&fd)) {
+
+        LOGDBG("write %d bytes to fd %d", count, fd);
```
adammoody: style: swap these two lines, to kill the blank line after if() and then separate LOGDBG statement from following block of code

client/src/unifyfs.c
```diff
@@ -914,6 +921,7 @@ int unifyfs_fid_create_file(const char* path)
 {
     /* check that pathname is within bounds */
     size_t pathlen = strlen(path) + 1;
+    int rc;
```
adammoody: style: let's either declare rc this at the very to of the function followed by a blank line, or move it down into the area where it is used like ```int rc = seg_tree_init(&meta->seg_tree);```

common/src/seg_tree.c
```diff
@@ -0,0 +1,311 @@
...
+        return -1;
+    } else {
+        return 0;
+    }
+}
+
+/* Returns 0 on success, positive non-zero error code otherwise */
+int seg_tree_init(struct seg_tree* seg_tree)
+{
+	memset(seg_tree, 0, sizeof(*seg_tree));
```
adammoody: style: extra leading space before memset?

client/src/unifyfs-fixed.c
```diff
@@ -543,78 +621,77 @@ static int unifyfs_logio_chunk_write(
...
+ * re-written.  The writes in the index will be flattened, non-overlapping,
+ * and sequential, for each file, one after another.  All seg_trees will be
+ * cleared.
+ *
+ * This function is typically called on an fsync().
+ */
+void unifyfs_rewrite_index_from_seg_tree(void)
+{
+    struct seg_tree_node* node;
+    int fid, gfid;
```
adammoody: style: let's move these three declarations down to the lines where the variables are used:
```
int fid = unifyfs_fds[i].fid;
```

client/src/unifyfs-fixed.c
```diff
@@ -471,6 +473,82 @@ static int unifyfs_split_index(
...
+                /* something went wrong when trying to flush key/values */
+                LOGERR("failed to flush key/value index to server");
+                return UNIFYFS_ERROR_IO;
+            }
+
+            /* flushed, clear buffer and refresh number of entries
+             * and number remaining */
+            num_entries = 0;
+            *(unifyfs_indices.ptr_num_entries) = num_entries;
+            remaining_entries = unifyfs_max_index_entries - num_entries;
```
adammoody: Let's we rewrite this code to call your new clear_index function now that we have it

client/src/unifyfs-fixed.c
```diff
@@ -543,78 +621,77 @@ static int unifyfs_logio_chunk_write(
...
+/*
+ * Clear all entries in the log index.  This only clears the metadata,
+ * not the data itself.
+ */
+static void
+unifyfs_clear_index(void)
+{
+    size_t index_size;
+    index_size = sizeof(*unifyfs_indices.index_entry) *
+        (*unifyfs_indices.ptr_num_entries);
```
adammoody: style: let's pull (*unifyfs_indices.ptr_num_entries) into a local variable.  This helps a lot when running under a debugger like totalview.
```
    off_t num_entries = *unifyfs_indices.ptr_num_entries;
    size_t index_size = sizeof(*unifyfs_indices.index_entry) * num_entries;
```

client/src/unifyfs-fixed.c
```diff
@@ -543,78 +621,77 @@ static int unifyfs_logio_chunk_write(
...
+        seg_tree_add(&meta->seg_tree, pos, pos + count - 1, log_offset);
     }
 
-    /* add new index entries if needed */
-    while (cur_idx.length > 0) {
-        /* remaining entries we can fit in the shared memory region */
-        off_t remaining_entries = unifyfs_max_index_entries - num_entries;
+    /* Update our write metadata with the new write */
+    return (unifyfs_logio_add_write_meta_to_index(gfid, pos, log_offset,
+        count));
```
adammoody: style: let's modify this line to call the function one line and return on another:
```
    int rc = unifyfs_logio_add_write_meta_to_index(gfid, pos, log_offset, count));
    return rc;
```
That makes it easier to see the return value from the function with a debugger.

common/src/seg_tree.c
```diff
@@ -0,0 +1,311 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+ /*
+  * This file is a simple, thread-safe, segment tree implementation.  The
+  * segments in the tree are no-overlapping.  Added segments overwrite the old
```
adammoody: typo? no-overlapping --> non-overlapping

tonyhutter:
@adammoody thanks for taking a look at this - let me go though your changes.  I think you bring up a good point with the LGPL.  Looks like I need to change the Makefile to build `rb.c` as a **dynamic** library rather than statically as I have now.  That will make it compliant.

tonyhutter:
@MichaelBrim I'll fix the typo, thanks for spotting that.

tonyhutter:
> Looks like I need to change the Makefile to build rb.c as a dynamic library rather than statically as
> I have now. That will make it compliant.

Maybe I spoke too soon.  We'd be distributing the modified `rb.c/rb.h` as well, which may be an issue.  Let me look into that.  In the meantime, if you know of a BSD/MIT red-black tree or AVL tree implementation, please let me know.

tonyhutter:
@adammoody my latest push swaps out the LGPL'd red black tree implementation in favor of OpenBSD's `tree.h` one. It also fixes all your comments.

I believe the travis errors are due to #415

scripts/checkpatch.sh
```diff
@@ -30,6 +30,29 @@ checkpatch_ignore+=",VOLATILE"            # Allow use of volatile keyword
 
 checkpatch_cmd+=" --ignore $checkpatch_ignore"
 
+# SPECIAL CASE: exclude tree.h
```
MichaelBrim: We have the same sorts of checkpatch failures for the macros in `common/src/unifyfs_configurator.h`. Why have this in the checkpatch.sh script? The whole point of TEST_CHECKPATCH_SKIP_FILES is so you can set it in the commit message. You can just do `TEST_CHECKPATCH_SKIP_FILES=common/src/tree.h` in the commit message.

client/src/unifyfs-sysio.c
```diff
@@ -1708,9 +1707,6 @@ static int process_read_data(read_req_t* read_reqs, int count, int* done)
         req.length  = msg->length;
         req.errcode = msg->errcode;
 
-        LOGDBG("read reply: gfid=%d offset=%zu size=%zu",
```
MichaelBrim: Why remove this LOGDBG()?

client/src/unifyfs-sysio.c
```diff
@@ -998,7 +997,7 @@ ssize_t UNIFYFS_WRAP(read)(int fd, void* buf, size_t count)
 /* TODO: find right place to msync spillover mapping */
 ssize_t UNIFYFS_WRAP(write)(int fd, const void* buf, size_t count)
 {
-    LOGDBG("write was called for fd %d", fd);
+    LOGDBG("write %d bytes to fd %d", count, fd);
```
MichaelBrim: should use `%zu` for size_t `count` parameter

scripts/checkpatch.sh
tonyhutter: I just tried that in the commit message, but it still flagged the errors.

MichaelBrim: Did you do a rebase too?

client/src/unifyfs-sysio.c
tonyhutter: I can't remember, I'll add it back in.

client/src/unifyfs-sysio.c
tonyhutter: Ok, will fix

scripts/checkpatch.sh
tonyhutter: I did a `git commit --amend` to re-write the commit message.  I also just did a rebase, but still no luck.

MichaelBrim: weird. I checked out your branch, removed the checkpatch change, and then did:
```
export TEST_CHECKPATCH_SKIP_FILES="common/src/tree.h"
git diff dev | ./scripts/checkpatch.sh
```
It worked for me.

tonyhutter: Yea, it works if you export it as a env var, but doesn't work if you add `TEST_CHECKPATCH_SKIP_FILES=common/src/tree.h` in the commit message.  That's why I had to add it to checkpatch.sh itself, to get it to pass Travis.

MichaelBrim: I also just tried to amend the commit message, and that worked for me too. Wonder what is different in your environment?

tonyhutter: Is `TEST_CHECKPATCH_SKIP_FILES` still exported?  

I don't see anywhere in the code where checkpatch.sh would be looking through the commit message to set env variables.

MichaelBrim: I recreated what travis does by the following commands:
```
> eval $(./scripts/git_log_test_env.sh)
> export TEST_CHECKPATCH_SKIP_FILES
> ./scripts/checkpatch.sh
> echo $TEST_CHECKPATCH_SKIP_FILES
common/src/tree.h
```

MichaelBrim: It's the first command the greps for the TEST_* variables in the commit message and adds them into the environment.

tonyhutter: Ah ok, I wasn't aware Travis doing some magic beforehand when it called checkpatch.  That seems wrong, but that's an issue for another time.  I'll update the commit message with `TEST_CHECKPATCH_SKIP_FILES` set and just export it when I want to checkpatch manually.

adammoody:
Here is where write index entries are combined as they are written into the log in the current code:
https://github.com/LLNL/UnifyFS/blob/3f0371ea455fb8d8eb48a1d0c7fcb48a206e24c4/client/src/unifyfs-fixed.c#L568

client/src/unifyfs-fixed.c
```diff
@@ -471,6 +472,96 @@ static int unifyfs_split_index(
...
+
+    /* add new index entries if needed */
+    while (cur_idx.length > 0) {
+        /* remaining entries we can fit in the shared memory region */
+        off_t remaining_entries = unifyfs_max_index_entries - num_entries;
+
+        /* if we have filled the key/value buffer, flush it to server */
+        if (0 == remaining_entries) {
+            /* index buffer is full, flush it */
+            int ret = invoke_client_fsync_rpc(cur_idx.gfid);
```
adammoody: There are a couple places like this where we can invoke the fsync rpc as the user is writing.  There is another one in unifyfs_split_index.  This is when the user has filled up the metadata index.  I suppose we should also rewrite and/or blank out the segment tree at these points.  Do you do that now?

tonyhutter: You bring up a really good issue.  I don't think things will work with my patch with that sync in there.  There may be a way I can reorder things so that the potential "sync on full index"  is done before the seg_tree_add.  Let me look into it and update the patch.

CamStan:
@tonyhutter, the issue with Travis was that `AX_PTHREAD` is apparently found in the autoconf archive, which Travis doesn't install by default. I added it to the .travis.yml and pushed it to your branch. It looks like it's working now.

tonyhutter:
@CamStan thanks!!!

tonyhutter:
@adammoody I pushed changes that address the `unifyfs_coalesce_index()`/`unifyfs_split_index()` issues you pointed out.  However, I broke something else along the way...  let me fix that first and then I'll have you take another look.

tonyhutter:
Fixed the issue.  @adammoody can you take another look at this?

adammoody:
With the break, I forgot we had this sitting out there.  I'll take a fresh look.

adammoody:
Ok, closer.  One concern is whether the rewrite function can overflow the limit of the index array.  I see it has a loop but it doesn't check that there is space.  That could be ok if we know that the combined set of index entries across all trees are not greater than the number of slots, but can we be sure.  If not, we need to fix that.  If so, we should probably add comments to describe that.

In particular, one problem I can see is that the number of entries in the tree for a file could be greater than the number in the index buffer, e.g., if someone were to write a large segment, and then later write a smaller segment that falls entirely within the larger segment.  I think we'd end up with two index entries in the buffer, and three in the tree.  Since we only sync when the buffer fills, the rewrite could then overflow.  Is that right?

client/src/unifyfs-fixed.c
```diff
@@ -471,6 +482,146 @@ static int unifyfs_split_index(
...
+int
+unifyfs_sync(int gfid)
+{
+    int ret;
+
+    if (unifyfs_flatten_writes) {
+        unifyfs_rewrite_index_from_seg_tree();
+    }
+
+    ret = invoke_client_fsync_rpc(gfid);
```
adammoody: Let's declare int ret here, and save a couple lines above.

client/src/unifyfs-fixed.c
```diff
@@ -543,78 +695,64 @@ static int unifyfs_logio_chunk_write(
...
     int gfid = unifyfs_gfid_from_fid(fid);
 
-    /* define an new index entry for this write operation */
-    unifyfs_index_t cur_idx;
-    cur_idx.gfid     = gfid;
-    cur_idx.file_pos = pos;
-    cur_idx.log_pos  = log_offset;
-    cur_idx.length   = count;
+    /* Update our write metadata with the new write */
+    rc = unifyfs_logio_add_write_meta_to_index(meta, pos, log_offset,
```
adammoody: let's move the type definition of int rc here and drop it from the top of the function, since I don't think we use it anywhere else above.

client/src/unifyfs-fixed.c
```diff
@@ -471,6 +482,146 @@ static int unifyfs_split_index(
...
+ * Clear all entries in the log index.  This only clears the metadata,
+ * not the data itself.
+ */
+static void
+unifyfs_clear_index(void)
+{
+    off_t num_entries = *unifyfs_indices.ptr_num_entries;
+    size_t index_size = sizeof(*unifyfs_indices.index_entry) * num_entries;
+
+    memset(unifyfs_indices.index_entry, 0, index_size);
```
adammoody: For performance, we could skip this memset.  Though it would be nice to have in a debug mode.  Can we conditionally do that?

tonyhutter:
> One concern is whether the rewrite function can overflow the limit of the index array. 

That's a good point.  We could simply force a sync if the seg_tree count is greater than or equal to `unifyfs_max_index_entries - 1` (since adding a single write could at most create two seg_tree entries).

client/src/unifyfs-fixed.c
tonyhutter: I'll just get rid of it.  There's really no need for the `memset()`

client/src/unifyfs-fixed.c
tonyhutter: Ok

client/src/unifyfs-fixed.c
tonyhutter: Ok

adammoody:
> > One concern is whether the rewrite function can overflow the limit of the index array.
> 
> That's a good point. We could simply force a sync if the seg_tree count is greater than or equal to `unifyfs_max_index_entries - 1` (since adding a single write could at most create two seg_tree entries).

Yeah, that makes sense.  Since we're flushing all segment trees in one shot, we'd want to use a global count of entries across all active trees.

I think a single write can replace an existing segment tree entry with three, right?  Something like:

existing entry: [50, 100)
new write: [60, 70)
resulting set: [50, 60), [60, 70), [70, 100)

Is that what you're thinking?

tonyhutter:
> I think a single write can replace an existing segment tree entry with three, right?

You are correct sir!  I should add in a proper `seg_tree_count()` function too.  So after a write, we'd be doing:
`if seg_tree_count(tree) >= (unifyfs_max_index_entries - 2) then 
    sync()
`.

adammoody:
@tonyhutter , the collective work is coming longer faster than I anticipated, and I'm at a point where I could start to you your new range tree in the server.  I see there is a merge conflict at the moment, preventing a merge.  Even if you're not quite done with the final details on this PR, could you fix up the merge conflict?  We can go ahead and merge this and then work in parallel if you're still polishing things up.

tonyhutter:
@adammoody will do.  Note that my last push updated the seg_tree with two functions you might need:
```
/* Return the number of segments in the segment tree */
unsigned long seg_tree_count(struct seg_tree* seg_tree)

/* Return the maximum segment value in the tree */
unsigned long seg_tree_max(struct seg_tree* seg_tree)
```

adammoody:
> @adammoody will do. Note that my last push updated the seg_tree with two functions you might need:
> 
> ```
> /* Return the number of segments in the segment tree */
> unsigned long seg_tree_count(struct seg_tree* seg_tree)
> 
> /* Return the maximum segment value in the tree */
> unsigned long seg_tree_max(struct seg_tree* seg_tree)
> ```

Cool, we could easily compute the total number of segments for a file with that count function and an allreduce.

tonyhutter:
@adammoody you can pull this in - I rebased and fixed the conflict.  I also added code to sync seg_tree if it's going to overflow the index but haven't actually tested it yet.

adammoody:
Awesome!  Thanks, @tonyhutter !

PR #416, opened 2019-12-11
Fix -wrap configure bug

CamStan:
### Description

When building on some systems, the configure check `checking if linker supports -wrap...` when checking for `HAVE_LD_WRAP`. This would fail with an undefined reference to `__wrap_malloc`. This would result in the static examples not being built and installed.

This adds a definition for __wrap_malloc for this test to pass.

### How Has This Been Tested?

Building without this does not build/install the static examples. Building with this change does.

### Types of changes

### Checklist:

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -176,11 +176,11 @@ AC_TRY_COMPILE(
     AC_MSG_RESULT(no)
 )
 
-AC_MSG_CHECKING(if linker suppots -wrap)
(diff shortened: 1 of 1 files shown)


CamStan:
@adammoody, I don't think I fully understand this. It seems weird that this check for `HAVE_LD_WRAP` would succeed on Travis CI (and possibly at ORNL from the sounds of it) but fail for us. Would not what I added just make it so this check passes no matter what?

PR #417, opened 2019-12-16
transfer API into unifyFS chmod()s to laminate

craigsteffen:
When transferring files INTO the unify FS, this chmods the file to 
0444 (read-only) so that the 
file gets laminated so it can be read by other processes.  

### Motivation and Context

When files were copied in, they weren't laminated, so nothing else could 
read the file.

### How Has This Been Tested?

With this change, I was able to set up a running unifyFS, copy a file in, and then 
back out with the transfer API, and then file was intact.  Without hte change, the file
came back out with zero bytes. 

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -2786,12 +2786,15 @@ int unifyfs_transfer_file(const char* src, const char* dst, int parallel)
     int ret = 0;
     int dir = 0;
     struct stat sb_src = { 0, };
+    mode_t source_file_mode_write_removed;
     struct stat sb_dst = { 0, };
     int unify_src = 0;
     int unify_dst = 0;
     char dst_path[PATH_MAX] = { 0, };
     char* pos = dst_path;
     char* src_path = strdup(src);
+
+    int local_return_val;
 
     if (!src_path) {
         return -ENOMEM;
@@ -2828,9 +2831,27 @@ int unifyfs_transfer_file(const char* src, const char* dst, int parallel)
     }
 
     if (parallel) {
-        return do_transfer_file_parallel(src_path, dst_path, &sb_src, dir);
+        local_return_val =
+	    do_transfer_file_parallel(src_path, dst_path, &sb_src, dir);
     } else {
-        return do_transfer_file_serial(src_path, dst_path, &sb_src, dir);
+        local_return_val =
+	    do_transfer_file_serial(src_path, dst_path, &sb_src, dir);
     }
+
+    // We know here that one (but not both) of the constituent files
+    // is in the unify FS.  We just have to decide if the *destination* file is.
+    // If it is, then now that we've transferred it, we'll set it to be readable
+    // so that it will be laminated and will be readable by other processes.
+    if (unify_dst) {
+      // pull the source file's mode bits, remove all the write bits but leave
+      // the rest intact and store that new mode.  Now that the file has been
+      // copied into the unify file system, chmod the file to the new
+      // permission.  When unify senses all the write bits are removed it will
+      // laminate the file.
+        source_file_mode_write_removed =
+	    (sb_src.st_mode) & ~(0222);
+        chmod(dst_path, source_file_mode_write_removed);
+    }
+    return local_return_val;
 }
 


craigsteffen:
Ok, someone said that they wanted to do this procedure of rejecting the PR
and then re-submitting it, so here's the chance.

I suspect made the static return variable that because the function was
that?  I can just remove the static from the variable definition.  I'll
build it and test it again.

And I'll fix the comment.

Then what?  I think I withdraw the PR in the web series.  Then I could fix
in my source, commit the changes, and then rebase to combine the commits
into one again and then push that?  Is that right?

Thanks for looking at this so quickly,

Craig

MichaelBrim:
@craigsteffen The procedure to update the PR is:
1. make changes and commit them to your branch
2. do an interactive rebase against dev to combine all commits into one
3. git push -f origin <your-branch-name>

There's no need to withdraw the PR or do anything via the web interface. Github will automatically pick up the changes to your branch.

craigsteffen:
I made the changes, rebased, and then re-commited to what I think is the right branch, so I think this is ready for re-review.

client/src/unifyfs.c
```diff
@@ -2828,9 +2830,20 @@ int unifyfs_transfer_file(const char* src, const char* dst, int parallel)
...
+        local_return_val =
+	    do_transfer_file_serial(src_path, dst_path, &sb_src, dir);
+    }
+
+    // We know here that one (but not both) of the constituent files
+    // is in the unify FS.  We just have to decide if the *destination* file is.
+    // If it is, then now that we've transferred it, we'll set it to be readable
+    // so that it will be laminated and will be readable by other processes.
+    if (unify_dst) {
+        chmod(dst_path, 0444);
```
MichaelBrim: @craigsteffen It's not clear to me if this will go through our chmod wrapper or not. When you tested this, did you see the `LOGDBG("chmod: setting ...")` message from `int UNIFYFS_WRAP(chmod)(const char* path, mode_t mode)`?

tonyhutter: Also, instead of setting 0444, we should remove the write bits from the existing permissions.

craigsteffen:
When I first tested it, when I added the chmod() call to the code, when I
copied in the file, I saw messages that mentioned that the file permissions
were changing and thus the file needed to be laminated.

I do know that before I made this change, files that I copied into unifyFS
using the transfer API would come out at 0 bytes.  After I made the change,
the (admittedly very few) files that I copied in and then back out came out
identical to the original file.

craigsteffen:
Ok, somehow I got the impression somewhere that a static octal permission
was the way to call the chmod() function.

If I can call it with "a-w" (which as you say, is what I actually want to
do) I'll do that.  I'll try that and if that works, then I'll change it and
re-submit it.

Thanks for pointing this out.

Craig

MichaelBrim:
@craigsteffen you are correct that octal is the correct method. Tony wasn't suggesting to use the command-line forms, but instead to retrieve the current octal permissions and mask off the write permissions.

craigsteffen:
Ah, gotcha.  When I'm at a terminal I'll look up the function that does
that.  I assume stat, or something like that that will give me info
equivalent to "ls -l" on a path.

craigsteffen:
Searching on the web it looks like to first order, stat() is the way to go
about it.  I just need to remember that the mode field has type int but it
only scans visually as permissions when printed in octal.

MichaelBrim:
Once you have the current mode, the following code with mask off the write perms.
```
int new_mode = old_mode & ~(S_IWUSR | S_IWGRP | S_IWOTH);
```
Explanation is left as an exercise for the reader :)

craigsteffen:
Yep.  Exactly.  As I was waiting for my son at school I was trying to work
out the bitwise math.  As long as "~" is bitwise-NOT, then that's what I
was looking for.  Thanks!

craigsteffen:
Oops--I built this new version (no errors or warnings) but I forgot to test it again.  I'll post here again when I've specifically tested it and made sure it still works as expected WRT the file.

craigsteffen:
This is tested.  It works as well as the first version (just statically setting the permissions to 0444) worked.  

I'm not sure that file permission bits are preserved or transferred out in any case.  In my testing, I set a group-read bit on the file before copying it in.  That bit wasn't set when the file was copied back out.  So I'm not sure that it matters if this chmod does anything other than set the permissions to 0444 anyway, but if in the future unify starts tracking them, then this is the right thing to do. 

This can go into Unify as far as I know.

adammoody:
Thanks, @craigsteffen !  Congrats fighting through that PR process :-)

PR #418, opened 2019-12-18
Simple addition of more debug info

clmendes:
Add more debug information to the client log

### Description
Include additional arguments to some of the LOGDBG commands in unifyfs-sysio.c, such that
during execution more fields are displayed in the client log. In addition, rectify some of the
comments in the source code such that those comments more precisely define what is done
by the specific fragment of the code. No changes to the logic of the original code are made.

### Motivation and Context
When debugging the execution of the client, it is useful to view as many variables as possible,
in particular those key variables that are passed to main Unify functions. In the original code,
only part of such variables were being displayed.

### How Has This Been Tested?
Testing has been done by running example 'app-hdf5-create-gotcha' (to create a simple HDF5 file under Unify) followed by 'sysio-cp-gotcha', which exports that created file to permanent storage.
For both examples, the logs generated contain the additional information that was inserted by
this change. which helps in verifying the proper behavior of each example.

### Types of changes
- Addition of arguments to LOGDBG () commands
- Change to comments in the code to more precisely define what the code fragment does

### Checklist:

- [x ] My code follows the UnifyFS code style requirements.

### DIFF
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -245,7 +245,8 @@ int UNIFYFS_WRAP(truncate)(const char* path, off_t length)
         /* truncate the file */
         int rc = unifyfs_fid_truncate(fid, length);
         if (rc != UNIFYFS_SUCCESS) {
-            LOGDBG("unifyfs_fid_truncate failed for %s in UNIFYFS", path);
+            LOGDBG("unifyfs_fid_truncate failed for %s, fid %d, length %zu", 
+                path, fid, length);
             errno = EIO;
             return -1;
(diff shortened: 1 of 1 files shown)


MichaelBrim:
@clmendes you need to remove the trailing spaces in your changes to get checkpatch to pass.

MichaelBrim:
@clmendes do you still want these debug additions, or can we close this?

clmendes:
Michael, I'd have more changes to add, but I'm not ready to commit them right now, so I think this PR can be closed, and I'll start a new one later on. Thanks!

PR #419, opened 2020-01-02
update size field in global meta on truncate

adammoody:
This adds support for truncate and ftruncate for extending and shrinking a file.  A new truncate rpc is defined on the server.  The implementation internally updates the file size field on the global meta data on a truncate call.  The filesize rpc has been updated to take the maximum of this file size field and the maximum offset of any write index as the true file size.

To shrink a file, there are two additional steps done in the truncate rpc in that we delete any index entries that extend beyond the new file size, and we rewrite any entries that happen to overlap (start before and end after) the new file size.

Note that unlinking a file current truncates the file to size=0, which will now delete any write index entries for the file.

If the file is not laminated, the stat wrappers have been updated to query the filesize rpc to get the current size whereas they used to always return 0.

### Types of changes

### Checklist:

### DIFF
--- a/t/sys/truncate.c
+++ b/t/sys/truncate.c
@@ -1,0 +1,667 @@
+/*
+ * Copyright (c) 2019, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2018, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyFS.
+ * For details, see https://github.com/LLNL/UnifyFS.
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
+
+ /*
+  * Test truncate and ftruncate
+  */
+#include <fcntl.h>
+#include <string.h>
+#include <errno.h>
+#include <linux/limits.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include "t/lib/tap.h"
+#include "t/lib/testutil.h"
+
+/* Get global, local, or log sizes (or all) */
+static
+void get_size(char* path, size_t* global, size_t* local, size_t* log)
+{
+    struct stat sb = {0};
+    int rc;
+
+    rc = stat(path, &sb);
+    if (rc != 0) {
+        printf("Error: %s\n", strerror(errno));
+        exit(1);    /* die on failure */
+    }
+    if (global) {
+        *global = sb.st_size;
+    }
+
+    if (local) {
+        *local = sb.st_rdev & 0xFFFFFFFF;
+    }
+
+    if (log) {
+        *log = (sb.st_rdev >> 32) & 0xFFFFFFFF;
+    }
+}
+
+int truncate_test(char* unifyfs_root)
+{
+    char path[64];
+    int rc;
+    int fd;
+    size_t global, local, log;
+
+    size_t bufsize = 1024*1024;
+    char* buf = (char*) malloc(bufsize);
+
+    testutil_rand_path(path, sizeof(path), unifyfs_root);
+
+    /* Open a new file for writing */
+    fd = open(path, O_WRONLY | O_CREAT | O_TRUNC, 0222);
+    ok(fd != -1, "%s:%d open(%s) (fd=%d): %s",
+        __FILE__, __LINE__, path, fd, strerror(errno));
+
+    /* file should be 0 bytes at this point */
+    get_size(path, &global, &local, &log);
+    ok(global == 0, "%s:%d global size is %d expected %d",
+        __FILE__, __LINE__, global, 0);
+    ok(local == 0, "%s:%d local size is %d expected %d",
+        __FILE__, __LINE__, local, 0);
+    ok(log == 0, "%s:%d log size is %d expected %d",
+        __FILE__, __LINE__, log, 0);
+
+    /* write 1MB and fsync, expect 1MB */
+    rc = write(fd, buf, bufsize);
+    ok(rc == bufsize, "%s:%d write(%d) (rc=%d): %s",
+        __FILE__, __LINE__, bufsize, rc, strerror(errno));
+
+    rc = fsync(fd);
+    ok(rc == 0, "%s:%d fsync() (rc=%d): %s",
+        __FILE__, __LINE__, rc, strerror(errno));
+
+    get_size(path, &global, &local, &log);
+    ok(global == 1*bufsize, "%s:%d global size is %d expected %d",
+        __FILE__, __LINE__, global, 1*bufsize);
+    ok(local == 1*bufsize, "%s:%d local size is %d expected %d",
+        __FILE__, __LINE__, local, 1*bufsize);
+    ok(log == 1*bufsize, "%s:%d log size is %d expected %d",
+        __FILE__, __LINE__, log, 1*bufsize);
+
+    /* skip a 1MB hole, write another 1MB, and fsync expect 3MB */
+    rc = lseek(fd, 2*bufsize, SEEK_SET);
+    ok(rc == 2*bufsize, "%s:%d lseek(%d) (rc=%d): %s",
+        __FILE__, __LINE__, 2*bufsize, rc, strerror(errno));
+
+    rc = write(fd, buf, bufsize);
+    ok(rc == bufsize, "%s:%d write(%d) (rc=%d): %s",
+        __FILE__, __LINE__, bufsize, rc, strerror(errno));
+
+    rc = fsync(fd);
+    ok(rc == 0, "%s:%d fsync() (rc=%d): %s",
+        __FILE__, __LINE__, rc, strerror(errno));
+
+    get_size(path, &global, &local, &log);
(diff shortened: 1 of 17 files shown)


tonyhutter:
Can you include the text from your PR description in your commit message?

adammoody:
Yeah, I'll do that.  I'll likely add more to this, so I'll come back to that on the final rebase of this PR.

t/sys/write-read.c
```diff
@@ -88,7 +88,7 @@ int write_read_test(char* unifyfs_root)
 
     /* Check global and local size on our un-laminated file */
     get_size(path, &global, &local, &log);
-    ok(global == 0, "%s: global size is %d: %s",  __FILE__, global,
+    ok(global == 15, "%s: global size is %d: %s",  __FILE__, global,
```
tonyhutter: I thought we decided that un-laminated files always reported size 0?

tonyhutter:
Can you add a test case for truncate()?  Possibly:
1. truncate() to create an new empty file (should be all zeros).  Test reading back from a non-zero offset as well.
2. truncate() to expand an existing file and read it back.
3. truncate() to shrink an existing file and read it back.
4. truncate() a file to 0 bytes.
5. truncate() to create a new empty file that's huge (petabytes?)

t/sys/write-read.c
adammoody: I still think that's a good optimization for codes that can use it.  Since computing file size before lamination is expensive, it'd probably be nice to have an option to skip it and report 0.  However, I seem to recall that HDF requires an accurate file size at times *before* lamination.

adammoody:
Added the PR description above to the commit message.

adammoody:
I do have a test case, but it needs to be converted into our test suite format.  I likely will need some consulting from @CamStan on that, so I'll submit that in a later PR.

adammoody:
@tonyhutter, @CamStan , I created a PR for the truncate test case here: https://github.com/LLNL/UnifyFS/pull/422

This covers a subset of the cases so far.  We can add more cases.

I have no idea if this runs yet, and I'll guess that the global/local/log size checks might not be correct.

adammoody:
The tests that I added here check file size as reported by stat calls.  The first test executes the following sequence:
- It checks the initial size of a file is 0.
- It writes 1MB and checks that the global size is 1MB after an fsync.
- It then skips a 1MB hole with lseek and writes another 1MB, fsyncs, and checks that the file size is 3MB.
- It extends the file with a trailing hole by truncating to 5MB.
- It truncates the file from 5MB down to 0.5MB.
- It truncates the file from 0.5MB to 0.

A second test creates a new file, truncates to 1TB and then back to 0.

tonyhutter:
If I do:
1. write 1MB of 0xFFs to a new file
2. truncate() the entire file
3. fsync()

Will the final state of the file be 1MB of 0xFFs?  If so, should truncate() first call fsync() before doing the truncate RPC?

adammoody:
> If I do:
> 
> 1. write 1MB of 0xFFs to a new file
> 2. truncate() the entire file
> 3. fsync()
> 
> Will the final state of the file be 1MB of 0xFFs? If so, should truncate() first call fsync() before doing the truncate RPC?

Yep, and yep

adammoody:
I need to double check on how we're handling file size on reads now, since it may be inconsistent since stat is returning the current global file size but reads and other functions may still be using the fid_logical_size function, which is returning local size until lamination.

adammoody:
To tabulate the various "file size" values we're using in different places
```
fid_local_size() -- returns meta->local_size
- write() on fd opened in APPEND mode

fid_logical_size() -- returns meta->global_size if laminated, else meta->local_size
- lseek(SEEK_END)
- fseek(SEEK_END)
- stream_write on stream opened with APPEND

get_meta_with_size() -- invokes filesize_rpc if not laminated
- __stat

invoke_client_filesize_rpc()
- get_meta_with_size if file not laminated
- fd_logreadlist (if read < length)
- __chmod on lamination
```

adammoody:
I've gone through a lot of our code that modifies size.  The short story is that we have a few inconsistencies that we should clean up, but those aren't made any worse this PR.  I'll create future PRs to clean those up.  In the meantime, I recommend we merge this one soon so that we have a working truncate/ftruncate to make progress on HDF.

Longer story:  there are a few spots in the client code where we set the file size field in the global meta data where we don't really mean to, and that will now interfere with how truncate works.  For example, operations like chmod/chgrp set the full global metadata, which sets size=0 as a side effect if the file is not laminated.  This will overwrite any value a previous truncate may have set it to so that, for example, the following code will blow away the size value:
```
fd = open(path, O_WRONLY | O_CREAT | O_TRUNC); // size = 0
truncate(path, 1024); // now size = 1024
chmod(path, ...) // forces size=0 again, blowing away correct size of 1024
```
One solution to this is to only set the global size field to 0 when the file is first created, and then to ignore the size field on later metadata update rpcs.  To do that, we need to know the difference between a create and an update, which we can do by either adding an flag to the existing set metadata rpc or by defining a new create rpc.  That change should go in a new PR.

adammoody:
Added PR to correct metadata updates from overwriting size field: https://github.com/LLNL/UnifyFS/pull/429

adammoody:
> Can you add a test case for truncate()? Possibly:
> 
> 1. truncate() to create an new empty file (should be all zeros).  Test reading back from a non-zero offset as well.
@tonyhutter , can you clarify what you're suggesting here?

> 2. truncate() to expand an existing file and read it back.
Yes.  write_read_hole_test() in t/sys/write-read-hole.c expands a file with truncate and reads from that trailing hole.

> 3. truncate() to shrink an existing file and read it back.
Yes. truncate_eof() in t/sys/truncate.c truncates a larger file down to 0.5MB and reads it back.

> 4. truncate() a file to 0 bytes.
Yes. truncate_eof in t/sys/truncate.c truncates a file down to 0 bytes.

> 5. truncate() to create a new empty file that's huge (petabytes?)
Yes.  See truncate_bigempty() in t/sys/truncate.c

tonyhutter:
>> truncate() to create an new empty file (should be all zeros). Test reading back from a non-zero offset as well.

> @tonyhutter , can you clarify what you're suggesting here?

1. Truncate a file and read it all back, starting from offset 0.
2. Read the same file starting from a non-zero offset (like byte offset 42).  This is just to test a seeked, non-sequential, read of the truncated range.

adammoody:
> > > truncate() to create an new empty file (should be all zeros). Test reading back from a non-zero offset as well.
> 
> > @tonyhutter , can you clarify what you're suggesting here?
> 
> 1. Truncate a file and read it all back, starting from offset 0.
> 2. Read the same file starting from a non-zero offset (like byte offset 42).  This is just to test a seeked, non-sequential, read of the truncated range.

@tonyhutter , I added 4 more tests.  One case writes a file with a known pattern, truncates it to a smaller size, then reads it back checking the total number of bytes read and the content along the way.  That case is run twice: one where it reads the entire file starting from offset=0 and a second test that reads the file starting from offset=2020.  There are two more tests that do a similar thing but with a file that was created empty and then truncated up to some size so that all "data" really comes from a hole and is 0.

t/sys/truncate.c
```diff
@@ -0,0 +1,719 @@
...
+    get_size(path, &global, &local, &log);
+    ok(global == 1*bufsize, "%s:%d global size is %d expected %d",
+        __FILE__, __LINE__, global, 1*bufsize);
+    ok(local == 1*bufsize, "%s:%d local size is %d expected %d",
+        __FILE__, __LINE__, local, 1*bufsize);
+    ok(log == 1*bufsize, "%s:%d log size is %d expected %d",
+        __FILE__, __LINE__, log, 1*bufsize);
+
+    /* skip a 1MB hole, write another 1MB, and fsync expect 3MB */
+    errno = 0;
```
tonyhutter: I'd get rid of all these `errno = 0`.  I understand you're trying to isolate the individual tests, but it's going to be a pain to keep that standard when writing future tests.  Most of the time you're going to just look at the first test that failed and start fixing it from there.

t/sys/truncate.c
```diff
@@ -0,0 +1,719 @@
...
+    rc = ftruncate(fd, bigempty);
+    ok(rc == 0, "%s:%d ftruncate(%llu) (rc=%d): %s",
+        __FILE__, __LINE__, (unsigned long long) bigempty,
+        rc, strerror(errno));
+
+    get_size(path, &global, &local, &log);
+    ok(global == (size_t)bigempty, "%s:%d global size is %llu expected %llu",
+        __FILE__, __LINE__, global, (unsigned long long)bigempty,
+        strerror(errno));
+    //ok(local == (size_t)bigempty, "%s:%d local size is %llu",
```
tonyhutter: Remove commented out lines

t/sys/truncate.c
```diff
@@ -0,0 +1,719 @@
...
+    }
+    return 0;
+}
+
+/* fill buffer with known pattern based on file offset */
+int check_pattern(char* buf, size_t size, size_t start)
+{
+    size_t i;
+    for (i = 0; i < size; i++) {
+        char expected = (char)((i + start) % 26) + 'A';
```
tonyhutter: `(char)` cast is not needed

adammoody:
Thanks for all of the suggestions on this, @tonyhutter !

PR #420, opened 2020-01-03
server: fix offset when reading from spillover file

adammoody:
Debugging https://github.com/LLNL/UnifyFS/issues/411, it looks like this is coming from an incorrect offset into the spillover file when the service manager reads data.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_service_manager.c
+++ b/server/src/unifyfs_service_manager.c
@@ -36,7 +36,6 @@
 #include "unifyfs_server_rpcs.h"
 #include "margo_server.h"
 
-
 /* Service Manager (SM) state */
 typedef struct {
     /* the SM thread */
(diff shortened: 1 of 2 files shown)


server/src/unifyfs_service_manager.c
```diff
@@ -102,141 +104,201 @@ int sm_issue_chunk_reads(int src_rank,
...
+
+            /* offset within spill over file, need to subtract off
+             * range of offsets that land in data region of
+             * superblock */
+            off_t spill_offset = (off_t)(offset - data_size +
+                sz_from_mem);
+
+            /* read data from the spillover file */
+            ssize_t nread = pread(spill_fd, (buf_ptr + sz_from_mem),
+                                  sz_from_spill, spill_offset);
```
adammoody: The fix is in the pread change, where is was using 0 as the last argument before (old line 201), it's now using an offset into the file based on the log_offset.  I had that singular change as its own commit and then layered a bunch of style edits and new comments on top of that in a later commit, but ended up merging these all into one to get past a checkpatch complaint.

PR #421, opened 2020-01-04
Nread

adammoody:
This adds an nread field to the client-side read request structure.  We use this to count the max offset within each read request that has valid data.  It is used to handle short reads and to deal with holes.

This also updates the server to handle read requests that generate no write index entries.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-internal.h
+++ b/client/src/unifyfs-internal.h
@@ -297,12 +297,23 @@ typedef struct {
 } unifyfs_filename_t;
 
 /*unifyfs structures*/
+
+/* This structure defines a client read request for a file.
+ * It is initialized by the client describing the global file id,
+ * offset, and length to be read and provides a pointer to
+ * the user buffer where the data should be placed.  The
+ * server sets the errcode field to UNIFYFS_SUCCESS if the read
+ * succeeds and otherwise records an error code pertaining to
+ * why the read failed.  The server records the number of bytes
+ * read in the nread field, which the client can use to detect
+ * short read operations. */
 typedef struct {
-    int gfid;
-    int errcode;
-    size_t offset;
-    size_t length;
-    char* buf;
+    int gfid;      /* global file id to be read */
+    int errcode;   /* error code for read operation if any */
+    size_t offset; /* logical offset in file to read from */
+    size_t length; /* number of bytes to read */
+    size_t nread;  /* number of bytes actually read */
+    char* buf;     /* pointer to user buffer to place data */
 } read_req_t;
 
 typedef struct {
--- a/server/src/unifyfs_request_manager.c
+++ b/server/src/unifyfs_request_manager.c
@@ -352,10 +352,10 @@ int create_chunk_requests(reqmgr_thrd_t* thrd_ctrl,
                           int num_vals,
                           unifyfs_keyval_t* keyvals)
 {
-    int thrd_id = thrd_ctrl->thrd_ndx;
+    /* get app id for this request batch */
     int app_id = thrd_ctrl->app_id;
-    int client_id = thrd_ctrl->client_id;
 
+    /* allocate read request structures */
     chunk_read_req_t* all_chunk_reads = (chunk_read_req_t*)
         calloc((size_t)num_vals, sizeof(chunk_read_req_t));
     if (NULL == all_chunk_reads) {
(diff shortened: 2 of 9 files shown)


tonyhutter:
Can you add a short commit message for each of these three commits?  Also, did you mean to push this branch into the main UnifyFS repo?  I was expecting to see it in a `adammoody:nread` branch.

server/src/unifyfs_request_manager.c
```diff
@@ -402,46 +411,115 @@ int create_chunk_requests(reqmgr_thrd_t* thrd_ctrl,
...
-            del_reads->reqs = all_chunk_reads + i;
-            del_reads->resp = NULL;
+            reads->rank     = curr_del;
+            reads->rdreq_id = rdreq->req_ndx;
+            reads->reqs     = all_chunk_reads + i;
+            reads->resp     = NULL;
         }
-        del_reads->num_chunks++;
+
+        /* increement number of read requests we're sending
```
MichaelBrim: minor typo in comment: "increement"

server/src/unifyfs_request_manager.c
```diff
@@ -1042,57 +1123,37 @@ static int rm_process_remote_chunk_responses(reqmgr_thrd_t* thrd_ctrl)
                     ret = rc;
                 }
             }
-        }
-    }
+        } else if ((req->num_remote_reads == 0) &&
```
MichaelBrim: I'm confused about this case (i.e., `num_remote_reads == 0`). Can you explain the scenario when this occurs? That field is a count of servers who need to respond (including the local server if applicable). Any request should have at least one server, right?

server/src/unifyfs_request_manager.c
adammoody: I like increment better

server/src/unifyfs_request_manager.c
adammoody: This happens in the case where a client issues a read for a segment of a file that does not exist, e.g., the read segment falls fully within a hole.

I was at first trying to modify things to have the server notify the client right away by adding a new field to the rpc, but that ran into snags.  Also I realized that this required the server to determine before it returns from the read/mread RPC whether it will find any data.  While we do that now, I can imagine we may move that out of the critical path later.  So it felt cleaner to have the client/server go through the normal flow where the client waits for read replies to come back from the server through shared memory.  In this case the server just informs the client that nothing is coming (repies=0 and flag=COMPLETE).

MichaelBrim: ah, ok. makes sense now

t/sys/write-read-hole.c
```diff
@@ -0,0 +1,182 @@
...
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
+
+ /*
+  * Test reading from file with holes
+  */
+#include <fcntl.h>
+#include <string.h>
+#include <errno.h>
+#include <linux/limits.h>
```
tonyhutter: This header isn't needed

client/src/unifyfs-internal.h
```diff
@@ -302,6 +302,7 @@ typedef struct {
     int errcode;
     size_t offset;
     size_t length;
+    size_t nread;
```
tonyhutter: Please add a comment here describing what `nread` is

adammoody:
> Can you replace this with `strchr()`?

I'm not sure what you mean here.

tonyhutter:
>>  Can you replace this with strchr()?

> I'm not sure what you mean here

Sorry, I misread what you were checking for.  Ignore me :-)

adammoody:
> > > Can you replace this with strchr()?
> 
> > I'm not sure what you mean here
> 
> Sorry, I misread what you were checking for. Ignore me :-)

Added some comments to the test case.

t/sys/write-read-hole.c
```diff
@@ -0,0 +1,209 @@
...
+#include <string.h>
+#include <errno.h>
+#include <stdio.h>
+#include <sys/stat.h>
+#include "t/lib/tap.h"
+#include "t/lib/testutil.h"
+
+/* Get global, local, or log sizes (or all) */
+static
+void get_size(char* path, size_t* global, size_t* local, size_t* log)
```
CamStan: Looks like we're getting multiple tests (and will probably get a lot more) that will need this function. We could move this function to [t/lib/testutil.h](https://github.com/LLNL/UnifyFS/blob/dev/t/lib/testutil.h), or could split it into three sharness tests (file_global_size, file_local_size, and file_log_size) and add them to our custom sharness tests in [t/sharness.d/02-functions.sh](https://github.com/LLNL/UnifyFS/blob/dev/t/sharness.d/02-functions.sh).

If either of these sounds good, I can create an issue and change this in a future PR.

adammoody: That makes sense to me.

clmendes: Adam, does this new "Nread" field follow the same logic of the 'nbytes' field that you had suggested in the past? I view it as a kind of water-mark tracking where valid data exists, is this right?

adammoody: Yep, I could have named this nbytes instead -- maybe should have.  I'll get this merged now so you can work from it for HDF testing.  We still need to get the truncate PR in there.

PR #425, opened 2020-01-08
The first implementation of unifyfs-shell utility.

sandrain:
Currently, this is not going to be too useful because ls command is not available. We can add more useful features as the client API becomes available.

### Description
An implementation of the unifyfs-shell utility program, which allows interactive shell-like interface to explore the unifyfs file system.

### Types of changes

### DIFF
--- a/util/unifyfs-shell/src/Makefile.am
+++ b/util/unifyfs-shell/src/Makefile.am
@@ -1,0 +1,42 @@
+bin_PROGRAMS = unifyfs-shell-gotcha unifyfs-shell-static
+
+unifyfs_shell_gotcha_SOURCES = unifyfs-shell.c \
+                               unifyfs-shell-util.c \
+                               unifyfs-shell-main.c \
+                               unifyfs-shell-status.c \
+                               unifyfs-shell-ls.c \
+                               unifyfs-shell-mkdir.c \
+                               unifyfs-shell-cd.c \
+                               unifyfs-shell-pwd.c \
+                               unifyfs-shell-stat.c
+
+unifyfs_shell_static_SOURCES = unifyfs-shell.c \
+                               unifyfs-shell-util.c \
+                               unifyfs-shell-main.c \
+                               unifyfs-shell-status.c \
+                               unifyfs-shell-ls.c \
+                               unifyfs-shell-mkdir.c \
+                               unifyfs-shell-cd.c \
+                               unifyfs-shell-pwd.c \
+                               unifyfs-shell-stat.c
+
+noinst_HEADERS = unifyfs-shell.h
+
+AM_CPPFLAGS = $(MPI_CFLAGS) \
+              -I$(top_srcdir)/common/src \
+              -I$(top_srcdir)/client/src
+
+AM_CFLAGS = -Wall
+
+unifyfs_shell_gotcha_LDADD = $(top_builddir)/common/src/libunifyfs_common.la \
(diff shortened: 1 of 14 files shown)


util/unifyfs-shell/src/unifyfs-shell.c
```diff
@@ -0,0 +1,295 @@
...
+}
+
+static inline void unifyfs_shell_help(void)
+{
+    int i = 0;
+    int n_commands = sizeof(commands)/sizeof(unifyfs_shell_cmd_t*);
+
+    printf(
+      "Builtin commands:\n"
+      "  .help           this help message\n"
```
MichaelBrim: any particular reason for prefixing the builtin commands with '.'?

util/unifyfs-shell/src/unifyfs-shell.c
```diff
@@ -0,0 +1,295 @@
...
+
+            MPI_Send(buf, size, MPI_CHAR, 0, 0, MPI_COMM_WORLD);
+
+            fclose(out);
+        }
+
+        MPI_Barrier(MPI_COMM_WORLD);
+    }
+
+out:
```
MichaelBrim: another case where you have an 'out' label and a function variable with the same name.

util/unifyfs-shell/src/Makefile.am
```diff
@@ -0,0 +1,26 @@
...
+
+AM_CPPFLAGS = $(MPI_CFLAGS) \
+              -I$(top_srcdir)/common/src \
+              -I$(top_srcdir)/client/src
+
+AM_CFLAGS = -Wall
+
+unifyfs_shell_LDADD = $(MPI_CLDFLAGS) \
+                      $(top_builddir)/common/src/libunifyfs_common.la \
+                      $(top_builddir)/client/src/libunifyfs_gotcha.la -lrt -lm
```
MichaelBrim: what if the configuration didn't include gotcha?

sandrain:
I will fix the issues and push again. Thanks.

util/unifyfs-shell/src/unifyfs-shell.c
sandrain: @MichaelBrim This was just to differentiate these meta commands from main commands that resemble binutil commands. I will remove the dot prefix if this is not agreeable.

MichaelBrim: It seems weird to me - I think removing the dot prefix would be preferable. Another suggestion would be to create an 'exit' alias for the 'quit' builtin.

tonyhutter:
I've opened an issue for an alternate proposal that would use busybox as our UnifyFS shell here:
https://github.com/LLNL/UnifyFS/issues/426.  Please post any comments about it in that issue, as I don't want to spam this pull request with busybox-related comments.

util/unifyfs-shell/src/unifyfs-shell.c
sandrain: @MichaelBrim Okay, I will remove the dot prefix.

util/unifyfs-shell/src/Makefile.am
sandrain: I have updated the build script to create two executables (-gotcha and -static)

MichaelBrim:
Marking this WIP until we decide if the busybox option (issue #426) is viable

adammoody:
We decided to move in a different direction with this, since LD_PRELOAD and busybox works to some degree.

PR #427, opened 2020-01-08
server tree and collective filesize reduce

adammoody:
Using margo rpcs between servers, this implements a tree-based reduction to compute the file size across servers.

One challenge here is that these operations are on-demand, non-blocking collectives.  When a file size request comes into to a server, it allocates a temporary structure to track the state of the collective.  It assigns a tag, unique to that local server, that the server will use to look up its state structure on later rpcs.  It forwards the tag to any children, and the children include that tag in their reply.

To implement a file size lookup, the server calls a file size request rpc on each of its children in the tree.  Each child in turn forwards that request down the tree.  At the leaves, the servers lookup their local max write offset for the given file and respond to their parent.  On each file size response, a server takes the max of the incoming message with its current max and saves that on the state structure.  Once it receives replies from all children, it forwards a reply to its own parent and releases the state structure.

This type of tree-based bcast/reduce can be used for other operations like truncate and unlink.  If we at some point opt to replicate the metadata info for each file on every server, this could be used for operations that act on the stat info too like chmod, chgrp.  Large read operations could be done with tree-based scatter/bcast of extents, followed by tree-based gathers to collect the data replies.

The second commit in this PR adds server-based segment trees to track file extents on each server.  These segment trees are updated when new file extents are fsync'd from the client.  The file size reduction is modified to use these segment trees to get the max extent offset on each node as input to the reduction.  Additionally, the segment trees define operations for file truncate and unlink to be used in a future commit.

The last commit uses the new server-based segment tree to service client read requests.  For each read request from the client, the server processes as much as it can from its local segment tree.  If any request has left over segments after this local processing, the server falls back to query the global key value store and requests data from other servers.  This eliminates remote key/value lookups in the case a client is reading data local to the node, which should speed up those reads.  Also, it will flatten writes across fsync operations from a single process as well as processes from the same node.  This optimization is not valid in the case that a process on another node writes to the same byte, since the local server would then retrieve its stale local value.  It can be used with truncate, though, so long as the extents are truncated appropriately.

### Types of changes

### Checklist:


server/src/unifyfs_tree.c
```diff
@@ -0,0 +1,352 @@
...
+
+unifyfs_state_filesize_t* alloc_state_filesize(void)
+{
+    state_filesize = (unifyfs_state_filesize_t*) malloc(sizeof(unifyfs_state_filesize_t));
+    return state_filesize;
+}
+
+static void filesize_response_forward(unifyfs_state_filesize_t* st)
+{
+    printf("%d: BUCKEYES response_forward\n", glb_pmi_rank);  fflush(stdout);
```
MichaelBrim: I think we need to find a way to get a "BUCKEYES" printf in the final product ;)

adammoody: Yeah, keen eye :-). Needed someway to find my new debug messages, and for some reason that name stands out to me.  O-H!

adammoody:
@CamStan , I'll check this again.  I see my other PRs are running the tests just fine, so must be something I messed up.

CamStan:
I cleared Travis' cache and ran this again. In watching it, it appears to hang after it gets to that last test that prints successfully, then eventually fails out. I believe the `ERROR: 0100-sysio-gotcha.t - missing test plan` line is getting printed because it's failing before reaching the `done_testing();` step at the end of _stdio_suite.c_, which is where the number of tests in the test plan is determined in our case.

I'm not seeing anything obvious, but I keep running into these situations where the unit tests hang in various places. Still haven't been able to figure out why, but there could be a bigger underlying issue that keeps getting bumped.

adammoody:
@CamStan , I did find a bug that I think was leading to a segfault.  When I cleaned that up, the tests started passing.  So it was my fault after all.  Sorry for the false alarm, but thanks for helping me figure it out.

tonyhutter:
Is there any way you can extend seg_tree to get what you want, rather than copy-n-pasting that code into separate extent_tree/gfid2ext_tree/int2void implementations?  Maybe add a `void * node.data` field to hold custom data, and allow the user to specify a custom comparitor function?

server/src/unifyfs_tree.c
```diff
@@ -476,3 +476,318 @@ static void filesize_response_rpc(hg_handle_t handle)
...
+    /* get info for tree */
+    int parent       = t->parent_rank;
+    int child_count  = t->child_count;
+    int* child_ranks = t->child_ranks;
+
+    /* send up to parent if we have gotten all replies */
+    if (st->num_responses == child_count) {
+        /* truncate the file */
+        int ret = gfid2ext_tree_truncate(&glb_gfid2ext, st->gfid, st->length);
+
```
adammoody: Looks great!  The only thing I can think we might add is to set the st->err flag here if the truncate operation were to fail.

adammoody:
Nice cleanup on the collective state structure, @sandrain !

server/src/unifyfs_request_manager.c
```diff
@@ -683,6 +686,7 @@ int rm_cmd_filesize(
     if (filesize_meta > filesize) {
         filesize = filesize_meta;
     }
```
adammoody: @sandrain , we'll still want to consider the file size field in the metadata even with our new collective.  This field is used in truncate operations, especially for extending a file beyond any extent that was actually written to the file.

server/src/unifyfs_tree.c
```diff
@@ -989,3 +989,268 @@ static void unlink_response_rpc(hg_handle_t handle)
...
+    int parent       = t->parent_rank;
+    int child_count  = t->child_count;
+    int* child_ranks = t->child_ranks;
+
+    /* send up to parent if we have gotten all replies */
+    if (st->num_responses == child_count) {
+        /* metaset the file */
+        int ret = gfid2ext_tree_metaset(&glb_gfid2ext, st->gfid, st->create,
+                                        &st->attr);
+
```
adammoody: We should update the st->err field if we failed to insert/update the metadata here.

common/src/unifyfs_server_rpcs.h
```diff
@@ -111,6 +114,50 @@ MERCURY_GEN_PROC(truncate_response_out_t,
...
+ *
+ * response to metaget request */
+MERCURY_GEN_PROC(metaget_response_in_t,
+                 ((int32_t)(tag))
+                 ((int32_t)(err)))
+MERCURY_GEN_PROC(metaget_response_out_t,
+                 ((int32_t)(ret))
+                 ((unifyfs_file_attr_t)(attr)))
+DECLARE_MARGO_RPC_HANDLER(metaget_response_rpc)
+
```
adammoody: These aren't really used, right?

server/src/unifyfs_cmd_handler.c
```diff
@@ -406,23 +398,17 @@ static void unifyfs_metaset_rpc(hg_handle_t handle)
...
-    fattr.ctime        = in.ctime;
-    fattr.is_laminated = in.is_laminated;
+
+    unifyfs_file_attr_t fattr;
+    fattr = in.attr;
 
     /* if we're creating the file,
      * we initialize both the size and laminate flags */
-    int ret = unifyfs_set_file_attribute(create, create, &fattr);
+    //int ret = unifyfs_set_file_attribute(create, create, &fattr);
```
adammoody: There is some subtle but important logic in the old unifyfs_set_file_attribute() implementation.  In particular, the first two parameters passed to that function are boolean flags that indicate whether the existing size and is_laminated fields should be preserved or replaced.  You can see what I mean if you look at the function code here:

https://github.com/LLNL/UnifyFS/blob/ec593f35f36a59ed5d3aeb84dbf053f50674172e/server/src/unifyfs_metadata.c#L212

I think we'll still want to keep that logic in the new implementation.

adammoody:
@sandrain , I took a quick first pass on the metaset/get rpcs.  This looks good.  I found a few spots that we need to fix up.  Also, there are numerous places where we current call set_file_attributes, like here:

https://github.com/LLNL/UnifyFS/blob/ec593f35f36a59ed5d3aeb84dbf053f50674172e/server/src/unifyfs_request_manager.c#L949

We should convert all of these over to use the new methods.  I didn't check to see if you found them all, but that's something we should do, as well.

server/src/unifyfs_tree.c
```diff
@@ -791,3 +737,255 @@ static void truncate_response_rpc(hg_handle_t handle)
...
+    /* get info for tree */
+    int parent       = t->parent_rank;
+    int child_count  = t->child_count;
+    int* child_ranks = t->child_ranks;
+
+    /* send up to parent if we have gotten all replies */
+    if (st->num_responses == child_count) {
+        /* unlink the file */
+        int ret = gfid2ext_tree_unlink(&glb_gfid2ext, st->gfid);
```
adammoody: Let's set st->err if the the unlink operation fails.

server/src/unifyfs_request_manager.c
sandrain: @adammoody I am a bit confused. Isn't the truncate collective apply the correct size to all extent trees?

adammoody: @sandrain , that's right.  The collective truncate will ensure that the extents do not extend past the end of the file.  However, it's possible that someone truncates the file so that the end of the file falls within a hole, where there is no extent.  On a truncate, we also record the file size in the meta data, and then to compute the file size like for a stat, we take the maximum offset of all extents (via filesize collective) and the maximum of the size value stored in the metadata.  We just need to move this block of code after the collective that gets the max extent offset.

adammoody: It helps to look at the old truncate code here:
https://github.com/LLNL/UnifyFS/blob/ec593f35f36a59ed5d3aeb84dbf053f50674172e/server/src/unifyfs_request_manager.c#L853

Note that it does two things:
1) it deletes/rewrites any extents that extend past the end of the file, in this case by removing/overwriting any keys in MDHIM and
2) it updates the size field in the metadata.

We'll want to do those same two steps here, except that step 1 is now using the collective to update the local extent maps on each server.

adammoody: And to complete the picture, also take a look at the old file size code:
https://github.com/LLNL/UnifyFS/blob/ec593f35f36a59ed5d3aeb84dbf053f50674172e/server/src/unifyfs_request_manager.c#L542
To compute the file size, this does two things:
1) Query all extents in MDHIM to find the max write offset across all extents and
2) query the size field from the file meta data to see if that's even larger

Again in our new collective-based method, we're replacing the query to MDHIM in step 1 with a collective across servers.  However, we still need to do step 2.

common/src/unifyfs_server_rpcs.h
sandrain: right, i will remove these.

server/src/unifyfs_cmd_handler.c
sandrain: I uncommented this function/line for now. Probably, we want to put the logic somewhere else if we are to get rid of mdhim.

server/src/unifyfs_request_manager.c
sandrain: @adammoody I think I understand your point. thanks.

server/src/unifyfs_cmd_handler.c
```diff
@@ -371,7 +371,8 @@ static void unifyfs_metaget_rpc(hg_handle_t handle)
      * from key/value store */
     unifyfs_file_attr_t attr_val;
     //int ret = unifyfs_get_file_attribute(in.gfid, &attr_val);
-    int ret = gfid2ext_tree_metaget(&glb_gfid2ext, in.gfid, &attr_val);
+//    int ret = gfid2ext_tree_metaget(&glb_gfid2ext, in.gfid, &attr_val);
```
adammoody: go ahead and delete this line

server/src/unifyfs_init.c
```diff
@@ -55,7 +55,9 @@ size_t glb_num_servers;     // size of glb_servers array
 server_info_t* glb_servers; // array of server_info_t
 
 /* maps a global file id to its extent map */
-struct gfid2ext_tree glb_gfid2ext;
+//struct gfid2ext_tree glb_gfid2ext;
```
adammoody: go ahead and delete this line

server/src/unifyfs_inode.c
```diff
@@ -0,0 +1,290 @@
...
+        }
+
+        unifyfs_inode_wrlock(ino);
+        {
+            unifyfs_file_attr_update(&ino->attr, attr);
+        }
+        unifyfs_inode_unlock(ino);
+    }
+out_unlock_tree:
+    unifyfs_inode_tree_wrlock(global_inode_tree);
```
adammoody: mean this to be unlock?

server/src/unifyfs_inode.c
```diff
@@ -0,0 +1,290 @@
...
+
+    unifyfs_inode_tree_rdlock(global_inode_tree);
+    {
+        ino = unifyfs_inode_tree_search(global_inode_tree, gfid);
+        if (!ino) {
+            ret = ENOENT;
+            goto out_unlock_tree;
+        }
+
+        unifyfs_inode_rdlock(ino);
```
adammoody: should this be a wrlock?

server/src/unifyfs_inode.c
```diff
@@ -0,0 +1,290 @@
...
+    struct extent_tree* extent_tree = NULL;
+
+    unifyfs_inode_tree_rdlock(global_inode_tree);
+    {
+        ino = unifyfs_inode_tree_search(global_inode_tree, gfid);
+        if (!ino) {
+            goto out_unlock_tree;
+        }
+
+        unifyfs_inode_wrlock(ino);
```
adammoody: can this be a rdlock?

server/src/unifyfs_request_manager.c
```diff
@@ -2152,7 +2159,8 @@ int rm_cmd_fsync(int app_id, int client_side_id, int gfid)
             extent_tree_init(extent_tree);
 
             /* insert emtpy tree into gfid-to-extent map */
-            gfid2ext_tree_add_extent(&glb_gfid2ext, gfid, extent_tree);
+            //gfid2ext_tree_add_extent(&glb_gfid2ext, gfid, extent_tree);
```
adammoody: go ahead and delete this line

server/src/unifyfs_request_manager.c
```diff
@@ -2137,7 +2143,8 @@ int rm_cmd_fsync(int app_id, int client_side_id, int gfid)
         /* TODO: need to lock/unlock gfid2ext tree */
         /* get extent map for this gfid */
         struct extent_tree* extent_tree;
-        extent_tree = gfid2ext_tree_extents(&glb_gfid2ext, gfid);
+        //extent_tree = gfid2ext_tree_extents(&glb_gfid2ext, gfid);
```
adammoody: go ahead and delete this line

server/src/unifyfs_request_manager.c
```diff
@@ -1405,7 +1410,8 @@ int get_local_keyvals(
         /* look up any entries we can find in our local extent map */
         int num_local = 0;
         struct extent_tree* extent_tree;
-        extent_tree = gfid2ext_tree_extents(&glb_gfid2ext, gfid);
+        //extent_tree = gfid2ext_tree_extents(&glb_gfid2ext, gfid);
```
adammoody: go ahead and delete this line

server/src/extent_tree.c
```diff
@@ -317,18 +317,21 @@ int extent_tree_truncate(
     struct extent_tree* tree, /* tree to truncate */
     unsigned long size)       /* size to truncate extents to */
 {
+    unsigned long target_offset = size - 1;
```
adammoody: the caller may pass in size==0, which would cause this to be (unsigned long)-1.  Rather than deal with that confusion, can we update the checks to consider that size is one more than the logical file offset?

server/src/extent_tree.c
```diff
@@ -317,18 +317,21 @@ int extent_tree_truncate(
...
+
     /* lock the tree for reading */
     extent_tree_wrlock(tree);
 
     /* iterate from node with maximum extent backwards until
      * we find an extent below the truncated size */
     struct extent_tree_node* node = RB_MAX(inttree, &tree->head);
-    while (node != NULL && node->end >= size) {
```
adammoody: Actually, I think this check is still ok as is, for example truncate(size=10) can keep any extent whose end is 9 or less.  However, any extent whose end=10 or more should be updated or removed.

server/src/extent_tree.c
```diff
@@ -317,18 +317,21 @@ int extent_tree_truncate(
...
+
+    while (node != NULL && node->end >= target_offset) {
         /* found an extent whose ending offset is equal to or
          * extends beyond the truncated size,
          * check whether the full extent is beyond the truncated
          * size or whether the new size falls within this extent */
-        if (node->start > size) {
```
adammoody: I think this can be replaced with ```node->start >= size```.

server/src/extent_tree.c
```diff
@@ -344,7 +347,7 @@ int extent_tree_truncate(
         } else {
             /* the range of this node overlaps with the truncated size
              * so just update its end to be the new size */
-            node->end = size;
+            node->end = target_offset;
```
adammoody: We could then change this to be ```node->end = size -1;```

server/src/unifyfs_inode_tree.c
```diff
@@ -0,0 +1,224 @@
...
+
+    memset(tree, 0, sizeof(*tree));
+    ret = pthread_rwlock_init(&tree->rwlock, NULL);
+    RB_INIT(&tree->head);
+
+    return ret;
+}
+
+/* Remove and free all nodes in the unifyfs_inode_tree. */
+void unifyfs_inode_tree_destroy(
```
adammoody: Are we also freeing the tree structure itself here, and if so, should we destroy the rwlock?

server/src/unifyfs_tree.c
```diff
@@ -298,18 +265,18 @@ static void filesize_response_forward(unifyfs_coll_state_t* st)
     /* send up to parent if we have gotten all replies */
     if (st->num_responses == child_count) {
         /* lookup max file offset we have for this file id */
-        size_t filesize = get_max_offset(st->gfid);
+        size_t filesize = 0;
+        int ret = unifyfs_inode_get_extent_size(st->gfid, &filesize);
+
+        if (ret) {
+            /* TODO: handle ENOENT */
```
adammoody: Let's set ```st->err = ret``` here.

tonyhutter:
Reiterating my [earlier comment](https://github.com/LLNL/UnifyFS/pull/427/#issuecomment-582505720), are there any plans to extend seg_tree to get rid of the separate, but very similar tree implementations of  extent_tree, int2void, unifyfs_inode_tree?

server/src/extent_tree.c
sandrain: Yes, you are right. Do we need to deallocate the extent tree when the size is zero?

server/src/unifyfs_inode.c
sandrain: Thanks :-)

server/src/unifyfs_inode.c
sandrain: Yes

server/src/unifyfs_inode.c
sandrain: Yes

server/src/unifyfs_inode_tree.c
sandrain: Yes, you are right that we need to destroy the lock. But the structure is allocated by the caller for now.

adammoody:
> Reiterating my [earlier comment](https://github.com/LLNL/UnifyFS/pull/427/#issuecomment-582505720), are there any plans to extend seg_tree to get rid of the separate, but very similar tree implementations of extent_tree, int2void, unifyfs_inode_tree?

That makes sense, but we're saving that step for later.  It may be that the requirements of these different data structures diverge as more functionality is added.  We've kept them separate to allow for that.  We can revisit this idea down the road.

adammoody:
Considering how to code up a replacement for the extra logic we have in unifyfs_set_file_attribute(), it basically comes down to selectively updating the size and is_laminated fields in the metadata.  I think we have the following cases:

client-to-server metaset_rpc with create==1:
- entry must not already exist in the tree (error if it exists)
- set all metadata fields with values from client, including size and is_laminated

client-to-server metaset_rpc with create==0:
- entry must exist in the tree (error if it does not exist)
- update all metadata fields with values from client, ensure that existing size and is_laminated values are not changed (mask out settings from client)

truncate:
- entry must exist in the tree (error if it does not exist)
- only update the size field in metadata, leave all other fields untouched

laminate:
- entry must exist in the tree (error if it does not exist)
- only update the size and is_laminated fields in metadata, leave all other fields untouched

Once we have that, we should be able to replace all calls to unifyfs_set_file_attribute() and unifyfs_get_file_attribute().

tonyhutter:
> That makes sense, but we're saving that step for later. It may be that the requirements of these different data structures diverge as more functionality is added. We've kept them separate to allow for that. We can revisit this idea down the road.

Please include this step in this PR though; it's not worth racking up the Technical Debt, especially since this PR isn't time sensitive.  The seg_tree code already has test cases, so we know it works.  It should be extended in this PR to be a more general tree library that would accommodate these new use cases.  It should be doable if you add a  `void * seg_tree.data` pointer for the custom data, and allow the user to specify a custom comparator function.

server/src/extent_tree.c
adammoody: No, we can allow new extents to come in.  I think we only need to deallocate when destroying the inode.

MichaelBrim:
Closing as margotree work has been merged.

PR #429, opened 2020-01-10
add client-to-server rpcs for unlink and laminate

adammoody:
This adds client-to-server rpcs for "create", unlink, and laminate operations.

We borrow the existing metaset rpc to serve double duty as a "create" rpc.  We add a flag to the rpc parameters to indicate whether the size and laminate fields should be set or not when updating the metadata from a client.  We set both size and is_laminated to 0 when first creating the file, otherwise changing file meta data from the client will not update either the size or is_laminated field.

For unlink, the rpc will remove the file metadata, and if the file is a regular file, it also removes any write extent keys for that file.

For laminate, the rpc gets the current file size and updates the metadata with the file size and it sets the is_laminated flag to 1.

This depends on https://github.com/LLNL/UnifyFS/pull/419

### Types of changes

### Checklist:

### DIFF
--- a/client/src/margo_client.c
+++ b/client/src/margo_client.c
@@ -45,6 +45,16 @@ static void register_client_rpcs(client_rpc_context_t* ctx)
                        unifyfs_truncate_out_t,
                        NULL);
 
+    ctx->rpcs.unlink_id = MARGO_REGISTER(mid, "unifyfs_unlink_rpc",
+                       unifyfs_unlink_in_t,
+                       unifyfs_unlink_out_t,
+                       NULL);
+
+    ctx->rpcs.laminate_id = MARGO_REGISTER(mid, "unifyfs_laminate_rpc",
+                       unifyfs_laminate_in_t,
+                       unifyfs_laminate_out_t,
+                       NULL);
+
     ctx->rpcs.fsync_id = MARGO_REGISTER(mid, "unifyfs_fsync_rpc",
                        unifyfs_fsync_in_t,
                        unifyfs_fsync_out_t,
@@ -273,8 +283,18 @@ int invoke_client_unmount_rpc(void)
     return (int)ret;
 }
 
-/* invokes the client metaset rpc function */
-int invoke_client_metaset_rpc(unifyfs_file_attr_t* f_meta)
+/*
+ * Set the metadata values for a file (after optionally creating it).
+ * The gfid for the file is in f_meta->gfid.
+ *
+ * create: If set to 1, attempt to create the file first.  If the file
+ *         already exists, then update its metadata with the values in
+ *         f_meta.  If set to 0, and the file does not exist, then
+ *         the server will return an error.
+ *
+ * f_meta: The metadata values to update.
+ */
+int invoke_client_metaset_rpc(int create, unifyfs_file_attr_t* f_meta)
 {
     /* check that we have initialized margo */
     if (NULL == client_rpc_context) {
@@ -286,6 +306,7 @@ int invoke_client_metaset_rpc(unifyfs_file_attr_t* f_meta)
 
     /* fill in input struct */
     unifyfs_metaset_in_t in;
+    in.create       = (int32_t) create;
     in.gfid         = f_meta->gfid;
     in.filename     = f_meta->filename;
     in.mode         = f_meta->mode;
@@ -421,6 +442,76 @@ int invoke_client_truncate_rpc(int gfid, size_t filesize)
 
     /* call rpc function */
     LOGDBG("invoking the truncate rpc function in client");
+    hg_return_t hret = margo_forward(handle, &in);
+    assert(hret == HG_SUCCESS);
+
+    /* decode response */
+    unifyfs_filesize_out_t out;
+    hret = margo_get_output(handle, &out);
+    assert(hret == HG_SUCCESS);
+    int32_t ret = out.ret;
+    LOGDBG("Got response ret=%" PRIi32, ret);
+
+    /* free resources */
+    margo_free_output(handle, &out);
+    margo_destroy(handle);
+    return (int)ret;
+}
+
+/* invokes the client unlink rpc function */
+int invoke_client_unlink_rpc(int gfid)
+{
+    /* check that we have initialized margo */
+    if (NULL == client_rpc_context) {
+        return UNIFYFS_FAILURE;
+    }
+
+    /* get handle to rpc function */
+    hg_handle_t handle = create_handle(client_rpc_context->rpcs.unlink_id);
+
+    /* fill in input struct */
+    unifyfs_unlink_in_t in;
+    in.app_id         = (int32_t)app_id;
+    in.local_rank_idx = (int32_t)local_rank_idx;
+    in.gfid           = (int32_t)gfid;
+
+    /* call rpc function */
+    LOGDBG("invoking the unlink rpc function in client");
+    hg_return_t hret = margo_forward(handle, &in);
+    assert(hret == HG_SUCCESS);
+
+    /* decode response */
+    unifyfs_filesize_out_t out;
+    hret = margo_get_output(handle, &out);
+    assert(hret == HG_SUCCESS);
+    int32_t ret = out.ret;
+    LOGDBG("Got response ret=%" PRIi32, ret);
+
+    /* free resources */
+    margo_free_output(handle, &out);
+    margo_destroy(handle);
+    return (int)ret;
+}
+
+/* invokes the client-to-server laminate rpc function */
+int invoke_client_laminate_rpc(int gfid)
+{
+    /* check that we have initialized margo */
+    if (NULL == client_rpc_context) {
+        return UNIFYFS_FAILURE;
+    }
+
+    /* get handle to rpc function */
+    hg_handle_t handle = create_handle(client_rpc_context->rpcs.laminate_id);
+
+    /* fill in input struct */
+    unifyfs_unlink_in_t in;
+    in.app_id         = (int32_t)app_id;
+    in.local_rank_idx = (int32_t)local_rank_idx;
+    in.gfid           = (int32_t)gfid;
+
+    /* call rpc function */
+    LOGDBG("invoking the laminate rpc function in client");
     hg_return_t hret = margo_forward(handle, &in);
     assert(hret == HG_SUCCESS);
 
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -818,15 +818,35 @@ int unifyfs_fid_update_file_meta(int fid, unifyfs_file_attr_t* gfattr)
     return UNIFYFS_FAILURE;
 }
 
-int unifyfs_set_global_file_meta(int gfid, unifyfs_file_attr_t* gfattr)
+/*
+ * Set the metadata values for a file (after optionally creating it).
+ * The gfid for the file is in f_meta->gfid.
+ *
+ * gfid:   The global file id on which to set metadata.
+ *
+ * create: If set to 1, attempt to create the file first.  If the file
+ *         already exists, then update its metadata with the values in
+ *         gfattr.  If set to 0, and the file does not exist, then
+ *         the server will return an error.
+ *
+ * gfattr: The metadata values to store.
+ */
+int unifyfs_set_global_file_meta(
+    int gfid,   /* file id to set meta data for */
+    int create, /* whether to set size/laminated fields (1) or not (0) */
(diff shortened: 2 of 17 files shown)


adammoody:
@MichaelBrim , @tonyhutter , now that the truncate PR is merged, this one is able to go.  Would you please take a look at this when you get time?

client/src/margo_client.c
```diff
@@ -286,6 +296,7 @@ int invoke_client_metaset_rpc(unifyfs_file_attr_t* f_meta)
 
     /* fill in input struct */
     unifyfs_metaset_in_t in;
+    in.create       = (int32_t) create;
```
tonyhutter: Question: what if you didn't have a `in.create` field, and simply created the file if it didn't exist?  Or are there times you would want to either call metaset on a non-existant file, or create=1 on an existing file?

client/src/margo_client.c
```diff
@@ -437,6 +448,76 @@ int invoke_client_truncate_rpc(int gfid, size_t filesize)
...
+    int32_t ret = out.ret;
+    LOGDBG("Got response ret=%" PRIi32, ret);
+
+    /* free resources */
+    margo_free_output(handle, &out);
+    margo_destroy(handle);
+    return (int)ret;
+}
+
+/* invokes the client-to-server laminate rpc function */
```
tonyhutter: Add to this comment: "We assume the gfid is not laminated."  ..to make it clear that the developer should first check if the file is laminated before doing the RPC (we currently do this check in chmod()).

client/src/margo_client.c
adammoody: Currently, invoke_client_metaset_rpc is serving two roles as both our "file create" and "update metadata" rpc depending on whether the user is specifying create=1 or not.

With create=1, the server updates all metadata in the global key/value store, including the file size and is_laminated flags.  It creates the entry if it doesn't exist, and it silently overwrite any existing entry.

With create=0, the server will preserve the size and is_laminated fields for any existing entry in the global/key value store, but it updates all other fields.  If there is no existing entry already, it uses the size/is_laminated values provided by the caller.

For a true "create" operation, it would probably be desirable to have the server return an error if the item already exists.  However, we have a problem in that our key/value store does not yet support atomic read/modify/write operations, which we'd need to safely enable this.

server/src/unifyfs_metadata.c
```diff
@@ -209,15 +209,45 @@ int meta_sanitize(void)
...
 {
     int rc = UNIFYFS_SUCCESS;
 
     /* select index for file attributes */
     md->primary_index = unifyfs_indexes[IDX_FILE_ATTR];
 
-    /* insert file attribute for given global file id */
     int gfid = fattr_ptr->gfid;
+
+    /* if we're want to preserve some setting,
```
tonyhutter: "If we want..."

client/src/margo_client.c
adammoody: I am thinking it'd be fine to have the laminate rpc act as a NOP in the case the file is already laminated, so it's fine to call the laminate_rpc multiple times on a file.  We could return an error like ELAMINATED in that case if the user wanted to know.

tonyhutter: So are you saying you'd add a check in `invoke_client_laminate_rpc()` to see if the locally cached `is_laminated` is already set, and just return immediately?  Or are you saying doing a laminate RPM on an already laminated file is harmless and essentially a NOP?

adammoody: Right now the server-side rpc just happily runs through all of the laminate logic a second time.  That won't hurt.  However, we could probably tweak this to check first, skip it if already laminated, and return an ELAMINATED error.  Can look at that later.

server/src/unifyfs_metadata.c
adammoody: Nice catch.

client/src/margo_client.c
adammoody: After double-checking the code, with create=0, it returns an error from the set metadata rpc if it does not find an existing entry, which is good.

client/src/margo_client.c
tonyhutter: Ok, thanks

adammoody: For now, I'll just leave this as the "run though again, but it doesn't hurt to re-laminate a laminated file" implementation.

client/src/margo_client.c
tonyhutter: Does it also error if you do create=1 when the file already exists?

adammoody: No, it behaves like an O_CREAT rather than an "O_CREAT | O_EXCL".  There is some other code within the server that depends on that at the moment.

I can imagine we'll want to be able to have both options but I'll save that extension for a later PR, especially since we can't do atomic creation operations with the current key/value store.

tonyhutter: Ok.  Is there any way we can make it clearer/simplify what metaset_rpc does?  Proposal:
1. Remove `unifyfs_metaset_in_t.is_laminated`, to explicitly disallow setting `is_laminated` from metaset_rpc.  That way, there's only one way to laminate (via laminate_rpc) and we can't shoot ourselves in the foot.

2. On the server, always update the file size on a metaset_rpc, *except* if the file is laminated.  Return an error in that case.

3. Add a more detailed comment describing the function:
```C
/* 
 * Set the metadata values for a file (after optionally creating it).  The
 * gfid for the file is in f_meta->gfid.
 * 
 * create:	If set to 1, attempt to create the file first.  If the file
 *              already exists, then update its metadata with the values in
 *              f_meta.  If set to 0, and the file does not exist, then
 *              the server will return an error.
 *
 * f_meta:	The metadata values to update.
 */
 ```

Comment (adammoody) [reply client/src/margo_client.c]: Thanks for the good function description.  I added that to margo_client.c file.  The header is comment-free at the moment, so putting it in the header makes me feel like I should go comment all of the functions there.

Before submitting this PR, I had considered removing the size and is_laminated fields from the metaset rpc input parameter list.  However, I thought there could be some case in the future where being able to set those fields from the client would be helpful.  I decided to leave them in for that reason.  I feel like the real solution would be to bite the bullet and define an actual mknod_rpc that would be the pair of the unlink_rpc.  Then we could make the metaset_rpc only update existing items and not create new ones.

Also for the metaset rpc, as @MichaelBrim was suggesting, we'll eventually want to be able to selectively update particular fields (e.g., change the gid but nothing else).  For that, we'd want to include some sort of bit flags in the input parameters or something...  That's more than I wanted to get into here.  For this PR, I just want metaset to stop changing the size field.

File: client/src/unifyfs.c [hunk 4]
```diff
@@ -1363,6 +1355,15 @@ int unifyfs_fid_unlink(int fid)
 {
     int rc;
 
+    /* invoke truncate rpc */
```
MichaelBrim: s/truncate/unlink/ in comment

common/src/unifyfs_client_rpcs.h
```diff
@@ -129,6 +130,30 @@ MERCURY_GEN_PROC(unifyfs_truncate_out_t,
                  ((int32_t)(ret)))
 DECLARE_MARGO_RPC_HANDLER(unifyfs_truncate_rpc)
 
+/* unifyfs_unlink_rpc (client => server)
+ *
+ * given an app_id, client_id, and global file id,
+ * unlink the file */
+MERCURY_GEN_PROC(unifyfs_unlink_in_t,
+                 ((int32_t)(app_id))
```
MichaelBrim: Neither `app_id` nor `local_rank_idx` are necessary parameters. seems they are there to support the calls to `rm_cmd_unlink()` and `rm_cmd_truncate()`, but those functions don't use the parameters.

adammoody: Worth holding on to the app_id for future proofing for namespacing purposes of different mount points?

MichaelBrim: good point. It's probably best to keep them. I think I even have updated code in the client API branch that needs that info.

tonyhutter:
Can you please add some test cases:

1. Write to a file, sync, close. unlink it.  Attempt to access the file via stat() (should fail).
2. Write to a file, sync, unlink it without closing. Attempt to access the file (should fail).
3. Write to a file, unlink it without closing. Attempt to access the file (should fail).
4. Write to a file, unlink it, sync it, close it .  The sync should fail.
5. Try to delete a laminated file.  (I don't know what the expected behavior would be).

adammoody:
> Can you please add some test cases:
> 
> 1. Write to a file, sync, close. unlink it.  Attempt to access the file via stat() (should fail).
> 2. Write to a file, sync, unlink it without closing. Attempt to access the file (should fail).
> 3. Write to a file, unlink it without closing. Attempt to access the file (should fail).
> 4. Write to a file, unlink it, sync it, close it .  The sync should fail.
> 5. Try to delete a laminated file.  (I don't know what the expected behavior would be).

Only 1 and 5 apply here.  This PR is only intending to define and invoke a new client-to-server unlink_rpc so the server can clean out state associated with the file.  The other items will require development on the client or modifying the fsync rpc.

tonyhutter:
@adammoody I'm confused.  From the description it sounds like it's doing a full delete:

> For unlink, the rpc will remove the file metadata, and if the file
> is a regular file, it also removes any write extent keys for that file.

Wouldn't 2-4 also be valid tests then?

adammoody:
> 
> 
> @adammoody I'm confused. From the description it sounds like it's doing a full delete:
> 
> > For unlink, the rpc will remove the file metadata, and if the file
> > is a regular file, it also removes any write extent keys for that file.
> 
> Wouldn't 2-4 also be valid tests then?

The fsync rpc just blindly adds any extents to the key/value store, regardless of whether the file actually exists or not.
https://github.com/LLNL/UnifyFS/blob/b9a37537ce935780b64867e7f314b22247037e32/server/src/unifyfs_request_manager.c#L1237
So the unlink_rpc will in fact remove the file metadata and any existing extents, but the next fsync_rpc that comes along will just add any extents for the deleted file back in.  It needs to be updated to check that the file actually exists, and then add extents.  Another problem is that our fsync() is more like a sync() in that it pushes extents in for all files, not just the one the user named.

Also, our support for a real unlink() in general is still very broken.  For one, we don't actually free any storage that is allocated to a file.

client/src/margo_client.c
```diff
@@ -273,8 +283,18 @@ int invoke_client_unmount_rpc(void)
     return (int)ret;
 }
 
-/* invokes the client metaset rpc function */
-int invoke_client_metaset_rpc(unifyfs_file_attr_t* f_meta)
+/*
+ * Set the metadata values for a file (after optionally creating it).
+ * The gfid for the file is in f_meta->gfid.
+ *
+ * create:If set to 1, attempt to create the file first.  If the file
```
tonyhutter: Add space after "create:" and "f_meta:" in comment.

client/src/margo_client.c
```diff
@@ -273,8 +283,18 @@ int invoke_client_unmount_rpc(void)
...
+ *        already exists, then update its metadata with the values in
+ *        f_meta.  If set to 0, and the file does not exist, then
```
tonyhutter: `If set to 0, and the file does not exist, then the server will return an error.` - I can't remember, is this the current behavior with this patch?  If not, then we should remove that sentence.

client/src/unifyfs.c
```diff
@@ -818,15 +818,18 @@ int unifyfs_fid_update_file_meta(int fid, unifyfs_file_attr_t* gfattr)
     return UNIFYFS_FAILURE;
 }
 
-int unifyfs_set_global_file_meta(int gfid, unifyfs_file_attr_t* gfattr)
```
tonyhutter: Can you add a variation of the `invoke_client_metaset_rpc()` comment here to better explain the `create` value, and what this function does?

tonyhutter: Alternatively, you could split this into two functions and get rid of the create field:
```
int unifyfs_set_global_file_meta(int gfid, unifyfs_file_attr_t* gfattr)    /* create=0 */
int unifyfs_update_global_file_meta(int gfid, unifyfs_file_attr_t* gfattr) /* create=1 */
```

client/src/margo_client.c
adammoody: that's still true

client/src/unifyfs.c
adammoody: Took the first option of adding a comment.  I imagine we'll want to turn this into a bit field or something to allow the caller to pick and choose exactly which subset of fields should be set.  But that's enough work for a separate PR.

adammoody:
@MichaelBrim , you ok with this being merged?

PR #433, opened 2020-01-14
fsync file data before truncate

adammoody:
This uses the unifyfs_sync(gfid) function to fsync data before truncating a file.  It updates the truncate(), ftruncate() and fsync() wrappers to call unifyfs_sync(gfid).

It ensures that the spillover file is synced to disk in unifyfs_fsync() before it invokes the client-to-server fsync rpc.  Without this, the data that was meant to be written to the spillover file might still be cached in memory and not on disk when another process attempts to read it.  We should be sure it's on disk before it's registered with the server.

Since the index rewrite that happens during sync rewrites entries for all files, this uses the global segment count across all files for the test that triggers a sync.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-fixed.c
+++ b/client/src/unifyfs-fixed.c
@@ -486,8 +486,7 @@ static int unifyfs_split_index(
  * Clear all entries in the log index.  This only clears the metadata,
  * not the data itself.
  */
-static void
-unifyfs_clear_index(void)
+static void unifyfs_clear_index(void)
 {
     *unifyfs_indices.ptr_num_entries = 0;
 }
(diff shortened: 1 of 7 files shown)


adammoody:
@tonyhutter , I realized there was a clean way after all to support the sync-before-truncate even for the truncate(path) option.  This adds that and cleans up some other stuff.  I think we may still have a couple bugs here I need to work out before review/merge.

adammoody:
@tonyhutter , can you also look these changes over when you get a chance?  Thanks.

adammoody:
@tonyhutter , thanks for the fast response!

PR #434, opened 2020-01-14
client: satisfy read requests from local extent info if possible

adammoody:
This adds another segment tree to the client metadata of each file (fid) that tracks all extents written to the file.  We use this to satisfy later read requests by having the client refer to this segment tree and then copy data directly from its data logs.  Any request that cannot be fully satisfied on the client is sent to the server for processing.

The goal is to make reads as fast as writes in the case that a client process reads back the same data it wrote.

This optimization cannot be used as currently written for apps that call truncate, because the server has no way to inform the client that it should truncate its extent list.  We'll need to be able to turn this on so apps can opt-in.  Added new UNIFYFS_CLIENT_LOCAL_EXTENTS=1 option to enable local read back. This is off by default.

This moves read request splitting from the client to the server.  Splitting the read requests into 1MB chunks was limiting performance, and this splitting is not necessary when the client is reading its own data.

This also moves the write index splitting from the client to the server.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-internal.h
+++ b/client/src/unifyfs-internal.h
@@ -266,26 +266,27 @@ typedef struct {
 } unifyfs_chunkmeta_t;
 
 typedef struct {
-    off_t global_size;               /* Global size of the file */
-    off_t local_size;                /* Local size of the file */
-    off_t log_size;                  /* Log size.  This is the sum of all the
-                                      * write counts. */
-    pthread_spinlock_t fspinlock;    /* file lock variable */
-    enum flock_enum flock_status;    /* file lock status */
+    off_t global_size;            /* Global size of the file */
+    off_t local_size;             /* Local size of the file */
+    off_t log_size;               /* Log size.  This is the sum of all the
+                                   * write counts. */
+    pthread_spinlock_t fspinlock; /* file lock variable */
+    enum flock_enum flock_status; /* file lock status */
 
-    int storage;                     /* FILE_STORAGE type */
+    int storage;                  /* FILE_STORAGE type */
 
-    int gfid;                        /* global file id for this file */
-    int needs_sync;                  /* have unsynced writes */
+    int gfid;                     /* global file id for this file */
+    int needs_sync;               /* have unsynced writes */
 
-    off_t chunks;                   /* number of chunks allocated to file */
-    off_t chunkmeta_idx;            /* starting index in unifyfs_chunkmeta */
-    int is_laminated;               /* Is this file laminated */
-    uint32_t mode;                  /* st_mode bits.  This has file
-                                     * permission info and will tell you if this
-                                     * is a regular file or directory. */
-    struct seg_tree seg_tree;       /* Segment tree containing our coalesced
-                                     * writes */
+    off_t chunks;                 /* number of chunks allocated to file */
+    off_t chunkmeta_idx;          /* starting index in unifyfs_chunkmeta */
+    int is_laminated;             /* Is this file laminated */
+    uint32_t mode;                /* st_mode bits.  This has file
+                                   * permission info and will tell you if this
+                                   * is a regular file or directory. */
+    struct seg_tree extents_sync; /* Segment tree containing our coalesced
+                                   * writes between sync operations */
+    struct seg_tree extents;      /* Segment tree of all local data extents */
 } unifyfs_filemeta_t;
 
 /* struct used to map a full path to its local file id,
@@ -393,7 +394,8 @@ extern int unifyfs_use_memfs;
 extern int unifyfs_use_spillover;
 
 extern int    unifyfs_max_files;  /* maximum number of files to store */
-extern bool   unifyfs_flatten_writes;    /* enable write flattening */
+extern bool   unifyfs_flatten_writes; /* enable write flattening */
+extern bool   unifyfs_local_extents;  /* enable tracking of local extents */
 extern size_t
 unifyfs_chunk_mem;  /* number of bytes in memory to be used for chunk storage */
 extern int    unifyfs_chunk_bits; /* we set chunk size = 2^unifyfs_chunk_bits */
@@ -483,6 +485,10 @@ const char* unifyfs_path_from_fid(int fid);
 
 /* Given a fid, return a gfid */
 int unifyfs_gfid_from_fid(const int fid);
+
+/* returns fid for corresponding gfid, if one is active,
+ * returns -1 otherwise */
+int unifyfs_fid_from_gfid(const int gfid);
 
 /* given an UNIFYFS error code, return corresponding errno code */
 int unifyfs_err_map_to_errno(int rc);
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -44,6 +44,7 @@
 #include "unifyfs-sysio.h"
 #include "margo_client.h"
 #include "ucr_read_builder.h"
+#include "seg_tree.h"
 
 /* -------------------
  * define external variables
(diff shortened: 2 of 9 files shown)


adammoody:
```
+ libexec/writeread-gotcha -p n1 -n 4 -c 524288 -b 16777216 -a 0 -m /unifyfs -f write-static_pn1_n32_c1MB_b16MB.app
File create time is 0.000 s
Aggregate Write BW is 2160.192 MiB/s
Effective Write BW is 2160.192 MiB/s

Minimum Sync Time is 0.001107 s
Maximum Sync Time is 0.001107 s

Aggregate Read BW is 1800.788 MiB/s
Effective Read BW is 1800.788 MiB/s

File laminate time is 0.000 s
```

MichaelBrim:
@adammoody this looks like a pretty nice improvement in read bandwidth. for server-side request splitting: it would be nice if you could make that a compile-time decision on whether to do the splitting. I expect if we go the route of only having extent-data stored at each server, then there is no technical reason to split the requests.

adammoody:
> @adammoody this looks like a pretty nice improvement in read bandwidth. for server-side request splitting: it would be nice if you could make that a compile-time decision on whether to do the splitting. I expect if we go the route of only having extent-data stored at each server, then there is no technical reason to split the requests.

Yes, I plan to avoid splitting if at all possible.  There is some splitting needed right now, but only to define the keys that are used to query MDHIM.  So it's not needed unless we are querying MDHIM.

adammoody:
Performance when using all cores on a node.
```
+ srun -n 36 -N 1 libexec/writeread-gotcha -p n1 -n 4 -c 1048576 -b 16777216 -a 0 -m /unifyfs -f write-static_pn1_n32_c1MB_b16MB.app
File create time is 0.005 s
Aggregate Write BW is 42589.578 MiB/s
Effective Write BW is 36661.628 MiB/s

Minimum Sync Time is 0.001299 s
Maximum Sync Time is 0.038337 s

Aggregate Read BW is 39936.228 MiB/s
Effective Read BW is 39345.606 MiB/s

File laminate time is 0.007 s
```

adammoody:
I'll open this up for (early) review, because it's mostly ready and passing tests.

I still need to add a configuration setting so that this can be enabled/disabled at runtime.  It's hardcoded to be on right now.

Also, a memory debugging tool shows that I'm leaking extent structs, perhaps just because I'm not freeing them on unlink?  Need to chase that down.

adammoody:
Added new UNIFYFS_CLIENT_LOCAL_EXTENTS=1 option to enable local read back.  This is off by default.  I can see the extents are not freed during unlink, because the fid_unlink is bailing out since the file is laminated.  We can revisit this once the improved unlink support is added.

adammoody:
Moved the write index splitting from client to server.

adammoody:
@MichaelBrim , @tonyhutter , @CamStan , this one is ready for review.  I'll revist the memory leak I found after we merge the unlink PR, which I think would resolve it.

This is only failing on a checkpatch of the new configuration option I added.  I can't remember the secret to get that to pass off hand.

Would you please reivew this when you get a chance?

common/src/unifyfs_configurator.h
```diff
@@ -71,6 +71,7 @@
     UNIFYFS_CFG_CLI(unifyfs, mountpoint, STRING, /unifyfs, "mountpoint directory", NULL, 'm', "specify full path to desired mountpoint") \
     UNIFYFS_CFG(client, max_files, INT, UNIFYFS_MAX_FILES, "client max file count", NULL) \
     UNIFYFS_CFG(client, flatten_writes, BOOL, on, "flatten writes", NULL) \
+    UNIFYFS_CFG(client, local_extents, BOOL, off, "track extents to read back local data", NULL) \
```
CamStan: Mind adding this to the client section in the configuration documentation as well?

https://github.com/LLNL/UnifyFS/blob/dev/docs/configuration.rst#unifyfsconf

client/src/unifyfs-internal.h
```diff
@@ -266,26 +266,27 @@ typedef struct {
...
+
+    off_t chunks;                 /* number of chunks allocated to file */
+    off_t chunkmeta_idx;          /* starting index in unifyfs_chunkmeta */
+    int is_laminated;             /* Is this file laminated */
+    uint32_t mode;                /* st_mode bits.  This has file
+                                   * permission info and will tell you if this
+                                   * is a regular file or directory. */
+    struct seg_tree seg_tree;     /* Segment tree containing our coalesced
+                                   * writes */
+    struct seg_tree extents;      /* Segment log of all local data */
```
tonyhutter: We should rename these, since they're both seg_trees and both store extents.  Maybe:
```
struct seg_tree write_tree;     /* Segment tree of the writes since last sync */
struct seg_tree all_write_tree; /* Segment tree of all the writes we have ever done */
```

common/src/seg_tree.c
```diff
@@ -217,6 +217,23 @@ int seg_tree_add(struct seg_tree* seg_tree, unsigned long start,
     return 0;
 }
 
+struct seg_tree_node* seg_tree_find(
```
tonyhutter: Please add a comment describing what this function does and returns.

server/src/unifyfs_request_manager.c
```diff
@@ -1022,6 +1022,142 @@ int create_gfid_chunk_reads(reqmgr_thrd_t* thrd_ctrl,
...
+/* return number of slice ranges needed to cover range */
+static size_t num_slices(size_t offset, size_t length)
+{
+    size_t start = offset / max_recs_per_slice;
+    size_t end   = (offset + length - 1) / max_recs_per_slice;
+    size_t count = end - start + 1;
+    return count;
+}
+
+static int split_request(
```
tonyhutter: Please add a comment on what this function does, it's arguments, and its return value.

server/src/unifyfs_request_manager.c
```diff
@@ -1022,6 +1022,142 @@ int create_gfid_chunk_reads(reqmgr_thrd_t* thrd_ctrl,
...
+
+        /* advance to first byte offset of next slice */
+        pos = end + 1;
+    }
+
+    /* return number of keys we generated */
+    return count;
+}
+
+/* split a write index at slice range boundaries */
```
tonyhutter: Please document this function's arguments and return value.

client/src/unifyfs-internal.h
adammoody: Went with "extents_sync" and "extents" here.  Sound good?

tonyhutter: Yea, I like that even better

adammoody:
@tonyhutter , since you have also been reviewing this, I'll wait for you to sign off before we merge it.  I pushed the changes you're requested so far as separate commits.  I'll squash those before we merge.

client/src/unifyfs-sysio.c
```diff
@@ -1600,59 +1464,335 @@ static int process_read_data(read_req_t* read_reqs, int count, int* done)
     return rc;
 }
 
+/* This uses information in the extent map for a file on the client to
+ * complete any read requests.  It only complets a request if it contains
+ * all of the data.  Otherwise the request is copied to the list of
+ * requests to be handled by the server. */
+static int service_local_reqs(
```
tonyhutter: It looks like this always returns `UNIFYFS_SUCCESS`, so make this a void function.

tonyhutter: ...or alternatively return `out_count` and remove it as a function argument

common/src/seg_tree.c
```diff
@@ -217,6 +217,40 @@ int seg_tree_add(struct seg_tree* seg_tree, unsigned long start,
     return 0;
 }
 
+/* search tree for entry that overlaps with given start/end
+ * offsets, return first overlapping entry if found, NULL otherwise,
+ * assumes caller has lock on tree */
```
tonyhutter: Is there a reason why you don't want to do the locking inside of this function?  All other seg_tree functions do the locking for you, unless there's an explicit reason for them not to.  Also, please capitalize sentences.

common/src/seg_tree.c
```diff
@@ -217,6 +217,40 @@ int seg_tree_add(struct seg_tree* seg_tree, unsigned long start,
...
+     * the target range (starting byte), or otherwise the
+     * node for the next biggest starting byte */
+    struct seg_tree_node* next = RB_NFIND(inttree, &seg_tree->head, node);
+
+    free(node);
+
+    /* we may have found a node that doesn't include our starting
+     * byte offset, but it would be the range with the lowest
+     * starting offset after the target starting offset, check whether
+     * this overlaps our end offset */
```
tonyhutter: This is super clever :+1:

common/src/unifyfs_configurator.h
tonyhutter: We should document somewhere (I don't know where's best) about the conditions under which it's safe to enable this option.  We definitely want all users to be aware of it since it's such a huge performance win.

adammoody: Yeah, I agree.  It doesn't feel like this table is the place to do it, since it would get too cluttered.  I think we need another breakout section where we can discuss things in more detail.  @CamStan , do you have an suggestion?

docs/configuration.rst
```diff
@@ -68,6 +68,7 @@ a given section and key.
    ==============  ======  =================================================================
    max_files       INT     maximum number of open files per client process
    flatten_writes  BOOL    enable flattening writes (optimization for overwrite-heavy codes)
+   local_extents   BOOL    enable extent tracking to service reads of local data
```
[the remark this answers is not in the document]
adammoody: Oh, well this was the table I was thinking of in the above comment.  Anyway, I think it would be nice to have a place to write more text for these options if needed, and then we could hyperlink to that from this table.

common/src/seg_tree.c
adammoody: The motivation here was thinking about cases where one would be looking up a large set of extents, where this would be called in a loop where locking outside of the loop would help with performance:
```
lock tree
foreach extent in many extents
  seg_tree_find(extent)
unlock tree
```
I agree that we should make things consistent, though, so let me check on this again.

tonyhutter: I don't think this is going to be an issue.  The lock acquisition is going to pale in comparison to the number of cycles needed for the tree traversal.   We can always add a `seg_tree_find_nolock()` if performance testing shows that it is a bottleneck.

adammoody: Ok, checking the code, actually there is a second reason I did this.  The actual usage looks like this:
```
lock tree
seg_tree_find
iterate over segments to compute something
seg_tree_find
iterate over segments again
unlock tree
```
The first loop iterates over the extents to figure something out, and if true, iterates over them again to do something with them.  To ensure consistency, it holds a read lock through all of this, so the lock happens on the outside of the find/iterate loops.

adammoody: See the code in service_local_reqs() in unifyfs-sysio.c.  So I think it has to be done that way.

client/src/unifyfs-sysio.c
adammoody: updated to return void

common/src/seg_tree.c
adammoody: Thanks.  Only thought to come up with that because I had a bug in my code before figuring this out.

docs/configuration.rst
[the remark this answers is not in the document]
adammoody: Updated to ```service reads from local data if possible (default: off)```

common/src/seg_tree.c
tonyhutter: Ok, that's a legit use case.  Can you have this patch add both variants then:
```
seg_tree_find()
seg_tree_find_nolock()  (or whatever you want to name this)
```

adammoody: I'll compromise with you.  I renamed the existing function to set_tree_find_nolock().  I'll defer adding seg_tree_find() to a later PR since we're not using it yet.

common/src/unifyfs_configurator.h
adammoody: Added some more text immediately below the table that lists the option.  This keeps the table concise, but also keeps the limitation close to where the option is listed.  You can see how it shows up here.

https://github.com/LLNL/UnifyFS/blob/clientread/docs/configuration.rst

I tried to also hyperlink the option name to the text where the description lives, but I couldn't tell whether it was really working so removed it.

CamStan: I agree that the best place would be where you put it. If we get more of these and the page looks cluttered, we can rethink. If you want it to stand out a bit more, could always put it in a caution/attention admonition, but probably not necessary.

E.g.:
```
.. Caution::

    Text inside admonition goes here. Indent and put a blank line 
    after to separate from subsequent text
```

tonyhutter:
@adammoody I should have asked this earlier, but what's you test plan for this?  I don't think you need to add any additional tests specifically, but you do need a way to exercise your code in Travis.  Can you make it so Travis will re-run all the standard IO tests with `local_extents=1`?

CamStan:
@tonyhutter, we can always do that in the integration tests as well/instead if it's better to do it with multiple nodes/processes.

adammoody:
@tonyhutter , you bring up a great point about needing to expand our test suites to handle this permutation of configuration settings.  That's a bigger task in general, so I created a separate issue to work through that.

As for testing on this particular features, in earlier versions of this PR, I originally had the client local_extents optimization enabled by default and it was passing out test suite.  I later decided that we should disable it by default and have the users opt in, so the more recent passing tests were with the option disabled.

adammoody:
@tonyhutter , @CamStan , @MichaelBrim , thanks for everyone's time and effort in reviewing this.  We hardened it quite a bit!

Issue #432, opened 2020-01-13
Group discussion of what a file size is


adammoody:
This is a placeholder to discuss what file sizes are (implied and explicit).  The current status looks like the following:

```
fid_logical_size() -- returns meta->global_size if laminated, else meta->local_size
- lseek(SEEK_END)
- fseek(SEEK_END)
- stream_write on stream opened with APPEND
- write() on fd opened in APPEND mode
- fid_open(O_APPEND) and file is defined locally and globally
- fd_read to check that read is not reading past end of file

invoke_client_filesize_rpc()
- get_meta_with_size, if file not laminated
- fd_logreadlist (if nread < length)
- __chmod on lamination

get_meta_with_size() -- invokes filesize_rpc if not laminated
- __stat

fid_update_file_meta() -- updates meta->global_size and meta->local_size if laminated
- fid_open, if global meta is found
- __stat, after call to get_meta_with_size

set_global_file_meta_from_fid() -- sets KV size field=meta->global_size if laminated, 0 otherwise
- fid_create_directory
- fid_open, if file not found in global meta
- __chmod ** chmod type of operations will effectively break new truncate maker

truncate_rpc sets KV size field
- fid_truncate
-- fid_unlink
-- fid_open(O_TRUNC) and file is defined locally and globally
-- truncate ** fid may not be defined yet (error)
-- ftruncate
```

PR #437, opened 2020-01-14
Newsize

adammoody:
This is a start of a PR to make the notion of file size consistent.  We'll need some group-wide discussion on how we want size to be defined in various places.

Related to https://github.com/LLNL/UnifyFS/issues/432

For the case that we need to support for HDF, where HDF requires an accurate file size from stat() before it has laminated the file, things get weird currently.  Our code can return one size with stat, but then if a process reads until EOF, it would deduce there is a different file size.  I'm guessing that inconsistency will be problematic, so this adds code so that the size is consistent in all of those instances (or at least that's the goal).  Thus, it ignores local_size.  It can still use global_size if the file has been laminated.

However, going to the server to get the current file size is more expensive, and codes that do not care about file size before lamination could make do with our existing approach.  If we want to keep both options, we'll need a runtime parameter to pick between them, and we'll want to clearly spell out the semantics in both cases.

This PR fails our test suite because it changes things from using local size to the current true file size as known by the servers.

As a concrete example consider this case:  Rank 0 writes to offsets [0, 10), and rank 1 writes to offsets [10,20).  Both ranks then call fsync() and barrier.  At this point, if rank 0 calls stat() it will get the true file size, which is 20.  However, if rank 0 opens the file and starts reading from offset 0, it will hit an EOF at offset=10.  So stat says the file is 20 bytes long, but read acts like the file is only 10 bytes long.  I'm guessing that will eventually be a problem for someone.  This PR changes the read so that it can use the true file size when determining EOF.

Having said that, there might also be value in keeping the notion of the local size.  Asking for the true file size on a read is somewhat expensive, because it will invoke a global collective across the servers.  For applications that don't require knowing the true file size, they might get by with the notion of our "local size".  To support that, this PR keeps the local size in as a runtime option.  But we might want to also change stat() in that case to return the local size, again to be consistent.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -617,17 +617,6 @@ off_t unifyfs_fid_global_size(int fid)
     return (off_t)-1;
 }
 
-/* Return the log size of the file */
-off_t unifyfs_fid_local_size(int fid)
-{
-    /* get meta data for this file */
-    unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-    if (NULL != meta) {
-        return meta->local_size;
-    }
-    return (off_t)-1;
-}
-
 /*
  * Return the size of the file.  If the file is laminated, return the
  * laminated size.  If the file is not laminated, return the local
@@ -639,14 +628,7 @@ off_t unifyfs_fid_logical_size(int fid)
     if (unifyfs_fid_is_laminated(fid)) {
         return unifyfs_fid_global_size(fid);
     } else {
-        /* TODO: move this to a runtime configurable? */
-        int use_local_size = 0;
-        if (use_local_size) {
-            return unifyfs_fid_local_size(fid);
-        }
-
-        /* otherwise, we invoke an rpc to ask the server
-         * what the file size is */
+        /* invoke an rpc to ask the server what the file size is */
 
         /* get gfid for this file */
         int gfid = unifyfs_gfid_from_fid(fid);
@@ -715,7 +697,6 @@ int unifyfs_fid_update_file_meta(int fid, unifyfs_file_attr_t* gfattr)
         if (meta->is_laminated) {
             /* update file size */
             meta->global_size = (off_t)gfattr->size;
-            meta->local_size = meta->global_size;
             LOGDBG("laminated file size is %zu bytes",
                    (size_t)meta->global_size);
         }
@@ -880,7 +861,6 @@ int unifyfs_fid_create_file(const char* path)
     /* initialize meta data */
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
     meta->global_size  = 0;
-    meta->local_size   = 0;
     meta->log_size     = 0;
     meta->flock_status = UNLOCKED;
     meta->storage      = FILE_STORAGE_NULL;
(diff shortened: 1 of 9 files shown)


tonyhutter:
I agree with this proposal.  Returning the correct size every time, even if it makes reads a little more costly, is also going to save us from user confusion, and may prevent some problems down the road when we have a real shell running on /unifyfs. Hopefully with the distributed size lookup we've been talking about, a size lookup really won't be that expensive after all.  I would refrain from making local_size a runtime option for size reporting though, since it's yet another permutation we'd need to test, without a real-world user need.  If someone really really wants it we could talk about adding it back.

client/src/unifyfs.c
```diff
@@ -698,8 +687,63 @@ off_t unifyfs_fid_logical_size(int fid)
...
+            /* failed to get file size */
+            return (off_t)-1;
+        }
+        return (off_t)filesize;
+    }
+}
+
+/* scan list of files and return fid corresponding to target gfid,
+ * returns -1 if not found */
+int TMP_unifyfs_fid_from_gfid(int gfid)
```
adammoody: @MichaelBrim , this is the ugly hack that I added which I meant to ask about.

CamStan: This the one you need? https://github.com/LLNL/UnifyFS/blob/872cb269f23d8e6473ba43032b47e34d5259769b/client/src/unifyfs.c#L553

It looks awful similar to your "hack" though...

adammoody: Yes, thanks.

CamStan:
@adammoody, looks like this ran again on Travis and, as you suggested, some of the tests are failing from #473. It appears all these failing tests are from `lseek()`. All the `fseek()` tests are still passing.

CamStan:
At first glance, I think the `fseek()` tests still passed because our `fseek()` wrapper calls `unifyfs_stream_flush`
https://github.com/LLNL/UnifyFS/blob/c1b0c41c9a7436bf929b0e55aaaeb3c94fa01670/client/src/unifyfs-stdio.c#L871
which eventually gets to `unifyfs_fid_logio_write`
https://github.com/LLNL/UnifyFS/blob/c1b0c41c9a7436bf929b0e55aaaeb3c94fa01670/client/src/unifyfs-fixed.c#L405
where `add_write_meta_to_index` calls
https://github.com/LLNL/UnifyFS/blob/c1b0c41c9a7436bf929b0e55aaaeb3c94fa01670/client/src/unifyfs-fixed.c#L225

Since you just added a check to sync to `unifyfs_fid_logical_size` that got all the tests passing (which both `lseek` and `fseek` call), is it possible and/or a problem that  `unifyfs_sync` happens twice on one call to `fseek()`?

adammoody:
This PR now forces a sync as part of the stream flush.  That's a bug fix because a sync was not necessarily happening before. as the fd_write only syncs if it happens to exhaust the index buffer, which is rare.

https://github.com/LLNL/UnifyFS/blob/b98a229679d965dd5e117af5a92ff8803b1396d6/client/src/unifyfs-stdio.c#L496

@CamStan , good point about fseek calling sync twice now.  Yes, I think that's possible and it's not efficient.  We need to do a thorough sweep of our implied sync logic and associated tracking of that state.  That's been added in spots as needed so far rather than through careful inspection.

PR #441, opened 2020-01-21
Add missing ax_phreads.m4

boehms:
### Description

Commit 6a4f566ba04e0288e5a6a9c263553eb8ff4fef43 adds a call to AX_PTHREADS to [configure.ac](https://github.com/LLNL/UnifyFS/blob/6a4f566ba04e0288e5a6a9c263553eb8ff4fef43/configure.ac#L199).
This adds the corresponding m4 file.

### Types of changes

### Checklist:

### DIFF
--- a/m4/ax_pthread.m4
+++ b/m4/ax_pthread.m4
@@ -1,0 +1,492 @@
+# ===========================================================================
+#        https://www.gnu.org/software/autoconf-archive/ax_pthread.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_PTHREAD([ACTION-IF-FOUND[, ACTION-IF-NOT-FOUND]])
+#
+# DESCRIPTION
+#
+#   This macro figures out how to build C programs using POSIX threads. It
+#   sets the PTHREAD_LIBS output variable to the threads library and linker
+#   flags, and the PTHREAD_CFLAGS output variable to any special C compiler
+#   flags that are needed. (The user can also force certain compiler
+#   flags/libs to be tested by setting these environment variables.)
+#
+#   Also sets PTHREAD_CC to any special C compiler that is needed for
+#   multi-threaded programs (defaults to the value of CC otherwise). (This
+#   is necessary on AIX to use the special cc_r compiler alias.)
+#
+#   NOTE: You are assumed to not only compile your program with these flags,
+#   but also to link with them as well. For example, you might link with
+#   $PTHREAD_CC $CFLAGS $PTHREAD_CFLAGS $LDFLAGS ... $PTHREAD_LIBS $LIBS
+#
+#   If you are only building threaded programs, you may wish to use these
+#   variables in your default LIBS, CFLAGS, and CC:
+#
+#     LIBS="$PTHREAD_LIBS $LIBS"
+#     CFLAGS="$CFLAGS $PTHREAD_CFLAGS"
+#     CC="$PTHREAD_CC"
+#
+#   In addition, if the PTHREAD_CREATE_JOINABLE thread-attribute constant
+#   has a nonstandard name, this macro defines PTHREAD_CREATE_JOINABLE to
+#   that name (e.g. PTHREAD_CREATE_UNDETACHED on AIX).
+#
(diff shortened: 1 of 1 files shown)


tonyhutter:
Many distos already provide `ax_pthreads.m4` as part of their `autoconf-archive` package.  Also this file is GPL'd, which may or may not be an issue.

boehms:
I ran into issues on Summit, and I am not sure if we can assume it will be available on all systems.
And the license for the autotools-archive is not GPL, so we should be fine on that end.

tonyhutter:
@boehms I don't understand.  We're not distributing an `autoconf-archive` package with Unify.  `ax_pthreads.m4` is GPL'd, and that's what we would be distributing.  Again, I don't know if this is actually a problem or not, just wanted to bring it up.

If we're worried about the GPL, one possible alternative would be to try to use ax_pthreads first, and when that doesn't work/exist, have the Makefile fall back to manually passing `-lpthreads`.  My guess is that most Linux machines are going to accept it (hopefully Summit included).

boehms:
Using AX_PTHREAD adds an external dependency, that we can not test for. To make it worse, if the macro is not installed, it fails silently.
You are right about the GPL, it looks like ax_pthread.m4 is GPL (I was looking at the autoconf archhive webpage).

I am not sure how to proceed here.

Do you have an oppinion ion this @MichaelBrim @adammoody @CamStan?

tonyhutter:
> Using AX_PTHREAD adds an external dependency, that we can not test for. 

Can't we use `m4_ifdef([AX_PTHREAD], ...`?

boehms:
I suppose we could. But this would still only fail at configure time, rather than at autogen time. I am not sure what the right solution is. We might be able to check in the autogen.sh script.

MichaelBrim:
My suggestion:
```
m4_ifdef([AX_PTHREAD], [AX_PTHREAD],
  [AC_MSG_ERROR([AX_PTHREAD missing from system. You need to install ax_pthread.m4 from autoconf-archive in the m4/ directory before rerunning configure. Try 'cd m4; wget http://git.savannah.gnu.org/gitweb/?p=autoconf-archive.git;a=blob_plain;f=m4/ax_pthread.m4'])])
```

MichaelBrim:
We could also add an optional step to autogen.sh to download all the autoconf-archive macros we use into `m4/`

tonyhutter:
Or just hardcode `-lpthread` and ignore ax_pthread entirely.  I would be shocked to find a Linux disto that didn't support `-lpthread` out of the box.

MichaelBrim:
@tonyhutter The one real benefit of using AX_PTHREAD is it sets the proper build flags for various compilers.

tonyhutter:
@MichaelBrim we don't use any AX_PTHREAD build flags.  We only use ax_pthread.m4 in a hacky check (that I added, sadly...):
```
configure.ac:200:AX_PTHREAD([],[AC_MSG_ERROR([pthreads are required])])
```
And then we hardcode `-lpthread` in a couple of places:
```
t/Makefile.am:81:	-lpthread -lm -lstdc++ -lrt
server/src/Makefile.am:36:    -lpthread -lm -lstdc++ -lrt
client/src/Makefile.am:42:  -lcrypto -lrt -lpthread
client/unifyfs-config.in:15:POST_LD_FLAGS="$UNIFYFS_LIB_PATH/libunifyfs.a -lcrypto -lm -lrt -lpthread"
```
I would just remove the configure.ac check - at best it's useless and and worst it harmful.  If you had ax_pthread.m4 installed, but not actual pthreads, I think it would pass...

MichaelBrim:
If we're not properly using the flags set by AX_PTHREAD, we should get rid of it.

tonyhutter:
@MichaelBrim agreed.  @boehms can you update your patch to simply get rid of this line and verify it fixes your issue?:
```
configure.ac:200:AX_PTHREAD([],[AC_MSG_ERROR([pthreads are required])])
```

MichaelBrim:
Closing in favor of removing the use of AX_PTHREAD, implemented in PR #457

PR #446, opened 2020-01-30
miscellaneous documentation updates

MichaelBrim:
### Description

Adds three unrelated bits of documentation:
1. Describes the various build configuration options available to users.
2. Adds information about the resource manager job integration scripts.
3. Improves the documentation on short-form boolean command-line options.

### Types of changes

### Checklist:

### DIFF
--- a/docs/build-intercept.rst
+++ b/docs/build-intercept.rst
@@ -16,6 +16,51 @@ In this section, we describe how to build UnifyFS with I/O interception.
 
 .. _build-label:
 
+---------------------------------------
+UnifyFS Build Configuration Options
+---------------------------------------
+
+Fortran
+*******
+
+To enable UnifyFS use with Fortran applications, pass the ``--enable-fortran``
+option to configure. Note that only GCC Fortran (i.e., gfortran) is known to
+work with UnifyFS.
+
+GOTCHA
+******
+
+GOTCHA is the preferred method for I/O interception with UnifyFS, but it is not
+available on all platforms. If GOTCHA is not available on your target system,
+you can omit it during UnifyFS configuration by using the ``--without-gotcha``
+configure option. Without GOTCHA, static linker wrapping is required for I/O
+interception.
+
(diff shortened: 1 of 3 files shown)


docs/build-intercept.rst
```diff
@@ -16,6 +16,51 @@ In this section, we describe how to build UnifyFS with I/O interception.
...
+you can omit it during UnifyFS configuration by using the ``--without-gotcha``
+configure option. Without GOTCHA, static linker wrapping is required for I/O
+interception.
+
+PMI2/PMIx Key-Value Store
+*************************
+
+When available, UnifyFS uses the distributed key-value store capabilities
+provided by either PMI2 or PMIx. To enable this support, pass either
+the ``--enable-pmi`` or ``--enable-pmix`` option to configure. Without
```
CamStan: Clarifying question for my sake. When either of these is enabled, is the `-S`/`--shared-dir` option still required for `unifyfs start`? If so, what should the user provide for that option?

Or are these two things unrelated?

MichaelBrim: They are related. That last sentence in the description is the key point. With PMI2/PMIx, you pass `--shared-dir`, and it will only be used as a backup if the PMI calls fail for some reason. If you don't have PMI2/PMIX, then we MUST have the shared filesystem directory. In either case, the usage of `unifycr start` requires that you provide the option.

CamStan: Got it! Thank you!

adammoody:
I found that --enable-pmi was breaking on our SLURM systems, which uses the PMI2 plugin.  I think that was because UnifyFS was calling PMI2_Init, and then MDHIM was calling it a second time via MPI_Init --> PMI2_Init.  That second call to PMI2_Init throws an error in my SLURM PMI2 plugin.  If that's the case, this will be fixed once we drop MDHIM.  However, other SLURM/PMI2 users may hit a similar problem until then.

We also have PMIx in SLURM, and that may have worked, but I haven't tried.

PR #449, opened 2020-01-30
client: do not unlink superblock file on unmount

adammoody:
The client was calling shm_unlink during unmount, which deleted the superblock shared memory file so that future clients could not reattach to it.  This has been changed to not shm_unlink that file on the client.  The server now deletes that file during its cleanup logic.

Also, this modifies the client to immediately delete its read request and read reply shared memory files upon returning from the mount_rpc.  The server attaches to those segments during the mount rpc, so it's safe to delete the files at that point.  It also deletes those files in more failures cases for improved cleanup after failure.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -2510,21 +2510,6 @@ int unifyfs_mount(const char prefix[], int rank, size_t size,
         return ret;
     }
 
-    /* call client mount rpc function here
-     * to register our shared memory and files with server */
-    LOGDBG("calling mount");
-    invoke_client_mount_rpc();
-
-#if defined(UNIFYFS_USE_DOMAIN_SOCKET)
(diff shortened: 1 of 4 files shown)


adammoody:
@clmendes , I think this may fix the unmount bug you found.

adammoody:
Thanks @MichaelBrim .

clmendes:
Yes, this did solve the problem that had been observed when "unmount" was used, and the next app in the same job script was not able to access the contents of the file. After rebuilding UnifyFS today, I can now get the full file exported correctly by sysio-cp-gotcha, after having run "sysio-write-gotcha -u" ! Thanks!

adammoody:
Great.  Thanks for verifying that @clmendes .

PR #451, opened 2020-01-31
Transfer API doc

craigsteffen:
An initial attempt at a very short example of using the transfer API.  

(Hopefully this time this is the right way to do it.)

### Types of changes

### Checklist:

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -146,8 +146,9 @@ Transfer API
 ------------
 
 UnifyFS has a transfer API to move files from UnifyFS to external storage (or from external storage into UnifyFS).  The transfer functionality can be invoked by using the "transfer" application in the examples directory as show in the following example:
+
 .. code-block:: Bash
 
     $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
 
-(assuming that /unifyfs/snapshot.bin is a file you've written within the UnifyFS file space by an application..)  The transfer application source code also serves as a template for using the UnifyFS transfer API in a C program.
+(assuming that /unifyfs/file1 is a file you've written within the UnifyFS file space by an application.)  The transfer application source code also serves as a template for using the UnifyFS transfer API in a C program.
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -151,4 +151,20 @@ UnifyFS has a transfer API to move files from UnifyFS to external storage (or fr
 
     $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
 
-(assuming that /unifyfs/file1 is a file you've written within the UnifyFS file space by an application.)  The transfer application source code also serves as a template for using the UnifyFS transfer API in a C program.
+(assuming that /unifyfs/file1 is a file you've written within the UnifyFS file space by an application.)  To use the transfer API functions directly in a C program, use the following as a template: 
+
+.. code-block:: C
+
+    #include <unifyfs.h>
+    ...
+    // in an MPI context
+    if (rank_to_copy == my_rank) {
+            ret = unifyfs_transfer_file_serial(srcpath, dstpath);
+            if (ret) {
+                fprintf(stderr,"unifyfs copy failed (%d: %s)", ret, strerror(ret));
+            }
+        }
+        
+For a more complete example, please see the transfer.c source file in the examples directory.
+ 
+
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -141,3 +141,14 @@ One form of running this example could be:
 .. explicit external hyperlink targets
 
 .. _examples: https://github.com/LLNL/UnifyFS/tree/dev/examples/src
+
+Transfer API 
+------------
+
+UnifyFS has a transfer API to move files from the UnifyFS file space out to external storage (or from external storage into the UnifyFS file space).  This can be invoked using the function API by linking, or using the available "transfer" application availalbe in the examples directory.  The example code also serves as a template for using the C function API. 
+
+.. code-block:: Bash
+
+    $ srun -N4 -n4 transfer-static /unifyfs/snapshot.bin /scratch/snaps/snapshot.bin
+
+(assuming that /unifyfs/snapshot.bin is a file you've written within the UnifyFS file space by an application.)
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -145,7 +145,7 @@ One form of running this example could be:
 Transfer API
 ------------
 
-UnifyFS has a transfer API to move files from UnifyFS to external storage (or from external storage into UnifyFS).  The transfer functionality can be invoked by using the "transfer" application in the examples directory as show in the following example: 
+UnifyFS has a transfer API to move files from UnifyFS to external storage (or from external storage into UnifyFS).  The transfer functionality can be invoked by using the "transfer" application in the examples directory as show in the following example:
 .. code-block:: Bash
 
     $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -151,7 +151,7 @@ UnifyFS has a transfer API to move files from UnifyFS to external storage (or fr
 
     $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
 
-(assuming that /unifyfs/file1 is a file you've written within the UnifyFS file space by an application.)  To use the transfer API functions directly in a C program, use the following as a template: 
+(assuming that /unifyfs/file1 is a file you've written within the UnifyFS file space by an application.)  To use the transfer API functions directly in a C program, use the following as a template:
 
 .. code-block:: C
 
@@ -164,7 +164,5 @@ UnifyFS has a transfer API to move files from UnifyFS to external storage (or fr
                 fprintf(stderr,"unifyfs copy failed (%d: %s)", ret, strerror(ret));
             }
         }
-        
-For a more complete example, please see the transfer.c source file in the examples directory.
- 
 
+For a more complete example, please see the transfer.c source file in the examples directory.
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -145,7 +145,7 @@ One form of running this example could be:
 Transfer API 
 ------------
 
-UnifyFS has a transfer API to move files from the UnifyFS file space out to external storage (or from external storage into the UnifyFS file space).  This can be invoked using the function API by linking, or using the available "transfer" application availalbe in the examples directory.  The example code also serves as a template for using the C function API. 
+UnifyFS has a transfer API to move files from the UnifyFS file space out to external storage (or from external storage into the UnifyFS file space).  This can be invoked using the "transfer" application in the examples directory.  The transfer source code also serves as a template for using transfer API through C functions. 
 
 .. code-block:: Bash
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -151,18 +151,36 @@ UnifyFS has a transfer API to move files from UnifyFS to external storage (or fr
 
     $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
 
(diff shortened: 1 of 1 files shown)


docs/examples.rst
```diff
@@ -141,3 +141,14 @@ One form of running this example could be:
 .. explicit external hyperlink targets
 
 .. _examples: https://github.com/LLNL/UnifyFS/tree/dev/examples/src
+
+Transfer API 
+------------
+
+UnifyFS has a transfer API to move files from the UnifyFS file space out to external storage (or from external storage into the UnifyFS file space).  This can be invoked using the function API by linking, or using the available "transfer" application availalbe in the examples directory.  The example code also serves as a template for using the C function API.
```
MichaelBrim: typo s/availalbe/available/

Also, Travis is complaining about the trailing whitespace on lines 145 an 148

MichaelBrim: also, you have redundant "available"s, get rid of the one before "transfer"

craigsteffen:
Ok, I cleaned up the text.  

I also did my best within the git editor to remove all the trailing white space.  If there's still more I may need to open it up in emacs on a system to clean that out.

craigsteffen:
Ok, it looks like it's still failing the whitespace checks.  I need to leave the house in about 2 hours.  I'll try to download it and remove the white space before I go, but I have a scout event at a hockey game tonight so I have a hard stop.  

The next time I have a bit of time I'm going to remove the trailing white space check in the checkpatch.  All compilers (except fortran pre-F90) ignore it.  All web tools ignore it, and in fact I don't seem to even be ABLE to remove it in the GUI tool.  While I admit some of persnickety style checks have a purpose, even if obscure, I cannot think of a real reason we have to stick with this one.

MichaelBrim:
@craigsteffen if you use vim, you can easily find trailing whitespace with the following search pattern
```
:set hlsearch
/\s\+$
```

adammoody:
Here is a handy one-liner command that should remove any trailing white space in a file:
```
sed -i 's/[[:space:]]*$//' file.c
```
I haven't tried this myself, but I've used similar commands in the past.

craigsteffen:
Yes, thank you.  If I take the extra step of pulling the commit onto a
machine with emacs, I can also easily find and remove the trailing white
space.

Finding it isn't the issue.  The issue is that having to edit it in emacs
or whatever adds several unnecessary steps just to please the syntax
checker, for no real reason that I'm aware.  So for the sake of being able
to edit documentation without having to pull every single solitary change
into a separate editor, I'm going to propose that we remove the trailing
white space check.  I will open that in a separate issue and we can talk
about it in Houston or the week when we get back.

Craig

docs/examples.rst
```diff
@@ -141,3 +141,14 @@ One form of running this example could be:
 .. explicit external hyperlink targets
 
 .. _examples: https://github.com/LLNL/UnifyFS/tree/dev/examples/src
+
+Transfer API
+------------
+
+UnifyFS has a transfer API to move files from the UnifyFS file space out to external storage (or from external storage into the UnifyFS file space).  This can be invoked using the "transfer" application in the examples directory.  The transfer source code also serves as a template for using transfer API through C functions.
```
kathrynmohror: Suggest: "... to move files from UnifyFS to external storage (or from external storage into UnifyFS)." I am not sure if the terminology "file space" will confuse people

kathrynmohror: "This can be invoked using the "transfer" application in the examples directory." Suggest to change this to: "The transfer functionality can be invoked by using the "transfer" application in the examples directory as show in the following example: " <then put the srun command srun -N4 -n4...>

Then follow with "The transfer application source code also serves as a template for using the UnifyFS transfer API in a C program."

docs/examples.rst
```diff
@@ -141,3 +141,14 @@ One form of running this example could be:
...
+
+.. code-block:: Bash
+
+    $ srun -N4 -n4 transfer-static /unifyfs/snapshot.bin /scratch/snaps/snapshot.bin
+
+(assuming that /unifyfs/snapshot.bin is a file you've written within the UnifyFS file space by an application.)
```
kathrynmohror: Perhaps a simpler example might be:
"srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1

where file1 is a file that was previously written and laminated in UnifyFS."

kathrynmohror: What does the C API look like? Can we put the prototype here similarly to what is done for mount/unmount? https://unifyfs.readthedocs.io/en/latest/api-mount.html

docs/examples.rst
craigsteffen: No objections to these changes.  

I think at some point we do need terminology in the documentation that refers to files that exist inside Unify vs. out on conventional disk.  I don't think that's clear at all in the initial pages of the documentation introduction.  But yes, let's definitely try not to confuse things here.

docs/examples.rst
craigsteffen: > Perhaps a simpler example might be:
> "srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
> 
> where file1 is a file that was previously written and laminated in UnifyFS."

Sounds fine to me.

craigsteffen: > What does the C API look like? Can we put the prototype here similarly to what is done for mount/unmount? https://unifyfs.readthedocs.io/en/latest/api-mount.html

Certainly, and I will do that.  With the ECP meeting imminent and my having a commitment on Friday late afternoon, I put into the docs the shortest thing I was confident would be correct and informative.

docs/examples.rst
```diff
@@ -141,3 +141,13 @@ One form of running this example could be:
...
+
+Transfer API
+------------
+
+UnifyFS has a transfer API to move files from UnifyFS to external storage (or from external storage into UnifyFS).  The transfer functionality can be invoked by using the "transfer" application in the examples directory as show in the following example: 
+.. code-block:: Bash
+
+    $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
+
+(assuming that /unifyfs/snapshot.bin is a file you've written within the UnifyFS file space by an application..)  The transfer application source code also serves as a template for using the UnifyFS transfer API in a C program.
```
kathrynmohror: Please change "snapshot.bin" to "file1" (missed a spot!)

docs/examples.rst
kathrynmohror: Okay, if you don't have time to document the API, can you put a reference to the header file where the API is defined perhaps? E.g. For more information on the transfer API please refer to client/src/unifyfs.h .

craigsteffen: I have time to today.  I just didn't yesterday afternoon.  I'll work on that as soon as I get the other changes to both work and satisfy the syntax checker.

craigsteffen:
Ok--I think this is reasonable at this point.  

I'm confused as to the result of the test builds here.  One ("continuous-integration/travis-ci/pr") says it failed, and has references to old versions of the source.  The second ("continuous-integration/travis-ci/push") succeeded, and the result says "1 failing and 1 successful checks".  I don't understand what that means.  

As a side note, I have a PR ready to go that will adjust the syntax checker scripts so that it won't take half a dozen commits to do a minor bit of text editing in the documentation.  Sorry about all the commit spam.

craigsteffen:
Ok, so I think the source file is now in good shape.  I think all the file names agree, i think it contains a reasonable level of example.  

This contains 9 changes.  On one hand, I seem to think that people generally wanted PR's to be rebased, and combining 9 commits into 1 or 2 would certainly make sense, and I'm happy to do it.  On the other hand, since we've been firing comments back and forth, and I don't know how having the various commits interact with githubs comment system, I don't want to rebase this and then force push it and blow blow up something in the github history or GUI or comment system.  
So please let me know if this branch/PR should be rebased, and I'll do it.

CamStan:
> One ("continuous-integration/travis-ci/pr") says it failed, and has references to old versions of the source. The second ("continuous-integration/travis-ci/push") succeeded, and the result says "1 failing and 1 successful checks". I don't understand what that means.

@craigsteffen, Travis is set up to build on pull requests and new branches, however checkpatch only runs when there are new files or a diff between files. 

Since you pushed this branch to the LLNL/UnifyFS repo, the passing check ("continuous-integration/travis-ci/push") is Travis simply building your branch (this check won't run if doing a PR from a forked repo). A branch by itself has nothing to generate a diff from, so checkpatch doesn't see anything that changed and won't catch anything here.

The failing check ("continuous-integration/travis-ci/pr") is Travis testing the build of what the resulting merge would be between your new branch and dev. In this case, there is a diff between files and checkpatch can see what changed. Here it is being the whitespace that wasn't there before and is what is causing the failed check.

docs/examples.rst
```diff
@@ -141,3 +141,28 @@ One form of running this example could be:
...
+    $ srun -N4 -n4 transfer-static /unifyfs/file1 /scratch/mydir/file1
+
+(assuming that /unifyfs/file1 is a file you've written within the UnifyFS file space by an application.)  To use the transfer API functions directly in a C program, use the following as a template:
+
+.. code-block:: C
+
+    #include <unifyfs.h>
+    ...
+    // in an MPI context
+    if (rank_to_copy == my_rank) {
```
adammoody: If the user is running an MPI job, I think we should recommend the parallel copy API rather than the serial.  For a large file, this should make a big difference in performance, assuming we don't manage to thrash our servers.

docs/examples.rst
```diff
@@ -141,3 +141,28 @@ One form of running this example could be:
...
+            ret = unifyfs_transfer_file_serial(srcpath, dstpath);
+            if (ret) {
+                fprintf(stderr,"unifyfs copy failed (%d: %s)", ret, strerror(ret));
+            }
+        }
```
adammoody: If we keep this if statement, are these lines are indented one too many levels or at least the closing brace for the "if"?

craigsteffen:
Cameron,

Sorry I didn't fork first.  I realized my mistake late,  and since this PR
had already generated way too many pushes abs comments, I wanted to ask
what to do before proceeding.  Please let me know how to proceed in such a
way that minimizes how much anyone else has to do.

Good point about the parallel transfer.  I will make that change and also
check the indentation levels.

Craig

craigsteffen:
Cameron,

After reading your message I think I finally understand what you're
saying.  My confusion is that the check seems to be doing the diff between
an earlier commit in the branch that I created than the latest one.  The
latest one has all the extra space removed, I'm pretty sure.  I'll probably
double-check by checking out that branch and checking the files with a text
editor.

Someone brought up the idea of just rebasing that branch down to a single
commit, force push it, and then the PR can be against that.  If that's what
I should do, I'm happy to do that.  Someone I lost that message so I'm
mentioning it here.

Craig

CamStan:
Craig, don't worry about not forking first. We haven't enforced that with this project, several people just prefer to work that way. It can keep the clutter of additional branches down, but they can also just be deleted after a PR is merged. Just know that the "push" check will run on your new branch if you push to LLNL/UnifyFS repo, which is not a problem. If that check fails, it typically means the build itself failed or some unit tests failed. The build and unit tests do also get run in the "pr" check.

As far as rebasing down to a single commit, this is also something that varies from project to project. Most of us do that for this project, perhaps even in cases where we shouldn't. I believe our practice has typically been to have your PR be one commit, unless your PR has multiple logical pieces that might best be viewed in separate commits. This is typically only the case with bigger PRs, which arguably could even be broken into multiple PRs in some cases.

tonyhutter:
A couple of thoughts:

- If you just need a copy util, why not pull in and build `rsync` or `cp` against Unify via the bootstrap.sh script?  They will have way more options than we could ever hope to implement (like rsync compression) and users would already be familiar with them.  There are ways to do multithreaded transfers with them as well (just google things like "rsync multiple threads").

- Do we need a custom transfer API?  Why not simply wrap the existing POSIX file transfer APIs like `splice()` or `copy_file_range()`, or the `FICLONE ioctl()` (which vanilla `cp` uses internally)?

- If we do need a transfer API, why not pull in AXL (https://github.com/ECP-VeloC/AXL) instead?  A number of us are contributors to that project.  It already defines a transfer API, a copy program (`axl_cp`), and supports synchronous and multithreaded transfers.

craigsteffen:
I'm going to re-create this in a new branch, and generate a new PR.

PR #455, opened 2020-02-02
trailing whitespace check small but nonzero

craigsteffen:
This patch adds "repeat" tags to the regex in the trailing white space 
check.  Up to 10 whitespace characters are allows on the end of a line 
with non-white space characters, and up to 15 white space characters 
are allowed on a line by themselves.  

However, it won't allow huge lines, or multiline white space lines. 
 It still flags those.  

This change will make editing the documentation via the 
github gui much much easier and more efficient and 
eliminate a lot of cleanup to the source files.  

### Motivation and Context

This is enough to allow people who use emacs to hit "tab" all the 
time (which either adds 4 or 6 characters per line or one tab, 
depending on configuration) to in most cases still have their 
ource code get by the syntax checker without being flagged.  It 
also allows edits from the github GUI editor (which inserts a 
space here and there when manipulating files) to go in without 
having to separately, externally, scrub out the end line spaces.  

### How Has This Been Tested?

I tested editing a source file, looked at the "git diff" and also 
the "git diff | scripts/checkpatch.sh".  As I added a few spaces, 
they showed up in the git diff but not in the checkpatch until 
they hit the desired threshold, at which point the complaint 
was "excessive trailing whitespace".  

I tested this by changing a documentation file with the github gui 
editor in the branch for this change.  The source file was 
successfully validated, only because the checkpatch allows 
the couple of trailing spaces that the gui generates.  I then 
removed that test commit from this branch.

### Types of changes

### Checklist:

### DIFF
--- a/scripts/linux_kernel_checkpatch/checkpatch.pl
+++ b/scripts/linux_kernel_checkpatch/checkpatch.pl
@@ -2747,10 +2747,10 @@ sub process {
 			    $fix) {
 				$fixed[$fixlinenr] =~ s/[\s\015]+$//;
 			}
-		} elsif ($rawline =~ /^\+.*\S\s+$/ || $rawline =~ /^\+\s+$/) {
+		} elsif ($rawline =~ /^\+.*\S\s{10,}$/ || $rawline =~ /^\+\s{15,}$/) {
 			my $herevet = "$here\n" . cat_vet($rawline) . "\n";
 			if (ERROR("TRAILING_WHITESPACE",
-				  "trailing whitespace\n" . $herevet) &&
+				  "excessive trailing whitespace\n" . $herevet) &&
 			    $fix) {
 				$fixed[$fixlinenr] =~ s/\s+$//;
 			}


craigsteffen:
Since this is effectively a (very small) change to the source coding standards, I presume we'll have a discussion about this one.  However, I believe this is rebased correctly and is ready to go in if and when it's Ok-ed.

MichaelBrim:
marked WIP until we make an informed decision

craigsteffen:
Let's open this up to discussion again.  

Premise 1: I fought with trailing white space complaints from the check scripts again with a recent commit.  I believe it adds a lot of hassle to contributing to the project, especially when editing documentation within the github gui itself.  Cleaning up all the trailing whitespace adds time and steps to the process. 

Premise 2: Furthermore, those steps aren't necessary.  All compilers (except perhaps pre-1990 Fortran) ignore trailing white space on lines.  All the documentation web engines ignore it.  Most editors don't display it.  

From my experience the trailing white space checks add a lot of overhead to developing.  I believe there is absolutely no benefit to having it.  I would like to open a discussion to remove that check, or else soften it (per my attached pull request) that would allow a few spaces in a blank line or a few trailing spaces on a line but still police long blank lines or long trailing space blocks.

craigsteffen:
I realize the last commit to that branch was in February.  However, I rebased it against the current dev, so it's not so out of date.  I also tested the modifications again after rebasing it.  It behaved correctly; it complained about "excessing whitespace" after the number of trailing spaces exceeded the threshold, but didn't when the number of spaces was below.

tonyhutter:
Prohibiting trailing whitespace is standard code practice.  It makes the code easier for others to edit, and just looks more professional.  It also makes diffs cleaner.  Imagine if you accidently add trailing whitespace to a line and commit it.  Then, if someone else does a `git blame` on that line to see where it was added, it annoyingly shows the commit where the whitespace was added rather then the relevant one where the code was actually added.  

It's trivial to autoremove trailing whitespace from files.  I use:
``` 
sed -i 's/[ \t]*$//g' [your files]
```
My vote is no on this change.

MichaelBrim:
I'm still in the "keep it as it is now" camp. I agree with @tonyhutter that it is easy to remove the trailing whitespace, when you use the right tools for the job. I have only used the github gui once in my life, for a minor typo fix. It is not the right tool for doing UnifyFS development or document editing. If you want to use a GUI IDE, download Visual Studio Code, it's wonderful and has great git integration.

craigsteffen:
The github gui works.  It is a perfectly reasonable workflow.  It has the
added advantage that it has a "preview" button right there to make sure
that it renders properly.  Yes, I could download Visual Studio or whatever,
but I *could* also keep using the tool that's available to me now, except
that the completely arbitrary trailing space check is making that workflow
annoying.

Again, yes, if I was editing the source directly, it would be fairly easy
(although still unnecessarily annoying) to invoke another script to strip
out the trailing whitespace.  But when editing in the github gui (which
works just fine, other than this) there isn't a hook as far as I know to do
post-processing scripting after you safe the file but before the checks are
done.

"It makes the code easier for others to edit, and just looks more
professional. "  What editor or workflow shows white space at the end of
lines?  Emacs and vi don't, as far as I know.  What editing interface or
compiler does someone actually use that shows trailing white space?  How
would someone ever even know?

I get the hashing file issue, but that seems like a tertiary concern at
worst, and mostly won't effect anything real.  Interfering with efficient
editing of documentation is something that has been messing up my workflow
of documentation, significantly, in the past three days.

Craig Steffen

tonyhutter:
> "It makes the code easier for others to edit, and just looks more
> professional. "  What editor or workflow shows white space at the end of
> lines?  Emacs and vi don't, as far as I know.  What editing interface or
> compiler does someone actually use that shows trailing white space?  How
> would someone ever even know?
> 
When I hit '$' to go to the end of a line in vi, I want it to go to the end of the line of code.  I don't want it to go to the end of the line of code + unnecessary whitespace.

craigsteffen:
The point here is that as far as I can tell, when editing documentation,
the github GUI editor sometimes introduces trailing whitespace characters
that trip the trailing whitespace check when it's run on the source file.
There is no way, as far as I know, to get the gui to not do this, and
requiring the check script to remove the spaces would then mess up the
hashing function of git, thus making the process even more complicated.

Since the read-the-docs web page has a "edit this" button, and the gui
editor is fast and efficient and has a preview button, I find it terribly
quick and easy to use, and will make editing documentation for Unify very
efficient for me.

However, the trailing whitespace is currently forcing me to them pull the
files onto a machine, like catalyst, where I can run scripts on the source
to "fix" the files that the GUI editor produced.  So I'm asking for that
one check to be relaxed so I can use the workflow that's easy to use.

If the group votes against this PR, we're saying as a group that the github
GUI is not a valid documentation editing method because it's less important
than having a trailing intact whitespace check.

I think preserving a syntax check that is largely invisible and only
matters for hashes and end-of-line editing in VI is not worth cutting out a
whole workflow path (especially a built-in efficient one).

Craig

MichaelBrim:
@craigsteffen I have been trying to reproduce the behavior you are seeing when using the edit functionality on the github web interface. When I add new lines, I don't end up with trailing whitespace. Are you by chance doing this on a Windows machine? I'm wondering if the extra whitespace is a Windows CRLF issue.

craigsteffen:
When I first ran into this a month or two ago, I might have done some edits
on a Windows machine.  However, not my recent round; my Windows laptop die
a couple/three weeks ago and I haven't recplaced it yet.  The only machine
I've done the edits is my work desktop which is a Mac Mini, running within
Chrome for the past several weeks.

I will do some edits and try to reproduce the problem on a separate
branch.

Craig

MichaelBrim:
My recent tests are on a Macbook in Chrome, so they should be comparable to your current experience.

kathrynmohror:
Perhaps a compromise would be to exclude the documentation files from being checked by checkpatch.sh? That way the docs could be edited from the GUI easily, but all the code would still be checked under the same rules they are now.

adammoody:
A test with Windows and Firefox using the github editor is not adding trailing spaces for me.  @craigsteffen , let us know if you can still reproduce this.  Maybe it was a bug that github has since fixed?

CamStan:
I realize I'm a bit late to the discussion, but to me, whitespace at the end of a line IS a bug.

Possibly an obscure edge case, but I've used shell scripts (also occurs in Python and Javascript to my knowledge) to setup and run our CI tests and within these I provide helper lines/output via here documents to assist other users running these that are less familiar with them. (example [here](https://github.com/LLNL/UnifyFS/blob/599a2429ef96b5da0f6b1e02543eb6adc585a232/t/ci/001-setup.sh#L30-L67))

Essentially, this is a bug when it comes to multi-line string output.
For example:
```bash
echo "Hello \
World"
```
has no whitespace after the line break and results in
```
Hello World
```
However, with a whitespace after the line break, this results in:
```bash
Hello \
World
```

In my case, if I add a whitespace at the end of my here document for help output:
```bash
#!/bin/sh

USAGE_TEXT="$(cat <<EOF
usage: ./sample.sh -h|--help)

Usage text here.
This is just an example.
EOF 
)"

while [[ $# -gt 0 ]]
do
    case $1 in
        -h|--help)
            echo "$USAGE_TEXT"
            exit
            ;;
        *)
            echo "usage: ./sample.sh -h|--help"
            exit 1
            ;;
    esac
done
```
This fails to prints the help message and instead results in:
```
./sample.sh: line 15: warning: here-document at line 9 delimited by end-of-file (wanted `EOF')
```
This kind of bug takes me a long time to track down if I'm not being warned about whitespace.

---

> What editor or workflow shows white space at the end of lines?  Emacs and vi don't, as far as I know.  What editing interface or compiler does someone actually use that shows trailing white space?

Personally, I have "show whitespace" turned on in my vim editor. I know emacs can not only do this, but can also be set up itself to automatically remove whitespace for you. I agree that whitespace breaks how my editor should function as the key commands no longer take me where I expect them to. 

---

> But when editing in the github gui there isn't a hook as far as I know to do post-processing scripting after you safe the file but before the checks are done.

Correct, GitHub doesn't allow for pre-commit hooks, but git itself has this. However, this again requires you to be editing in your own environment (I was always advised never to use the GitHub GUI because of a variety of issues it can cause), but then git can be set up to remove whitespace for you automagically when you commit. 

---

I mostly agree with Kathryn's compromise, however I would still like the whitespace check for when I'm writing documentation locally. As a potential alternative, we could add something like the following to our .travis.yml to keep checkpatch from checking docs only when Travis runs the check, but running checkpatch on our own systems would still flag them.
```bash
DOCS_DIR=$(echo $(ls docs/*.rst))
export TEST_CHECKPATCH_SKIP_FILES=${DOCS_DIR//[[:blank:]]/,}
```

kathrynmohror:
Thanks @CamStan! That was a really helpful analysis.

craigsteffen:
Ok.  It seems that everyone wants to keep the trailing whitespace checking
strict except me.

Operationally I would like a way to be able to edit the Unify documentation
and be able to validate it and ideally see it rendered without having to
create a new commit every time I do so.  I will do some digging before the
call this afternoon.  If anyone knows of a command-line validator that
would likely be available on lassen/summitdev and the like, that would be
very useful.  I will look for one myself.  I notice that github seems to
have an installable Mac OS "Github desktop".  I will try that now to see if
that does better with saving source files in a reasonable way.

Craig

MichaelBrim:
I would just note that Visual Studio Code (free for use, no Visual Studio required) also allows you to do preview without committing. It really is a great IDE if you haven't tried it.

regards,
mike
________________________________

CamStan:
> I would just note that Visual Studio Code (free for use, no Visual Studio required) also allows you to do preview without committing. It really is a great IDE if you haven't tried it.

This is available on our x86 systems. I still need to play with it.

MichaelBrim:
@craigsteffen @CamStan Craig, can we close this now and go ahead with Cameron's suggestion to tell checkpatch to skip the docs in Travis checks?

craigsteffen:
Yep.  I meant to close this.  I'll do it now.

craigsteffen:
Visual Studio Code definitely works without the occasional trailing white space in the source files.  So closing this for good.

PR #460, opened 2020-02-10
move unifyfs_generate_gfid() to common

MichaelBrim:
### Description
Includes five important changes:
1. Fixes a bug in the hash computation, where abs() was being called on the hash value to enforce a positive integer. This had the potential to result in conflicts between the positive and
negative values of any number (i.e., x and -x). Now, we treat the hash value as unsigned, and just cast to signed, which doesn't change its value.
2. Uses a uint64_t for the hash value, and converts the most significant 32 bits into an integer for use as gfid.
3. Implements a method of hash value construction that will result in the same values on machines with different endianness.
4. Fixes a bug in MDHIM where all extent metadata was being sent to server rank 0.
5. For IBM LSF-CSM systems, have the `unifyfs` command-line tool capture server stdout/err into files. This is helpful since all MDHIM errors get printed to stderr.

### Types of changes

### Checklist:

### DIFF
--- a/common/src/unifyfs_meta.c
+++ b/common/src/unifyfs_meta.c
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2020, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2020, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyFS.
+ * For details, see https://github.com/LLNL/UnifyFS.
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
+
+#include <endian.h>
+#include <string.h>
+#include <openssl/md5.h>
+
+#include "unifyfs_meta.h"
+
+/**
+ * Hash a file path to a 64-bit unsigned integer using MD5
+ * @param path absolute file path
+ * @return hash value
+ */
+uint64_t compute_path_md5(const char* path)
+{
+    unsigned long len;
+    unsigned char digested[16] = {0};
+
+    len = strlen(path);
+    MD5((const unsigned char*) path, len, digested);
+
+    /* construct uint64_t hash from first 8 digest bytes */
+    uint64_t hash = be64toh(*((uint64_t*)digested));
+    return hash;
+}
--- a/common/src/unifyfs_meta.h
+++ b/common/src/unifyfs_meta.h
@@ -27,20 +27,6 @@
 extern "C" {
 #endif
 
-/**
- * Server commands
- */
-typedef enum {
-    COMM_MOUNT,
-    COMM_META_FSYNC,
-    COMM_META_GET,
-    COMM_META_SET,
-    COMM_READ,
-    COMM_UNMOUNT,
-    COMM_DIGEST,
-    COMM_SYNC_DEL,
-} cmd_lst_t;
-
 typedef struct {
     int gfid;
     char filename[UNIFYFS_MAX_FILENAME];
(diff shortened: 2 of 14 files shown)


common/src/unifyfs_meta.h
```diff
@@ -102,6 +103,34 @@ typedef struct {
...
+ */
+static inline
+unsigned int compute_path_md5(const char* path)
+{
+    unsigned char digested[16] = {0};
+    unsigned long len = strlen(path);
+
+    MD5((const unsigned char*) path, len, digested);
+
+    unsigned int* ival = (unsigned int*) digested;
```
tonyhutter: Technically we should do:
```C
unsigned int ival = (digested[0] << 24) | (digested[1] << 16) | (digested[2] << 8) | digested[3];
```
That way, we calculate the same gfid on both little and big endian nodes.  However, if we're currently storing the gfids in the server's metadata, doing this change may corrupt all little endian Unify filesystems out in the wild...

And on a totally unrelated note, we should consider changing the gfid to a int64_t (so it has a larger hash space, and is signed so we can return errors).

Both these issues are outside the scope of this PR though.  Just thought I'd mention them since it jumped out at me while reviewing this code.

tonyhutter: I'll open a separate issue for this

adammoody: @tonyhutter , better to break things now than later, when the number of active Unify users is small.

Another item to include in the new issue is that we should normalize the path before we hash so that ```/unifyfs/./foo``` is the same as ```/unifyfs/foo```.  Currently, we hash those to two different integers.  Also, we may want to allow for ```chdir("/unifyfs")``` and use of relative paths.

Hashing on the full path makes a ton of sense when we thought support for ```rename()``` would not be needed.  If we do support rename calls, we might also want to think about how we're hashing.  Currently a rename of a top-level directory will require us to rehash every item it its subtree.  That may be fine, but if that's a problem, we'll need to adjust.

tonyhutter: Issue opened for endianness:
https://github.com/LLNL/UnifyFS/issues/461

MichaelBrim: Decided on today's concall to support 64-bit hash value and Tony's endianess fix from #461 .

MichaelBrim:
Looks like the "gfid isn't negative" assumption is being broken by this change, and it's showing up in the unit tests. I'll need to work on this some more.

MichaelBrim:
This should be good to go now.

adammoody:
Looks good to me.  @tonyhutter , can you also review this again?

adammoody:
Having all extents mapped to a single server would explain why our fsync() cost has been scaling linearly with the number of extents.

common/src/unifyfs_meta.c
```diff
@@ -0,0 +1,41 @@
...
+    len = strlen(path);
+    MD5((const unsigned char*) path, len, digested);
+
+    /* construct uint64_t hash from first 8 digest bytes */
+    uint64_t hash = 0;
+    for (int i = 0; i < 8; i++) {
+        int shift_bits = 8 * (7 - i);
+        uint64_t byteval = (uint64_t)(digested[i]);
+        hash |= (byteval << shift_bits);
+    }
```
tonyhutter:
```C
#include <endian.h>
uint64_t hash = be64toh( * ((uint64_t *) digested));
```

meta/src/partitioner.c
```diff
@@ -85,13 +88,11 @@ long double get_str_num(void *key, uint32_t key_len) {
...
 }
 
-unsigned long * get_meta_pair(void *key, uint32_t key_len) {
-	ulong *meta_pair = (ulong *)malloc (2*sizeof(ulong));
-	memset(meta_pair, 0, 2*sizeof(unsigned long));
-	meta_pair[0] = *((unsigned long *)(((char *)key)));
-	meta_pair[1] = *((unsigned long *)(((char *)key)+sizeof(unsigned long)));
-	//	printf("meta_pair[1] is %ld\n", meta_pair[1]);
-	return meta_pair;
+void* copy_unifyfs_key(void* key, uint32_t key_len)
```
tonyhutter: Add comment above `copy_unifyfs_key()`:
```
/* Allocate a copy of a key and return it.  The returned key must be freed. */
```

common/src/unifyfs_meta.c
MichaelBrim: @tonyhutter Thanks for the reference - hadn't encountered those conversion functions before. Can we safely assume endian.h is available on our target platforms, or should I add a configure check? I found it was added back in glibc 2.9, which seems pretty old, so I'm guessing it's safe to not check.

meta/src/partitioner.c
MichaelBrim: done

common/src/unifyfs_meta.c
tonyhutter: It's safe to skip the check - glibc 2.9 is 2008-2009 era.

MichaelBrim: done

PR #471, opened 2020-03-13
implement and use common log-based I/O methods

MichaelBrim:
### Description

This implements common methods for client and server access to log-based I/O. A new `logio_context` structure contains all state for managing both the shared memory and spill file portions. The shared memory for data chunks is no longer allocated out of the client superblock, but instead uses a dedicated shmem region. Chunk management is now handled by a new `slot_map` structure that maintains a bitmap of free/used chunk slots. The chunk slotmap
resides within the first page of the shmem region and spill file.

This also includes the following code cleanup:
* remove unused constants
* remove unused read request shmem region
* remove unused client socket state
* libcommon needs to link against libpthread
* compiler warning fixes

NOTE: The configuration settings (and associated environment variables) used to control memory and spill usage for data chunks have changed. The test environment scripts have been updated to use the new env vars. See changes to `common/src/unifyfs_configurator.h` for details.

### How Has This Been Tested?

Tested using existing examples in multiple modes on Summitdev. The new `slot_map` functionality also has its own unit test.

### Types of changes

### Checklist:


common/src/unifyfs_configurator.h
```diff
@@ -72,10 +72,15 @@
...
+    UNIFYFS_CFG(client, write_index_size, INT, UNIFYFS_INDEX_BUF_SIZE, "write metadata index buffer size", NULL) \
     UNIFYFS_CFG_CLI(log, verbosity, INT, 0, "log verbosity level", NULL, 'v', "specify logging verbosity level") \
     UNIFYFS_CFG_CLI(log, file, STRING, unifyfsd.log, "log file name", NULL, 'l', "specify log file name") \
     UNIFYFS_CFG_CLI(log, dir, STRING, LOGDIR, "log file directory", configurator_directory_check, 'L', "specify full path to directory to contain log file") \
-    UNIFYFS_CFG(logfs, index_buf_size, INT, UNIFYFS_INDEX_BUF_SIZE, "log file system index buffer size", NULL) \
+    UNIFYFS_CFG(logio, chunk_size, INT, UNIFYFS_LOGIO_CHUNK_SIZE, "log-based I/O data chunk size", NULL) \
+    UNIFYFS_CFG(logio, shmem_size, INT, UNIFYFS_LOGIO_SHMEM_SIZE, "log-based I/O shared memory region size", NULL) \
+    UNIFYFS_CFG(logio, spill_size, INT, UNIFYFS_LOGIO_SPILL_SIZE, "log-based I/O spillover file size", NULL) \
+    UNIFYFS_CFG(logio, spill_dir, STRING, NULLSTRING, "spillover directory", configurator_directory_check) \
```
CamStan: Could you update https://github.com/LLNL/UnifyFS/blob/dev/docs/configuration.rst to reflect these changes as well?

adammoody:
Really nice work, @MichaelBrim .  Thanks!

I've skimmed this and it looks great to me based on that.  I likely won't have time for a really in depth review.  A couple questions that are easier to answer here:
1) With the previous code, we could configure things so that we could set the memory and spillover log space to each be 0 bytes (i.e., not used).  I think that's still possible here, right?
2) Can a client MPI rank reattach to its shared memory region and spill over files, say if it were to fail and then be restarted in a new job?  Though we're not relying on this right now, this would be useful for checkpoint/restart assuming that we also continue to allow the client to directly access areas for its data storage.

common/src/unifyfs_configurator.h
MichaelBrim: @CamStan Yes, forgot to update that doc.

MichaelBrim:
@adammoody 
1. Yes, UNIFYFS_LOGIO_SHMEM_SIZE sets the shared memory size, and UNIFYFS_LOGIO_SPILL_SIZE sets the spillfile size. I removed UNIFYFS_USE_SPILLOVER since it is implied by setting the spillfile size to non-zero.
2. I haven't verified that client re-attach works, but it is coded to allow that. Some more client management work still yet to come will make it easier to test this.

common/src/unifyfs_configurator.h
MichaelBrim: pushed a new version that updates the configuration doc page

adammoody:
Thanks @MichaelBrim !  That's a ton of nice work.

Issue #470, opened 2020-03-09
Add support for read-after-write w/o explicit sync


adammoody:
In order to support a program that does the following, we need to track whether any outstanding writes need to be flushed to the server and then implicitly sync before we execute the read operation:
```
int fd = open("file.txt", O_RDWR);
pwrite(fd, buf, 10, 0);
pread(fd, buf, 10, 0);
```
I believe we have a flag to track whether the file needs to be sync'd already.  We need to verify that we're setting this bit on/off appropriately everywhere.  Then we need to refer to it on read operations.

With our current implementation, we require an fsync in there to get the expected data on read.
```
int fd = open("file.txt", O_RDWR);
pwrite(fd, buf, 10, 0);
fsync(fd);
pread(fd, buf, 10, 0);
```

adammoody:
Thinking about this more, there are other read-like operations that might similarly have to imply a sync when the file has outstanding writes, like fstat() and lseek().

adammoody:
Here, we already have truncate sync writes, unconditionally:
https://github.com/LLNL/UnifyFS/blob/be26e612cd5b5dcecc4f92f495970535605aaf3e/client/src/unifyfs-sysio.c#L241

And here is an example of where we set the "needs_sync" flag to track that the file has a writes that need to be synced with the server:
https://github.com/LLNL/UnifyFS/blob/be26e612cd5b5dcecc4f92f495970535605aaf3e/client/src/unifyfs-sysio.c#L667

PR #473, opened 2020-03-23
Add unit tests for lseek and fseek and fix related bugs

CamStan:
### Description (Updated)

This adds unit tests for our `lseek()`, `fseek()`, `ftell()`, and `rewind()`wrappers and fixes some related bugs. These mostly had to do with setting errno when invalid offsets are passed.

The `-D_GNU_SOURCE` is needed in `client/src/Makefile.am` in order to use `SEEK_DATA` and `SEEK_HOLE`.

Also adds tests for `SEEK_DATA` and `SEEK_HOLE` within `lseek()`, and adds the fallback support as stated in the man page:

> For  the  purposes of these operations, a hole
       is a sequence of zeros that (normally) has not
       been allocated in the underlying file storage.
       However, a  file  system  is  not  obliged  to
       report  holes,  so  these operations are not a
       guaranteed mechanism for mapping  the  storage
       space actually allocated to a file.  (Further‐
       more, a sequence of zeros  that  actually  has
       been written to the underlying storage may not
       be reported  as  a  hole.)   In  the  simplest
       implementation,  a file system can support the
       operations by making SEEK_HOLE  always  return
       the  offset of the end of the file, and making
       SEEK_DATA always return offset (i.e., even  if
       the  location referred to by offset is a hole,
       it can be considered to consist of  data  that
       is a sequence of zeros).

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-stdio.c
+++ b/client/src/unifyfs-stdio.c
@@ -892,6 +892,11 @@ static int unifyfs_fseek(FILE* stream, off_t offset, int whence)
     switch (whence) {
     case SEEK_SET:
         /* seek to offset */
+        if (offset < 0) {
+            /* negative offset is invalid */
+            errno = EINVAL;
+            return -1;
+        }
         current_pos = offset;
         break;
     case SEEK_CUR:
@@ -901,8 +906,12 @@ static int unifyfs_fseek(FILE* stream, off_t offset, int whence)
             errno  = EOVERFLOW;
             return -1;
         }
+        if (current_pos + offset < 0) {
+            /* offset is negative and will result in a negative position */
+            errno = EINVAL;
+            return -1;
+        }
         current_pos += offset;
-
         break;
     case SEEK_END:
         /* seek to EOF + offset */
@@ -910,6 +919,11 @@ static int unifyfs_fseek(FILE* stream, off_t offset, int whence)
         if (unifyfs_would_overflow_offt(filesize, offset)) {
             s->err = 1;
             errno  = EOVERFLOW;
+            return -1;
+        }
+        if (filesize + offset < 0) {
+            /* offset is negative and will result in negative position */
+            errno = EINVAL;
             return -1;
         }
         current_pos = filesize + offset;
@@ -1559,7 +1573,11 @@ void UNIFYFS_WRAP(rewind)(FILE* stream)
         /* lookup stream */
         unifyfs_stream_t* s = (unifyfs_stream_t*) stream;
 
-        /* TODO: check that stream is active */
+        /* check that stream is active */
+        if (s->fd < 0) {
+            errno = EBADF;
+            return;
+        }
 
         /* seek to front of file */
         int rc = unifyfs_fseek(stream, (off_t) 0L, SEEK_SET);
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -885,15 +885,15 @@ off_t UNIFYFS_WRAP(lseek)(int fd, off_t offset, int whence)
         unifyfs_fd_t* filedesc = unifyfs_get_filedesc_from_fd(fd);
 
         /* get current file position */
-        off_t logical_eof, current_pos = filedesc->pos;
-
-        /* TODO: support SEEK_DATA and SEEK_HOLE? */
+        off_t current_pos = filedesc->pos;
+        off_t logical_eof;
 
         /* compute final file position */
         switch (whence) {
         case SEEK_SET:
             /* seek to offset */
             if (offset < 0) {
+                /* negative offset is invalid */
                 errno = EINVAL;
                 return (off_t)(-1);
             }
@@ -902,6 +902,7 @@ off_t UNIFYFS_WRAP(lseek)(int fd, off_t offset, int whence)
         case SEEK_CUR:
             /* seek to current position + offset */
             if (current_pos + offset < 0) {
+                /* offset is negative and will result in negative position */
                 errno = EINVAL;
                 return (off_t)(-1);
             }
@@ -911,10 +912,31 @@ off_t UNIFYFS_WRAP(lseek)(int fd, off_t offset, int whence)
             /* seek to EOF + offset */
             logical_eof = unifyfs_fid_logical_size(fid);
             if (logical_eof + offset < 0) {
+                /* offset is negative and will result in negative position */
                 errno = EINVAL;
                 return (off_t)(-1);
             }
             current_pos = logical_eof + offset;
+            break;
+        case SEEK_DATA:
+            /* Using fallback approach: always return offset */
+            logical_eof = unifyfs_fid_logical_size(fid);
+            if (offset < 0 || offset > logical_eof) {
+                /* negative offset and offset beyond EOF are invalid */
+                errno = ENXIO;
+                return (off_t)(-1);
+            }
+            current_pos = offset;
+            break;
+        case SEEK_HOLE:
+            /* Using fallback approach: always return offset for EOF */
+            logical_eof = unifyfs_fid_logical_size(fid);
+            if (offset < 0 || offset > logical_eof) {
+                /* negative offset and offset beyond EOF are invalid */
+                errno = ENXIO;
+                return (off_t)(-1);
+            }
+            current_pos = logical_eof;
             break;
         default:
             errno = EINVAL;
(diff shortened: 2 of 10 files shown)


CamStan:
I can add the fallback support for SEEK_DATA and SEEK_HOLE to this PR.

client/src/unifyfs-sysio.c
```diff
@@ -885,23 +885,36 @@ off_t UNIFYFS_WRAP(lseek)(int fd, off_t offset, int whence)
         unifyfs_fd_t* filedesc = unifyfs_get_filedesc_from_fd(fd);
 
         /* get current file position */
-        off_t current_pos = filedesc->pos;
+        off_t logical_eof, current_pos = filedesc->pos;
```
adammoody: Let's split this declaration into two lines for clarity.

adammoody:
Thanks, @CamStan .  Nice find.  We should support the fallback for SEEK_HOLE/DATA at least.  Sounds like those can effective behave similarly to SEEK_END/SEEK_CUR.

With this, do you mind also doing a check of fseek?

CamStan:
> We should support the fallback for SEEK_HOLE/DATA at least. Sounds like those can effective behave similarly to SEEK_END/SEEK_CUR.
> 
> With this, do you mind also doing a check of fseek?

Yup, I can add all that to this PR

client/src/unifyfs-stdio.c
```diff
@@ -1559,7 +1573,12 @@ void UNIFYFS_WRAP(rewind)(FILE* stream)
         /* lookup stream */
         unifyfs_stream_t* s = (unifyfs_stream_t*) stream;
 
-        /* TODO: check that stream is active */
+        /* check that stream is active */
+        int fd = s->fd;
+        if (s->fd < 0) {
+            errno = EBADF;
+            return;
+        }
```
CamStan: This should be good for review now. It's currently in three logically separate commits, but can squash them if need be.

One question though. One of the tests I wrote was making `rewind` fail when called on a non-open stream. I noticed this `TODO` within most of our `stdio` wrappers and did the check in this way, which made the test pass. However, not sure if this is the right approach as I'm assuming there is a reason the check hadn't been done in any of the wrappers yet.

client/src/unifyfs-sysio.c
```diff
@@ -886,22 +886,57 @@ off_t UNIFYFS_WRAP(lseek)(int fd, off_t offset, int whence)
...
-            current_pos = unifyfs_fid_logical_size(fid);
+            logical_eof = unifyfs_fid_logical_size(fid);
+            if (logical_eof + offset < 0) {
+                /* offset is negative and will result in negative position */
+                errno = EINVAL;
+                return (off_t)(-1);
+            }
+            current_pos = logical_eof + offset;
+            break;
+        case SEEK_DATA:
```
adammoody: What does the team think of adding a ```#ifdef _GNU_SOURCE``` to conditionally compile these?  Worth doing that?

client/src/unifyfs-stdio.c
CamStan: Noticed I assigned to a local var but didn't use. I'll fix that.

And I verified that `unifyfs_stream_init` does also set `fd` back to -1.
https://github.com/LLNL/UnifyFS/blob/872cb269f23d8e6473ba43032b47e34d5259769b/client/src/unifyfs-stdio.c#L1851-L1852

adammoody:
Nice addition, @CamStan .

I don't see SEEK_HOLE/DATA in the man page for fseek, but are those supported there?  Guessing not, since you didn't add them.

I also realize now that a lot of these lseek/fseek tests will fail upon merging https://github.com/LLNL/UnifyFS/pull/437 due to our lack of implied fsync noted in https://github.com/LLNL/UnifyFS/issues/470.

client/src/unifyfs-sysio.c
CamStan: I had that thought as well, and wouldn't be opposed. If so, does this effect how it is added in the Makefile (i.e., does it need to become a config option or anything)?

CamStan:
> I don't see SEEK_HOLE/DATA in the man page for fseek, but are those supported there? Guessing not, since you didn't add them.

I didn't see them in the man page either and so assumed they were not supported in `fseek`. A quick test program (not using UnifyFS) of this seem to confirm that (did this with both SEEK_DATA and SEEK_HOLE):
```
/* continued from tests that already wrote to a file */

/* fseek passed end of file */
ret = fseek(fd, 25, SEEK_END);

printf("ftell after seeking beyond end of file with SEEK_END\n");
pos = ftell(fd);
printf("pos: %ld, errno: %d, strerror: %s\n", pos, errno, strerror(errno));

/* fwrite passed end of file to create a hole */
rc = fwrite("hello universe", 15, 1, fd);
printf("ftell after write\n");
pos = ftell(fd);
printf("pos: %ld, errno: %d, strerror: %s\n", pos, errno, strerror(errno));

/* fseek with SEEK_DATA should go to current offset (15) or offset of first data (37) */
printf("fseek to data passed the hole with SEEK_DATA\n");
ret = fseek(fd, 15, SEEK_DATA);
printf("ret: %d, errno: %d, strerror: %s\n", ret, errno, strerror(errno));

printf("ftell after seeking passed hole with SEEK_DATA\n");
pos = ftell(fd);
printf("pos: %ld, errno: %d, strerror: %s\n", pos, errno, strerror(errno));
```
Results in:
```
ftell after seeking beyond end of file with SEEK_END
pos: 37, errno: 0, strerror: Success
ftell after write
pos: 52, errno: 0, strerror: Success
fseek to data passed the hole with SEEK_DATA
ret: -1, errno: 22, strerror: Invalid argument
ftell after seeking passed hole with SEEK_DATA
pos: 52, errno: 0, strerror: Success
```

> I also realize now that a lot of these lseek/fseek tests will fail upon merging #437 due to our lack of implied fsync noted in #470.

CamStan:
> I don't see SEEK_HOLE/DATA in the man page for fseek, but are those supported there? Guessing not, since you didn't add them.

I didn't see them in the man page either and so assumed they were not supported in `fseek`. A quick test program (not using UnifyFS) of this seem to confirm that they are seen as invalid arguments (did this with both SEEK_DATA and SEEK_HOLE):
```
/* continued from tests that already wrote "hello world" to a file */

printf("fseek to beyond end of file with SEEK_END\n");
ret = fseek(fd, 25, SEEK_END);
printf("ret: %d, errno: %d, strerror: %s\n", ret, errno, strerror(errno));

printf("ftell after seeking beyond end of file with SEEK_END\n");
pos = ftell(fd);
printf("pos: %ld, errno: %d, strerror: %s\n", pos, errno, strerror(errno));

/* fwrite passed end of file to create a hole */
rc = fwrite("hello universe", 15, 1, fd);
printf("ftell after write\n");
pos = ftell(fd);
printf("pos: %ld, errno: %d, strerror: %s\n", pos, errno, strerror(errno));

/* fseek with SEEK_DATA should go to current offset (15) or offset of first data (37) */
printf("fseek to data passed the hole with SEEK_DATA\n");
ret = fseek(fd, 15, SEEK_DATA);
printf("ret: %d, errno: %d, strerror: %s\n", ret, errno, strerror(errno));

printf("ftell after seeking passed hole with SEEK_DATA\n");
pos = ftell(fd);
printf("pos: %ld, errno: %d, strerror: %s\n", pos, errno, strerror(errno));
```
Results in:
```
fseek to beyond end of file with SEEK_END
ret: 0, errno: 0, strerror: Success
ftell after seeking beyond end of file with SEEK_END
pos: 37, errno: 0, strerror: Success
ftell after write
pos: 52, errno: 0, strerror: Success
fseek to data passed the hole with SEEK_DATA
ret: -1, errno: 22, strerror: Invalid argument
ftell after seeking passed hole with SEEK_DATA
pos: 52, errno: 0, strerror: Success
```

---

> I also realize now that a lot of these lseek/fseek tests will fail upon merging #437 due to our lack of implied fsync noted in #470.

Is this because #437 changes `unifyfs_fid_logical_size(int fid)` to call `invoke_client_filesize_rpc()`?
If I just put an `fsync()` after the writes in these tests, will that fix it?

adammoody:
Thanks for double checking the question about fseek.  That's good to know.

Yes, I think adding an fsync() after the write() calls in the new tests here would fix the problem.  However, let's hold off on doing that.  First, we can look into what it would take to fix things so we can avoid those fsync() calls.  It may be enough to check the dirty bit and sync if needed before invoking the filesize rpc.  If things get too complicated with that approach, we can come back to add fsync() in the tests here.

adammoody:
Will go ahead and merge this, but let's still bring up the question about _GNU_SOURCE in our next group meeting.

CamStan:
> ... let's still bring up the question about _GNU_SOURCE in our next group meeting.

Will do

PR #476, opened 2020-04-01
Fix compiler warnings, set -Werror by default

tonyhutter:
### Description
This enables -Werror by default and fixes all compiler warnings on GCC 9.2.1.

### Motivation and Context
Fix build warnings

### How Has This Been Tested?
Build tested only on Fedora 31.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -1722,7 +1722,7 @@ int unifyfs_fd_logreadlist(read_req_t* in_reqs, int in_count)
         if (reqs != NULL) {
             free(reqs);
         }
-        return read_rc;
+        return ENOSPC;
     }
 
     /* order read request by increasing file id, then increasing offset */
--- a/common/src/ini.c
+++ b/common/src/ini.c
@@ -17,6 +17,7 @@ https://github.com/benhoyt/inih
 #include <string.h>
 
 #include "ini.h"
+#include "unifyfs_misc.h"
 
 #if !INI_USE_STACK
 #include <stdlib.h>
@@ -73,8 +74,7 @@ static char *find_chars_or_comment(const char *s, const char *chars)
 /* Version of strncpy that ensures dest (size bytes) is null-terminated. */
 static char *strncpy0(char *dest, const char *src, size_t size)
 {
-    strncpy(dest, src, size);
-    dest[size - 1] = '\0';
+    strlcpy(dest, src, size);
     return dest;
 }
 
--- a/common/src/unifyfs_keyval.c
+++ b/common/src/unifyfs_keyval.c
@@ -15,6 +15,7 @@
 #include "unifyfs_const.h"
 #include "unifyfs_keyval.h"
 #include "unifyfs_log.h"
+#include "unifyfs_misc.h"
 
 //#include "config.h"
 
@@ -543,7 +544,7 @@ static int unifyfs_fskv_init(unifyfs_cfg_t* cfg)
         }
 
         // find or create rank-specific subdir
-        snprintf(sharedfs_rank_kvdir, sizeof(sharedfs_rank_kvdir), "%s/%d",
+        scnprintf(sharedfs_rank_kvdir, sizeof(sharedfs_rank_kvdir), "%s/%d",
                  sharedfs_kvdir, kv_myrank);
         memset(&s, 0, sizeof(struct stat));
         rc = stat(sharedfs_rank_kvdir, &s);
@@ -591,7 +592,7 @@ static int unifyfs_fskv_fini(void)
                 continue;
             }
             memset(kvfile, 0, sizeof(kvfile));
-            snprintf(kvfile, sizeof(kvfile), "%s/%s",
+            scnprintf(kvfile, sizeof(kvfile), "%s/%s",
                      localfs_kvdir, de->d_name);
             rc = remove(kvfile);
             if (rc != 0) {
@@ -628,7 +629,7 @@ static int unifyfs_fskv_fini(void)
                     continue;
                 }
                 memset(rank_kvfile, 0, sizeof(rank_kvfile));
-                snprintf(rank_kvfile, sizeof(rank_kvfile), "%s/%s",
+                scnprintf(rank_kvfile, sizeof(rank_kvfile), "%s/%s",
                         sharedfs_rank_kvdir, de->d_name);
                 rc = remove(rank_kvfile);
                 if (rc != 0) {
@@ -688,8 +689,9 @@ static int unifyfs_fskv_lookup_local(const char* key,
     FILE* kvf;
     char kvfile[UNIFYFS_MAX_FILENAME];
     char kvalue[kv_max_vallen];
+    int rc;
 
-    snprintf(kvfile, sizeof(kvfile), "%s/%s",
+    scnprintf(kvfile, sizeof(kvfile), "%s/%s",
              localfs_kvdir, key);
     kvf = fopen(kvfile, "r");
     if (NULL == kvf) {
@@ -697,8 +699,12 @@ static int unifyfs_fskv_lookup_local(const char* key,
         return (int)UNIFYFS_ERROR_KEYVAL;
     }
     memset(kvalue, 0, sizeof(kvalue));
-    fscanf(kvf, "%s\n", kvalue);
+    rc = fscanf(kvf, "%s\n", kvalue);
     fclose(kvf);
+    if (rc != 1) {
+        *oval = NULL;
+        return (int)UNIFYFS_FAILURE;
+    }
 
     *oval = strdup(kvalue);
     return (int)UNIFYFS_SUCCESS;
@@ -711,12 +717,13 @@ static int unifyfs_fskv_lookup_remote(int rank,
     FILE* kvf;
     char rank_kvfile[UNIFYFS_MAX_FILENAME];
     char kvalue[kv_max_vallen];
+    int rc;
 
     if (!have_sharedfs_kvstore) {
         return (int)UNIFYFS_ERROR_KEYVAL;
     }
 
-    snprintf(rank_kvfile, sizeof(rank_kvfile), "%s/%d/%s",
+    scnprintf(rank_kvfile, sizeof(rank_kvfile), "%s/%d/%s",
              sharedfs_kvdir, rank, key);
     kvf = fopen(rank_kvfile, "r");
     if (NULL == kvf) {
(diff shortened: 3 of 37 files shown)


t/sys/creat64.c
```diff
@@ -47,15 +46,16 @@ int creat64_test(char* unifyfs_root)
     ok(fd >= 0, "creat64 non-existing file %s (fd=%d): %s",
        path, fd, strerror(errno));
 
-    rc = close(fd);
+    ok(close(fd) != 1, "close() worked");
```
CamStan, line 49: @tonyhutter, I have been doing some refactoring of the testing suite (in a PR coming later this week) to remove repetition and adding convenience functions as well as new tests and ended up doing the opposite with a lot of things like this. That is, removing `rc` and just turning it into `close(fd);` rather than putting it in a test.

Since, in this case, `close()` is already tested in [creat-close.c](https://github.com/LLNL/UnifyFS/blob/dev/t/sys/creat-close.c) and will show up there first if it breaks, is there an advantage I'm not thinking of to testing it again in latter testing files like this?
For context, my thoughts were that, in turning this into another test, we end up with a large number of our overall tests being repeated `open()` or `close()` tests which can clutter the results and make it look like we have more tests that we we actually do. Also can help keep the tests logically separated.

That said, it looks like we're getting some compiler warnings that are making travis fail for doing a `write()` but not asking for the result back. This is for tests where calls like `open()`, `write()`, and `close()` are simply used to set up the testing environment to test the function/wrapper in question.

I'm not sure the best approach here and would like any thoughts you have (I'm getting the feeling I've spent too much time thinking about the testing suite 😆).

tonyhutter:
@CamStan thanks for the comments.  Should Travis have been run for this PR?  I'm not seeing the results.

CamStan:
It ran, just not showing up here for some reason.
https://travis-ci.org/github/LLNL/UnifyFS/builds/669486621

t/sys/creat64.c
CamStan: As stated on the call, let's do what you've done here with our other tests and I'll do this for future ones as well. Since I'm touching a lot of the testing files this week, I can always add `-Werror` to the testing Makefile with my PR. In there I'll fix the warnings for writing but not asking for a result back.

tonyhutter:
Looks like Travis is catching some off my style errors, and also came up with a few new compiler warnings of its own.  Let me re-factor my commit to fix those too.

CamStan:
> ... also came up with a few new compiler warnings of its own.

Our current Travis environment defaults to version 5.4 of gcc and I imagine what warnings are caught varies by version. We can always adjust this (or add more) if need be.

common/src/unifyfs_keyval.c
```diff
@@ -543,8 +544,8 @@ static int unifyfs_fskv_init(unifyfs_cfg_t* cfg)
         }
 
         // find or create rank-specific subdir
-        snprintf(sharedfs_rank_kvdir, sizeof(sharedfs_rank_kvdir), "%s/%d",
-                 sharedfs_kvdir, kv_myrank);
+        assert(snprintf(sharedfs_rank_kvdir, sizeof(sharedfs_rank_kvdir), "%s/%d",
```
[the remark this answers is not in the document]
tonyhutter: GCC gives a warning basically saying 'you could be truncating, so you should be checking the return code".  However is super-super unlikely we're going to truncate the string (which is why the old code didn't check it), so the assert() was a convenient way to check it to get rid of the warning.  You're right though that it's going to break if they #define NODEBUG, so that's not gonna work.   I'll need to change all the asserts() to something else.

tonyhutter:
> One last question, can we put the '-Werror' in top-level Makefile.am to avoid putting it all the src dirs?

I tried that, but couldn't get it to work.  It's possible I was doing something wrong though.  Patches welcome!

CamStan:
@tonyhutter, Travis CI does now have a newer environment available that runs with gcc 7.4.0 ([here](https://docs.travis-ci.com/user/reference/bionic/)). We can alway update to that one and see if the warnings are closer to what you've tested on.

If this is desired, simply change this
https://github.com/LLNL/UnifyFS/blob/b0344034cdcc23cd10c36ec1e7fa1a29441aec34/.travis.yml#L3
to 
```
dist: bionic
```

tonyhutter:
> dist: bionic

@CamStan thanks, that's a good idea.  I'll include that in my next push.

t/sys/creat64.c
CamStan, line 49: -1 instead of 1?

t/sys/creat64.c
```diff
@@ -47,15 +46,16 @@ int creat64_test(char* unifyfs_root)
...
 
     /* Verify creating an already created file succeeds. */
     errno = 0;
     fd = creat64(path, mode);
     ok(fd >= 0, "creat64 existing file %s (fd=%d): %s",
        path, fd, strerror(errno));
 
-    rc = close(fd);
+    ok(close(fd) != 1, "close() worked");
```
CamStan, line 57: -1 instead of 1?

tonyhutter: Fixed in latest push

tonyhutter:
Ok, everything should be fixed now.

common/src/unifyfs_keyval.c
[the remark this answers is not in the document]
tonyhutter: Fixed latest push

common/src/unifyfs_const.h
```diff
@@ -49,6 +49,9 @@
 #define UNIFYFS_MAX_FILENAME KIB
 #define UNIFYFS_MAX_HOSTNAME 64
 
+/* key + '/' value string */
```
MichaelBrim, line 52: @tonyhutter I understand the intention of why you added this, but my assertion is that no file or directory path should ever be greater than UNIFYFS_MAX_FILENAME. If we don't think the current value (1 KiB) is big enough, we should increase its size, rather than defining a new constant for one particular use case.

tonyhutter: The problem is that we do a bunch of `sprintf(buf, %s/%s", str1, str2)` where `buf`, `str1` and `str2` are each UNIFYFS_MAX_FILENAME in length, so increasing UNIFYFS_MAX_FILENAME won't help get rid of the compiler warning.

MichaelBrim: This is the reason I always use `snprintf()` rather than `sprintf()`.

MichaelBrim: There are ways to pacify the compiler for the keyval code, but they are less than ideal from a code cleanliness point of view. One option would be to define UNIFYFS_MAX_DIR_NAMELEN (max name length of a directory path, including any trailing slash) and UNIFYFS_MAX_DIRENTRY_NAMELEN (max name length of a directory entry) and then make UNIFYFS_MAX_FILENAME = UNIFYFS_MAX_DIR_NAMELEN + UNIFYFS_MAX_DIRENTRY_NAMELEN. To me, that's ugly, but it would make the compiler happy.

tonyhutter: snprintf() doesn't help.   It complains that we should be checking the return code since we could be truncating the buffer.
```C
        char str1[1024] = "a";
        char str2[1024] = "b";
        char buf[1024];

        snprintf(buf, 1024, "%s/%s", str1, str2);
```
```
error: ‘%s’ directive output may be truncated writing up to 1023 bytes into a region of size between 0 and 1023 [-Werror=format-truncation=]
  snprintf(buf, 1024, "%s/%s", str1, str2);
```
Thats why I added `UNIFYFS_MAX_KVFILENAME` to get around it ever truncating.

MichaelBrim: This would avoid the compiler warning, but it seems excessive to me.
```
#define UNIFYFS_MAX_DIRENTRY_NAMELEN 64
#define UNIFYFS_MAX_DIR_NAMELEN (UNIFYFS_MAX_FILENAME - UNIFYFS_MAX_DIRENTRY_NAMELEN)
char dirname[UNIFYFS_MAX_DIR_NAMELEN];
char entryname[UNIFYFS_MAX_DIRENTRY_NAMELEN];
char buf[UNIFYFS_MAX_FILENAME+2]; // +2 allows for '/' and NUL
snprintf(buf, sizeof(buf), "%s/%s", dirname, entryname);
```

client/src/unifyfs-sysio.c
```diff
@@ -1722,7 +1722,7 @@ int unifyfs_fd_logreadlist(read_req_t* in_reqs, int in_count)
         if (reqs != NULL) {
             free(reqs);
         }
-        return read_rc;
+        return -ENOSPC;
```
MichaelBrim, line 1725: use positive ENOSPC

tonyhutter: Thanks, I'll fix in my next push

common/src/unifyfs_const.h
tonyhutter: @MichaelBrim yea, no matter how it's done, it's going to be a pain to get all these buffers sized correctly.  Might be better to just implement the kernel's `scnprintf()` function:

https://lwn.net/Articles/69419/

```C
/*
 * This is a re-implementation of the Linux kernel's scnprintf() function.
 *
 * It's snprintf() but returns the number of chars actually written into buf[]
 * not including the '\0'.  It also avoids the -Wformat-truncation warnings.
 */
int scnprintf(char *buf, size_t size, const char *fmt, ...)
{
        va_list args;
        int rc;

        va_start(args, fmt);
        rc = vsnprintf(buf, size, fmt, args);
        va_end(args);

        if (rc >= size) {
                /* We truncated */
                return size - 1;
        }

        return rc;
}
```

tonyhutter: I ended up implementing `scnprintf()` in my latest push.  This gets rid of the need for all the buffer resizing.

tonyhutter:
@MichaelBrim can you take another look at this code?  I added `scnprintf()` in my latest push to get around the all the buffer resizing I was doing.

client/src/unifyfs-sysio.c
MichaelBrim, line 1725: still negative

common/src/ini.c
```diff
@@ -17,6 +17,7 @@ Go to the project home page for more info:
 #include <string.h>
 
 #include "ini.h"
+#include "unifyfs_misc.h"
```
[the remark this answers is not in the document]
MichaelBrim: no one else chimed in on this, so marking as resolved

common/src/unifyfs_const.h
```diff
@@ -49,6 +49,9 @@
...
+#define UNIFYFS_MAX_KVFILENAME (UNIFYFS_MAX_FILENAME + 1 + UNIFYFS_MAX_FILENAME)
```
MichaelBrim, line 53: Don't need this anymore, correct?

PR #477, opened 2020-04-01
improve server management of app/client state

MichaelBrim:
### Description

Changes to application management:
 * app_id derived from mountpoint prefix (using same method used
   for gfid calculation)
 * `app_config` structure holds all state for a given application
 * `app_configs[]` - simple array of app_config structure pointers
   (replaces `app_config_list` arraylist)
   - added methods for adding/getting structure for given app_id

Changes to client management:
 * client_id assigned by server during mount rpc
   - can still be passed explicitly for re-attach
 * `app_client` structure holds all state for a given client
 * each app_config has an array of clients (`app_client*`)
   - added methods for adding/getting/removing clients

Miscellaneous code cleanup:
 * static functions should not use `unifyfs_` prefix, only ones
   used by external sources
 * don't use local function variables that shadow global variables
 * renamed `max_recs_per_slice` to `meta_slice_sz`

### Types of changes

### Checklist:

### DIFF
--- a/client/src/margo_client.c
+++ b/client/src/margo_client.c
@@ -15,6 +15,11 @@ static void register_client_rpcs(client_rpc_context_t* ctx)
     /* shorter name for our margo instance id */
     margo_instance_id mid = ctx->mid;
 
+    ctx->rpcs.attach_id = MARGO_REGISTER(mid, "unifyfs_attach_rpc",
+                       unifyfs_attach_in_t,
+                       unifyfs_attach_out_t,
+                       NULL);
+
     ctx->rpcs.mount_id = MARGO_REGISTER(mid, "unifyfs_mount_rpc",
                        unifyfs_mount_in_t,
                        unifyfs_mount_out_t,
@@ -182,8 +187,7 @@ int unifyfs_client_rpc_finalize(void)
         /* shut down margo */
         margo_finalize(ctx->mid);
 
-        /* free memory allocated for context structure,
-         * and set caller's pointer to NULL */
+        /* free memory allocated for context structure */
         free(ctx->client_addr_str);
         free(ctx);
     }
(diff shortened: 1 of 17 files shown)


client/src/margo_client.c
```diff
@@ -240,9 +280,17 @@ int invoke_client_mount_rpc(void)
     LOGDBG("Got response ret=%" PRIi32, ret);
 
     /* get slice size for write index key/value store */
-    unifyfs_key_slice_range = out.max_recs_per_slice;
+    unifyfs_key_slice_range = out.meta_slice_sz;
     LOGDBG("set unifyfs_key_slice_range=%zu", unifyfs_key_slice_range);
 
+    /* get assigned client id, and verify app_id */
+    unifyfs_client_id = (int) out.client_id;
```
adammoody, line 287: I didn't look closely, but in the case the client wants to reattach with a certain id, is there a place where a check is done to verify the returned client id matches the expected id?  Or does this get replaced with a new value?

adammoody:
Thanks, @MichaelBrim .  This looks great to me.  I'll let it sit for a day or two in case others have time to review it before we merge it.

client/src/margo_client.c
MichaelBrim: The support for reattach is there with the new attach rpc, but we currently don't expose an API that is suitable for reattach. We used to use the client rank passed to unifyfs_mount() as the client id, but that's no longer viable with the server dynamically assigning client ids. One solution I have thought about is maintaining a map from client-provided rank to server-assigned client id, and then introducing an alternative unifyfs_remount(prefix, rank) that would look up the existing mapping and return the previously assigned client id. In general, proper support for reattach requires a broader discussion with the team to define how we think it should work, and then we can define methods to support that.

MichaelBrim: and to better answer your specific question, currently, the client id is always assigned by the output value from the mount rpc.

adammoody: Thanks for the clarification.  I like the idea of a map or something, from a client "name" to a server assigned id.

adammoody:
@MichaelBrim , I merged in other PRs that led to a conflict here.

MichaelBrim:
This has been rebased on latest dev.

adammoody:
Thanks, @MichaelBrim !

PR #478, opened 2020-04-01
Move _GNU_SOURCE define to before system headers

CamStan:
### Description

Moves the _GNU_SOURCE definition to before the system headers in unifyfs-internal.h so it is properly defined for the `lseek` cases needed in #473. In turn, removes _GNU_SOURCE from client/src/Makefile.am to avoid redefine warnings. Also removed `#define __USE_GNU` as `#define _GNU_SOURCE` does this implicitly.

Our travis.yml runs `make distcheck` which apparently ignores configure options. Need to use `DISTCHECK_CONFIGURE_FLAGS` envar to pass config options to `make distcheck`.
This also updates .travis.yml to fix warnings caused by depricated `sudo` key, and missing `os` key.

### Motivation and Context

Properly define `_GNU_SOURCE` to remove warnings and make available for wrappers that need it (i.e., `lseek`).

### Types of changes

### Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: c
-sudo: required
 dist: xenial
+os: linux
 
 addons:
   apt:
@@ -74,9 +74,10 @@ before_cache:
 script:
   # Force git to update the shallow clone and include tags so git-describe works
   - git fetch --unshallow --tags
-  - sh autogen.sh
-  - ./configure --enable-fortran || cat config.log
-  - make -k && make distcheck
+  - export DISTCHECK_CONFIGURE_FLAGS="--enable-fortran"
+  - ./autogen.sh
+  - ./configure $DISTCHECK_CONFIGURE_FLAGS || cat config.log
+  - make distcheck
   - ./scripts/checkpatch.sh || test "$TEST_CHECKPATCH_ALLOW_FAILURE" = yes
 
 after_failure:
--- a/client/src/Makefile.am
+++ b/client/src/Makefile.am
@@ -20,8 +20,7 @@ endif
 
 CLIENT_COMMON_CPPFLAGS = \
   -I$(top_builddir)/client \
-  -I$(top_srcdir)/common/src \
-  -D_GNU_SOURCE
+  -I$(top_srcdir)/common/src
 
 CLIENT_COMMON_CFLAGS = \
   $(MPI_CFLAGS) \
--- a/client/src/unifyfs-internal.h
+++ b/client/src/unifyfs-internal.h
@@ -57,6 +57,10 @@
  * -------------------------------
  */
 
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 // system headers
 #include <aio.h>
 #include <assert.h>
@@ -85,7 +89,6 @@
 #include <wchar.h>
 #include <dirent.h>
 
-#define _GNU_SOURCE
 #include <pthread.h>
 #include <sched.h>
 
@@ -147,7 +150,6 @@
  * dlsym */
 
 /* we need the dlsym function */
-#define __USE_GNU
 #include <dlfcn.h>
 
 /* define a static variable called __real_open to record address of


CamStan:
Looks like `_GNU_SOURCE` isn't in the default travis environment. I'll add that to the yaml as well.

tonyhutter:
Is there a real-world need to build without _GNU_SOURCE?  If not, then I don't think we should include it.  It's just another permutation we'd have to build & test against.

CamStan:
Was decided on the Monday call to not force people to build with it since we didn't have it in the client until a recent PR. Something about portability issues can cause?

CamStan:
Finally got it working. Was a combination of a bug in the creation of my configure option and also apparently configure options passed to `./configure` don't persist in `make distcheck` ([here](https://www.gnu.org/software/automake/manual/html_node/Checking-the-Distribution.html)). 

Updated travis.yml to fix this and remove deprecated keys that travis was giving us warnings about.
Should be good for review now.

tonyhutter:
> Something about portability issues it can cause? 

It's only an issue if you're building on BSD or Solaris or something, which we don't care about.  Gotcha, bmi, and mercury all #define _GNU_SOURCE, so it's not like calling configuring Unify with `--enable-gnu-source` will actually cause Unify to actually build on those non-Linux platforms.  It's also going to be really annoying to litter our code with:
```C
#ifdef __USE_GNU
<code>
#endif  
```
... whenever we need to use _GNU_SOURCE.  And we'll have to ./configure and build Unify twice to test our code anytime we add a new  #ifdef __USE_GNU block, which is also annoying.    The work I'm doing on wrapping fcntl() needs _GNU_SOURCE, so it affects me.  

My vote is no on this.

MichaelBrim:
I tend to agree with @tonyhutter on this. We are only targeting Linux. To avoid the redefintion warnings, we can simply do:
```
#ifndef _GNU_SOURCE
#define _GNU_SOURCE
#endif
```

CamStan:
@tonyhutter & @MichaelBrim, so do you recommend just leaving `-D_GNU_SOURCE` in client/src/Makefile.am as was added in [PR 473](https://github.com/LLNL/UnifyFS/commit/5ea739194caccb2dace3a9bd7785e07ec83616be) so that we always build with it?

I don't have a strong opinion on this either way. @adammoody, do you have any thoughts as you originally suggested this? @boehms also had some thoughts on the call. I'm fine with backing out this PR, just want to make sure we get a consensus.

adammoody:
@CamStan , I don't have a strong opinion either way either.  Sorry that led to you doing a bunch of unnecessary work at this point!  Ugh.  My hope was that by bringing this up on the call, we would reach a consensus.

MichaelBrim:
@CamStan doing it in the Makefile is still problematic I think, due to the redefinition warnings (you should verify). My suggestion is to use the #ifndef method I suggested above, and put that in unifyfs_internal.h somewhere near the top of the file before all the system #includes.

CamStan:
@adammoody, I still learned some things while working on this, so I'm not worried about it. And I can push up the fixes for travis in a different PR.

@MichaelBrim, it appears the redefinition warnings go away when I pull the `#define _GNU_SOURCE` out of unifyfs-internal.h and leave it in the Makefile. I can change this PR or submit a new one that either
* pulls out the `#define _GNU_SOURCE` or
* moves `#define _GNU_SOURCE` to above the system #includes in unifyfs-internal.h and removes it from the makefile 

From the reading I've done, it seems they are two ways of doing the same thing as long as you don't do both.

CamStan:
@MichaelBrim & @tonyhutter, updated this PR and changed the description/title. Essentially just a small bug fix now so that `GNU_SOURCE` is available for the `lseek` definitions added in my last PR as this is how I should have done it in that PR to begin with.

Also fixes the bug and warnings that were in our .travis.yml.

CamStan:
Oh, and @tonyhutter, this will conflict with your removal of `#define _GNU_SOURCE` in #476.

PR #480, opened 2020-04-15
undef before (re) defining MIN/MAX macros

sandrain:
The recent addition of `-Werror` fails the build because of redefiniting the `MIN`/`MAX` macros. This prevents the compiler warning by `undef` those macros before defining them.

```
../../../../../client/src/unifyfs-sysio.c:50: error: "MAX" redefined [-Werror]
 #define MAX(a, b) (a > b ? a : b)
 
In file included from /ccs/techint/home/hs2/projects/UnifyFS/__run/testbed/prefix/include/margo-diag.h:10,
                 from /ccs/techint/home/hs2/projects/UnifyFS/__run/testbed/prefix/include/margo.h:20,
                 from ../../../../../client/src/margo_client.h:8,
                 from ../../../../../client/src/unifyfs-sysio.c:45:
/usr/include/sys/param.h:98: note: this is the location of the previous definition
 #define MAX(a,b) (((a)>(b))?(a):(b))
```

### How Has This Been Tested?
RHEL7.4/x86_64.

### Types of changes

### DIFF
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -46,8 +46,9 @@
 #include "ucr_read_builder.h"
 #include "seg_tree.h"
 
-
-#define MAX(a, b) (a > b ? a : b)
+#ifndef MAX
+#define MAX(a, b) ((a) > (b) ? (a) : (b))
+#endif
 
 /* ---------------------------------------
  * POSIX wrappers: paths
--- a/common/src/seg_tree.c
+++ b/common/src/seg_tree.c
@@ -36,8 +36,13 @@
 #include "seg_tree.h"
 #include "tree.h"
 
-#define MIN(a, b) (a < b ? a : b)
-#define MAX(a, b) (a > b ? a : b)
+#ifndef MIN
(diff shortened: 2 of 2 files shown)


sandrain:
@MichaelBrim I actually was for that direction but found the following post:

https://stackoverflow.com/questions/4234004/is-maxa-b-defined-in-stdlib-h-or-not

MichaelBrim:
I still prefer the #ifndef method. The system-provided definition is safer, since it places the args in parens. If you want to use #undef, you should change our definition to use the safer style.

tonyhutter:
@sandrain that article is talking about undefining `max()` because it's using a lower-case macro name (which is wrong).  That's not the case with our macros, since they're upper case.  We should be using the #ifndef method that @MichaelBrim suggests.

sandrain:
I will resubmit using ifndef. 

@tonyhutter I was thinking otherwise because it would be clearer to explicitly define the one that we use in our code instead of relying on the definition somewhere else. I was not specifically saying about the problem with the max macro.

PR #481, opened 2020-04-15
client: drop log_size field from meta since tracking moved to logio

adammoody:
This PR contains a number of changes:
* Drops the log_size field from the client file metadata structure, since tracking of the log size has moved into the new logio functions.
* Sets the needs_sync flag on the file metadata in fid_write instead of fd_write, which is deeper one layer in case functions may call fid_write directly instead of through fd_write.
* Renames fd_logreadlist to fid_read_reqs since the name "fd" was a misnomer as it operates on file ids and not file descriptors.  It also moves the function from unifyfs-sysio.c to unifyfs.c to go with fid_write which is effectively the counterpart write function.
* Changes fd_read/write to both return either UNIFYFS_SUCCESS or an error code, also adds a new output argument to return number of bytes actually read/written which may be less than the count requested.  Sets errno within wrapper functions instead of fd_read.
* Fix a number of wrappers to use original fd rather than the translated fd when a wrapper delegates calls to another wrapper.
* Adds output parameter to fid_write and fid_logio_write to return number of bytes written.
* Drops the chunks field from the client file metadata structure.  This used to track the number of fixed-size chunks allocated to a file, but it has been replaced with the new logio abstraction.

### Types of changes

### Checklist:


adammoody:
@MichaelBrim and @tonyhutter , now that the log management has been moved to the new logio functions, this PR is dropping the log_size field from our file metadata.  A consequence of that is that we can no longer use that field in the stat call to ask for the current log size, which is used in some of our existing test cases.  We could either drop those checks from our tests, or we could try to replace that with a corresponding lookup in the logio.

Do you have a preference?

If we opt to keep the checks in our test cases, @MichaelBrim , is there an existing recommended function we can use to query the current log size?

MichaelBrim:
The new logio code in common does track the total amount of data in each of the log storage areas (shmem and spill), but doesn't currently provide a method to retrieve the total data in the log. It would be relatively easy to add that.

adammoody:
@MichaelBrim and @tonyhutter , a second item I'd like to resolve in this PR...

Upon an error, read(), pread(), write(), and pwrite() all return -1 and set errno appropriately.

Internally, our wrappers for those four functions call either unifyfs_fd_read() or unifyfs_fd_write().  With unifyfs_fd_read() we return -1 and set errno.  With unifyfs_fd_write() we do not set errno and we return the value of errno as the return value.  I'd like to change this so that unifyfs_fd_read/write behave similarly (either both return -1 and set errno or both return errno as the return value).

Do you guys have a preference?

MichaelBrim:
My preference is that we only set errno in the method that wraps the system call, right before returning. That would mean returning the errno value from unifyfs_fd_write() so we can set it in the wrapper.

adammoody:
Oh, it also occurs to me that fd_write does not support short writes, because it has no way to return a count of bytes written.  I'll fix that, too.

tonyhutter:
@adammoody I don't have an opinion on keeping/removing log_size from stat.  I agree with @MichaelBrim on setting ERRNO in the high level wrapper functions.

adammoody:
I'll take a look at adding a new query for log size in there.

MichaelBrim:
@adammoody Yes, each write gets a new slot allocation. There is no support for fine-grained splitting of the data chunks, so if the app does tiny writes, it will use a lot of slots. An app that does use small writes could tune the chunk size to something much smaller. Another option, if we feel it's worth implementing, is to add some logic to the client to try to bundle smaller writes together before writing to the log, which would help avoid excessive fragmentation.

As for keeping per-file information about log usage, I think the only reasonable place to do that is in the per-file metadata. That would mean retaining the log size and the logic to maintain it.

adammoody:
I dropped the log size checks in the test cases for now.  Keeping track of log size by file is not quite trivial to add at this point.  It's a little easier task to report total log size.  However, our test cases would need to written to know the underlying slot size being used, so we'd probably also want a way to expose that.

CamStan:
> I decided to drop the log size checks in the test cases for now.

Just a heads up for my future self, this PR and #482 will have a merge conflict on these testing files.

adammoody:
@MichaelBrim , @tonyhutter , @CamStan , this one is ready for review now.

client/src/unifyfs-fixed.c
```diff
@@ -369,7 +369,8 @@ int unifyfs_fid_logio_write(int fid,
                             unifyfs_filemeta_t* meta,
                             off_t pos,
                             const void* buf,
-                            size_t count)
+                            size_t count,
+                            size_t* bytes)
```
tonyhutter, line 373: I would rename `bytes` to `nwritten` so it's clearer what it is.

client/src/unifyfs-sysio.c
```diff
@@ -628,19 +605,8 @@ int unifyfs_fd_write(int fd, off_t pos, const void* buf, size_t count)
...
-
     /* finally write specified data to file */
-    int write_rc = unifyfs_fid_write(fid, pos, buf, count);
-    if (write_rc == 0) {
-        unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
-        meta->needs_sync = 1;
-        meta->log_size = newlogsize;
-    }
+    int write_rc = unifyfs_fid_write(fid, pos, buf, count, bytes);
     return write_rc;
```
tonyhutter, line 610: So what happens if `write_rc == 0`, but `count != *bytes`?  Do we silently ignore that we didn't write enough bytes?  Would it make sense for `unifyfs_fid_write()` to instead return a `long`, equal to the number of bytes written or negative <ERRNO> on error?  That way we could encode all that information in a single value to pass it back up the stack.

client/src/unifyfs-fixed.c
adammoody: Renamed bytes to nwritten/nread depending on the call.

client/src/unifyfs-sysio.c
adammoody: Here bytes is also an output param of fd_write, so it's the caller of fd_write that must check whether count == *bytes.  I figured we could end up with a short write either with error or without error so decided we should just pass that up the stack for someone else to figure out what to do with it.  This can return an error code indicating why it failed, so we end up with two output values: the return code for an error code and the number of bytes actually written.

client/src/unifyfs.c
```diff
@@ -973,14 +961,664 @@ int unifyfs_fid_create_directory(const char* path)
...
+    return rc;
+}
+
 /* Write count bytes from buf into file starting at offset pos.
  *
  * Returns UNIFYFS_SUCCESS, or an error code
  */
-int unifyfs_fid_write(int fid, off_t pos, const void* buf, size_t count)
+int unifyfs_fid_write(
+    int fid,          /* global file id to write to */
```
MichaelBrim, line 1611: again, local file id, not global file id

client/src/unifyfs-internal.h
```diff
@@ -514,8 +508,16 @@ int unifyfs_fid_create_file(const char* path);
  * returns the new fid, or a negative value on error */
 int unifyfs_fid_create_directory(const char* path);
 
+int unifyfs_fid_read_reqs(read_req_t* in_reqs, int in_count);
+
 /* write count bytes from buf into file starting at offset pos */
-int unifyfs_fid_write(int fid, off_t pos, const void* buf, size_t count);
+int unifyfs_fid_write(
+    int fid,         /* global file id to write to */
```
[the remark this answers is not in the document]
adammoody: Yes, nice catch.

client/src/unifyfs.c
adammoody: Thanks

client/src/unifyfs-sysio.c
tonyhutter: Sorry, I totally blanked and didn't see that `unifyfs_fd_write()` was returning `nwritten`.  Ignore me.

adammoody:
> 
> 
> I only understand about 20% of what's going on in this patch 😄 but I don't see any obvious issues. Nice work!

Well, you're doing better than me, because I only understand 10% what I did.

adammoody:
@MichaelBrim , the merge is pending on your sign off.  Let me know if it's good.

PR #482, opened 2020-04-17
Create test get_size function and additional tests

CamStan:
### Description

This pulls the `get_size` function out of several tests, since it is defined and used multiple times, and turns it into a [testutil](t/lib/testutil.c) function.

Fixes a few compiler warnings/errors that showed up for gcc/4.9.3 when `-Werror` was enabled

Updates docs to show examples of how to write additional tests using various sharness and lib/tap testing functions.

Add a testing checkbox to the PR template

Also some additional tests are added as well as some refactoring, to include:
* Add additional tests for failure cases
    * tests with bad parameters or function calls on bad file descriptors
    * tests for bad function calls that segfault on posix. We segfault as well due to falling through to the `REAL` function. This how we want them handled?
* Add `read` tests to t/sys/write-read.c
* Add `__FILE__`/`__LINE__` to tests in files I was already touching for better future debugging
* Cleanup of redundant `errno` resetting
    * Removed calls to reset `errno` from unnecessary places. However, `errno` is not reset after successful function calls. Need to reset after test for failures to avoid subsequent tests appearing to have wrong `errno` value. Added note in docs.
* Turn function calls used for test setup into tests themselves for better future debugging
* Found two issues not yet addressed in this PR
    1. Test for `stat` after `unlink` was failing, as it should, but turns out it's failing with wrong `errno`. Call seems to be making it farther into our wrapper(s) then it should. Initial, quick, attempt to check `fid`/`gfid` where appropriate in wrappers didn't work (need to try again); something might not be getting removed in `unlink` call.

``` C
    todo("Failing with wrong errno. Should fail with errno=ENOENT=2");
    ok(stat(path, &sb) == -1 && errno == ENOENT, "%s:%d stat() after unlink fails (errno=%d): %s", __FILE__, __LINE__, errno, strerror(errno));
    end_todo;
```
Results in:
```
0500-sysio-static.t 351 - ../../t/sys/unlink.c:105 stat() after unlink fails (errno=5): Input/output error
```
*
    2. Test to `read` from a file open as write-only reads successfully, but returns 0 bytes read. Not as issue with `fread`. Potentially a problem in `open` call when flags are set.
``` C
    fd = open(path, O_WRONLY | O_CREAT, 0222);
    ok(fd != -1, "%s:%d open(%s) (fd=%d): %s", __FILE__, __LINE__, path, fd, strerror(errno));
    ⋮
    /* write, fsync, and size check tests */
    ⋮
    ok(lseek(fd, 0, SEEK_SET) == 0, "%s:%d lseek(0): %s", __FILE__, __LINE__, strerror(errno));
    todo("Successfully reads and gets 0 bytes back");
    ok(read(fd, buf, 15) == -1 && errno == EBADF, "%s:%d read() from file open as write-only (errno=%d): %s", __FILE__, __LINE__, errno, strerror(errno));
    end_todo;
```
Last test results in:
```
0500-sysio-static.t 79 - ../../../t/sys/write-read.c:88 read() from file open as write-only (errno=0): Success
```

Skipped checkpatch on the docs as it is complaining about a testing call that it thinks is a typo (The `is/isnt` function).

TEST_CHECKPATCH_SKIP_FILES=docs/testing.rst

### Motivation and Context

Add additional tests and clean up and remove duplicate code from tests.

### How Has This Been Tested?

Running unit tests locally and on Travis CI

### Types of changes

### Checklist:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -20,6 +20,7 @@
 - [ ] Performance enhancement (non-breaking change which improves efficiency)
 - [ ] Code cleanup (non-breaking change which makes code smaller or more readable)
 - [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Testing (addition of new tests or update to current tests)
 - [ ] Documentation (a change to man pages or other documentation)
 
 ### Checklist:
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -55,6 +55,24 @@ Here is an example of a sharness test:
         test "P" == "NP"
     '
 
+    # Various tests available to use inside test_expect_success/failure
+    test_expect_success "Show various available tests" '
+        test_path_is_dir /somedir
+        test_must_fail test_dir_is_empty /somedir
+        test_path_is_file /somedir/somefile
+    '
+
+    # Use test_set_prereq/test_have_prereq to conditionally skip tests
+    [[ -n $(which h5cc 2>/dev/null) ]] && test_set_prereq HAVE_HDF5
+    if test_have_prereq HAVE_HDF5; then
+        # run HDF5 tests
+    fi
+
+    # Can also check for prereq in individual test
+    test_expect_success HAVE_HDF5 "Run HDF5 test" '
+        # Run HDF5 test
+    '
+
     test_done
 
 .. _C-tests-label:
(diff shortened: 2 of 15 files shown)


adammoody:
For stat returning EIO after an unlink, that's probably coming from here:

https://github.com/LLNL/UnifyFS/blob/599a2429ef96b5da0f6b1e02543eb6adc585a232/client/src/unifyfs-sysio.c#L397

We could do better by calling our new error mapping function on returned error value rather than using EIO.  I'll bet we may have a lot of spots in the code like this.

For this particular case, I think our error value may be UNIFYFS_ERROR_MDHIM, which is more informative but it's still not ENOENT:

https://github.com/LLNL/UnifyFS/blob/599a2429ef96b5da0f6b1e02543eb6adc585a232/server/src/unifyfs_metadata.c#L334

adammoody:
@CamStan , looks good to me after a quick scan.  This one ready to go?

CamStan:
@adammoody yeah, should be good. There will be a merge conflict with #481 if this goes in first, or vice-versa.

adammoody:
Thanks @CamStan !

PR #483, opened 2020-04-17
Enable compile time (configure) option for using MPI for unifyfsd. 

sandrain:
This also fixes an compile error when `UNIFYFSD_USE_MPI` is defined.

- configure.ac: adding the configure option
- server/src/unifyfs_global.h: include config.h
- server/src/unifyfs_server.c: fix MPI_Init error

### Types of changes

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -132,6 +132,15 @@ AC_ARG_ENABLE(st-dev-workaround,
     ]
 ,[])
 
+# use mpi for bootstraping unifyfsd
+AC_ARG_ENABLE(unifyfsd-mpi,
+    [AS_HELP_STRING([--enable-unifyfsd-mpi],[Use MPI for bootstrapping unifyfsd])],[
+        AS_IF([test "x$enableval" = "xyes"],[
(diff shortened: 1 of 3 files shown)


MichaelBrim:
@sandrain These changes look fine, but why are they needed? We're trying to eliminate the use of MPI in the server, not keep it alive. The only thing still using MPI is MDHIM, which has its own call to MPI_Init_thread().

sandrain:
@MichaelBrim I am testing with our testbed cluster and using MPI for the server setup. Somehow the file kv method (without pmix) didn't work, which I was not able to debug the problem.

MichaelBrim:
@sandrain If this is just for the testbed, make your changes to your working copy and don't submit a PR. If you need help debugging the sharedfs server startup on the testbed, we can set up a screen share next week.

sandrain:
I can now run without using MPI, and the problem was the server hostfile (missing the number of hosts in the 1st line). Is there anywhere that we document about how the server host file should be formatted? If not, I think we need to document somewhere. I was able to figure out by looking at the source code.

PR #490, opened 2020-04-28
Margotreenew

adammoody:
This is a rebase of https://github.com/LLNL/UnifyFS/pull/427.  This will either be force pushed to that existing PR and dropped or just dropped altogether, but it's in a separate PR until then.

### Types of changes

### Checklist:


adammoody:
@sandrain , here is the old logic in MDHIM to compute the file size.
https://github.com/LLNL/UnifyFS/blob/4998aabf1450f925bffc2b4f901645a67322bbf4/server/src/unifyfs_request_manager.c#L523

Note how it does two things:
1) query all key value pairs from MDHIM and compute max extent offset
2) lookup metadata for the file from MDHIM and take the max of that file size with value obtained from step 1

This works in combination with the implementation for a truncate rpc.  On a truncate call, in addition to deleting or rewriting extents that exceed the new file size, we also record the file size in the MDHIM metadata.  For example, you can see that here:
https://github.com/LLNL/UnifyFS/blob/4998aabf1450f925bffc2b4f901645a67322bbf4/server/src/unifyfs_request_manager.c#L836

This is all to support a file that happens to end on a hole.  In case the file ends with a hole, we use the size recorded in the metadata to store the real file size.  To be scalable, I chose not to update the size that is recorded in the metadata on fsync.  All procs will call fsync, and perhaps many times during the run.  Instead, I only update the metadata size field on truncate and laminate calls, which are more likely to be called once and by a single process instead of by all processes.  During lamination, I compute the final file size and store that in the metadata.  So after lamination, we can just look up the file size from the metadata and skip querying the extent offsets.

With our new margotree, step 1 has been replaced with a collective operation to find the max extent across all servers.  However, we still need to account for step 2.  We probably need to update the truncate, laminate, and filesize rpcs.

MichaelBrim:
closing in preparation for new PR based on latest code

Issue #474, opened 2020-03-30
Define configuration to specify initial current working directory


adammoody:
Many applications write files in paths relative to the current working directory, where the user will cd into a particular directory and then start their application.  Since the OS does not know about UnifyFS directories, one cannot cd into a UnifyFS directory before starting the application.

To support these apps, we can define a configuration value that if set acts as the current working directory.  Then on intercepting a path, if the path is relative, we first append the "current working directory" as given by this setting.  For example,
```
UNIFYFS_CWD=/unifyfs/pathname
open("filename", O_WRONLY) means to open "/unifyfs/pathname/filename"
```

adammoody:
Related to https://github.com/LLNL/UnifyFS/issues/436

adammoody:
@clmendes , we have PR https://github.com/LLNL/UnifyFS/pull/493 where I'm hoping this might support HDF's use of external files with relative file names.  This moves things a step in the right direction, but we should test with the original test case you had to see if it fully works.  We might have some more work to do.

Issue #436, opened 2020-01-14
Add wrappers for getcwd and chdir


adammoody:
Looking at the strace from the HDF test the other day, there were getcwd calls.  I don't know if that application was using that on its UnifyFS files, but it raises the point that we're missing wrappers for that.  In particular, we should consider adding wrappers for chdir() and getcwd().

We could grab the current working directory when unify starts up, and then track any changes from application calls to chdir.  This would let an application chdir() into a UnifyFS directory space.  If the apps then opens a file with a relative path, we can prepend the current working dir to identify that it's a UnifyFS file.  We can intercept getcwd calls to return the UnifyFS path when in the UnifyFS space, and pass off to the normal getcwd call otherwise.

adammoody:
A related issue is dealing with path normalization, i.e., processing ".", "..", and extra or trailing "/" characters.  One option here is to use spath https://github.com/ECP-VeloC/spath/blob/90f5581e8d61d9dfab80988c1eb07058a28e4411/src/spath.h#L180

PR #493, opened 2020-04-28
client: add UNIFYFS_CLIENT_CWD for current working dir

adammoody:
Some applications ```cd``` into a working directory before starting the job, and then use relative path names for files within the run.  Since one cannot ```cd``` into a UnifyFS directory, we add a UNIFYFS_CLIENT_CWD configuration option.  When set, UnifyFS will prepend this string to any relative file name to act as a current working directory when determining whether to intercept a given path.

Resolves: https://github.com/LLNL/UnifyFS/issues/474

Related but separate to this topic:
- need to add wrappers for chdir/getcwd: https://github.com/LLNL/UnifyFS/issues/436
- need to add logic to normalize paths as in ```/unifyfs/foo/.././/bar --> /unifyfs/bar```

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -279,7 +279,7 @@ inline int unifyfs_stack_unlock(void)
     return 0;
 }
 
-inline void unifyfs_normalize_path(const char* path, char* normalized)
+static void unifyfs_normalize_path(const char* path, char* normalized)
 {
     /* if we have a relative path, prepend the current working directory */
     if (path[0] != '/' && unifyfs_cwd != NULL) {
@@ -2376,6 +2376,13 @@ static int unifyfs_init(void)
         cfgval = client_cfg.client_cwd;
         if (cfgval != NULL) {
             unifyfs_cwd = strdup(cfgval);
+
+            /* check that cwd falls somewhere under the mount point */
+            if (strncmp(unifyfs_cwd, unifyfs_mount_prefix,
+                unifyfs_mount_prefixlen) != 0) {
+                LOGERR("UNIFYFS_CLIENT_CWD '%s' must be within the mount '%s'",
+                    unifyfs_cwd, unifyfs_mount_prefix);
+            }
         }
 
         /* determine max number of files to store in file system */
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -294,7 +294,7 @@ static void unifyfs_normalize_path(const char* path, char* normalized)
 }
 
 /* sets flag if the path is a special path */
-inline int unifyfs_intercept_path(const char* path)
+inline int unifyfs_intercept_path(const char* path, char* upath)
 {
     /* don't intecept anything until we're initialized */
     if (!unifyfs_initialized) {
@@ -318,6 +318,12 @@ inline int unifyfs_intercept_path(const char* path)
             intercept = 0;
         }
     }
+
+    /* copy normalized path into upath */
+    if (intercept) {
+        strncpy(upath, target, UNIFYFS_MAX_FILENAME);
+    }
+
     return intercept;
 }
 
@@ -3038,7 +3044,8 @@ int unifyfs_transfer_file(const char* src, const char* dst, int parallel)
         return -ENOMEM;
     }
 
-    if (unifyfs_intercept_path(src)) {
+    char src_upath[UNIFYFS_MAX_FILENAME];
+    if (unifyfs_intercept_path(src, src_upath)) {
         dir = UNIFYFS_TX_STAGE_OUT;
         unify_src = 1;
     }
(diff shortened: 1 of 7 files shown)


adammoody:
We'll also want to store the full, normalized path in our internal structures rather than the path as given by the user.  One option would be to normalize the path in the intercept_path, return that as an output path, and then be sure we only use the normalized path name for internal unifyfs functions.

tonyhutter:
Do you plan to wrap `chdir()`/`getcwd()` in this PR?  That would remove the need for `UNIFYFS_CLIENT_CWD`, and allow you to easily write test cases.

client/src/unifyfs.c
```diff
@@ -276,6 +279,20 @@ inline int unifyfs_stack_unlock(void)
...
+    /* if we have a relative path, prepend the current working directory */
+    if (path[0] != '/' && unifyfs_cwd != NULL) {
+        /* got a relative path, add our cwd */
+        snprintf(normalized, UNIFYFS_MAX_FILENAME, "%s/%s", unifyfs_cwd, path);
+    } else {
+        snprintf(normalized, UNIFYFS_MAX_FILENAME, "%s", path);
+    }
+
+    /* TODO: normalize path to handle '.', '..',
+     * and extra or trailing '/' characters */
```
tonyhutter, line 293: Would `realpath()` help here?

tonyhutter: ...ah maybe not if it does real file lookups...

adammoody:
> 
> 
> Do you plan to wrap `chdir()`/`getcwd()` in this PR? That would remove the need for `UNIFYFS_CLIENT_CWD`, and allow you to easily write test cases.

We can add wrappers for chdir/getcwd in a follow on PR.

Actually, there is a need for CLIENT_CWD even with wrappers for chdir/getcwd.  Some users run a cd command in their job script and then launch their app, so that their app never actually make a call to chdir().

```
cd /dir
srun -n2 myapp
# myapp then calls open("file") --> users expect this to open "/dir/file"
```

But it's impossible for someone to ```cd /unifyfs```.  With this, they can now set ```export UNIFYFS_CLIENT_CWD=/unifyfs``` to fill that gap.

client/src/unifyfs.c
adammoody: Right, we need something like realpath but that operates on unifyfs paths.  Is there something available that works for path strings that don't actually exist?

tonyhutter: Not that I know of :disappointed:

tonyhutter:
> But it's impossible for someone to cd /unifyfs. With this, they can now set export UNIFYFS_CLIENT_CWD=/unifyfs to fill that gap.

Hadn't thought of that, thanks for the info!

adammoody:
Thanks for the recommendation to add a note to the docs, @CamStan .  Can you review that and let me know if you have suggestons?

client/src/unifyfs.c
```diff
@@ -2351,6 +2372,12 @@ static int unifyfs_init(void)
         unifyfs_max_long = LONG_MAX;
         unifyfs_min_long = LONG_MIN;
 
+        /* set our current working directory if user gave us one */
+        cfgval = client_cfg.client_cwd;
+        if (cfgval != NULL) {
+            unifyfs_cwd = strdup(cfgval);
+        }
```
CamStan, line 2379: Question, to see if it's worth considering; do we know what the UnifyFS mount point is at this point? If so, should we check that `client_cwd` matches the mount point here, and if not then at least print something to the error log if not wanting to fail?

Now that I write it, if according to what you put in the docs is true:
> The value specified in ``cwd`` must match the UnifyFS mount point.

can this just be a `yes|no` flag? If yes, then just prepend what they have set as their mount point to the path in your logic in `unifyfs_normalize_path`? Potentially less user error to worry about.

Not sure if either option would cause issues though if we're eventually on planning to support multiple mount points.

adammoody: That's a good idea about checking the mount point.  Yes, we probably can do that.  I'll look at adding that check.

I wanted people to be able to specify a path in case they want to specify something deeper in their directory space, like cwd=/unifyfs/dir1

adammoody: Bonus points:  Adding that check led to discovery of a bug in out intercept path logic.  With our old check, we would also intercept any paths with extra chars, so mount="/unifyfs" would also intercept any path like "/unifyfstrailingchars".

CamStan: > I wanted people to be able to specify a path in case they want to specify something deeper in their directory space, like cwd=/unifyfs/dir1

Ah, yup, that makes sense.

> With our old check, we would also intercept any paths with extra chars...

Wow, hadn't thought of that. Nice catch!

PR #497, opened 2020-05-07
Chdir

adammoody:
This adds the VeloC spath component to normalize paths similar to realpath().  The intercept_path function is updated to return a normalized path for intercepted paths, and wrappers then use the normalized path instead of the original path provided by the user.  This helps ensure we always refer to the same underlying file, regardless of the various strings one might use to name it.

Adds wrappers for chdir, fchdir, getcwd, getwd, and get_current_dir_name.  When unifyfs is initialized, we attempt to track the current working directory for all directory changes.  The starting current working dir is taken as UNIFYFS_CLIENT_CWD if specified, and the actual current working directory (getcwd) otherwise.  We then return the working directory in calls like getcwd.

Aims to resolve https://github.com/LLNL/UnifyFS/issues/436

Depends on https://github.com/LLNL/UnifyFS/pull/493

Tests added for chdir, getcwd, and get_current_dir_name.

Test for getwd throws compile error:
```
sys/chdir.c:126:5: error: 'getwd' is deprecated (declared at /usr/include/unistd.h:525) [-Werror=deprecated-declarations]
     str = getwd(pathmax);
```
TODO: add wrappers for openat(), dirfd()

### Types of changes

### Checklist:


adammoody:
@CamStan , I could use an assist with the travis build.  I'm trying to add spath via spack, which kicks off a fresh build of everything.  Things are failing on numactl with a compile error.  I have tried to install this through debian instead to avoid the spack build as a work around, but I can't get the package.yaml set up correctly.

CamStan:
@adammoody, had to play around with it a bit, but I pushed a commit to fix this. Now failing at the compile error you mentioned.

It looks like Spack was trying to install all these dependencies in order to install OpenMPI.
```
==> Warning: Skipping build of openmpi since numactl failed
==> Warning: Skipping build of spath since openmpi failed
```
OpenMPI was already in the travis environment, but was never added to the packages.yaml since UnifyFS will try to find mpicc itself. Spath (or at least the spackage for spath) isn't set up to do this, so Spack was trying to install OpenMPI again.

It looked to me like you had the packages.yaml set up correctly set up for numactl, so not sure why it wasn't finding it. However, the solution in my commit just bypasses the need to install any of those dependencies. Also fixed a few other things I ran into along they way.

adammoody:
Thanks, @CamStan !  That's a big help.  I'll debug the rest of that.

adammoody:
Something strange...  Accoding to the man page, the getcwd function should return NULL and set errno=ERANGE if the size of the buffer that the user provides is too small to hold the string and terminating NUL for the current worknig dir.  I have a test for that where I get the current working dir, compute it's length with strlen, and then call again with a buffer that is one character too short:
```
    /* check that we're in unifyfs root directory */
    str = getcwd(path, sizeof(path));
    ok(str != NULL, "%s:%d getcwd: %s",
       __FILE__, __LINE__, strerror(errno));
    ok(str != NULL && strcmp(str, unifyfs_root) == 0,
       "%s:%d getcwd returned %s expected %s",
       __FILE__, __LINE__, str, unifyfs_root);

    /* get length of current directory */
    size_t len = strlen(str); // this fails, getcwd below will not call our wrapper (gotcha or static)
    //size_t len = 25; // this works, getcwd will call our wrapper

    /* just to be sure we're still in the root unifyfs_root directory */
    rc = chdir(unifyfs_root);
    ok(rc == 0, "%s:%d chdir(%s): %s",
       __FILE__, __LINE__, unifyfs_root, strerror(errno));

    /* try getcwd with a buffer short by one byte,
     * should fail with ERANGE */
    str = getcwd(path, len); // pass
    ok(str == NULL && errno == ERANGE,
       "%s:%d getcwd(buf, %d): expected NULL got %s errno %s",
       __FILE__, __LINE__, len, str, strerror(errno));
    errno = 0;
```
However, that leads to the following: https://travis-ci.org/github/LLNL/UnifyFS/builds/693211044
```
intercepting chdir wrapper /tmp/tmp.G0mcYMKP70/mount upath /tmp/tmp.G0mcYMKP70/mount
ok 328 - ../../../t/sys/chdir.c:101 chdir(/tmp/tmp.G0mcYMKP70/mount): Success
PASS: 0500-sysio-static.t 328 - ../../../t/sys/chdir.c:101 chdir(/tmp/tmp.G0mcYMKP70/mount): Success
not ok 329 - ../../../t/sys/chdir.c:108 getcwd(buf, 25): expected NULL got / errno Success
FAIL: 0500-sysio-static.t 329 - ../../../t/sys/chdir.c:108 getcwd(buf, 25): expected NULL got / errno Success
#   Failed test '../../../t/sys/chdir.c:108 getcwd(buf, 25): expected NULL got / errno Success'
#   at ../../../t/sys/chdir.c line 108.
```
After a bunch of debugging, it looks to me like this getcwd call is being passed to the real getcwd and not being intercepted.  Above, we're not invoking the wrapper -- a bunch of debug printf's that I added aren't showing up.  Also, getcwd is returning "/" which is the active current working dir that actually in as far as the OS is concerned due to a previous chdir("/") that we passed to the real chdir.  This happens in both gotcha and static tests.

However, if I modify the test to set the len variable with a hard coded value of 25 rather than by calling strlen:
```
    /* get length of current directory */
    //size_t len = strlen(str); // this fails, getcwd below will not call our wrapper (gotcha or static)
    size_t len = 25; // this works, getcwd will call our wrapper
```
then it all works as expected: https://travis-ci.org/github/LLNL/UnifyFS/builds/693216600
```
PASS: 0500-sysio-static.t 328 - ../../../t/sys/chdir.c:101 chdir(/tmp/tmp.554WMiy4kn/mount): Success
PASS: 0500-sysio-static.t 329 - ../../../t/sys/chdir.c:108 getcwd(buf, 25): expected NULL got (null) errno Numerical result out of range
```
So it seems that by calling strlen in the program, the compiler / linker routes this particular getcwd call to a different name?

Other getcwd calls in the same test do call our wrapper in both gotcha and static, regardless of whether strlen is called or not.

adammoody:
I added a line to travis to run ```nm``` on the sysio-static.t executable, and that provides a clue:
```
# when the chdir.c test calls strlen
$ nm unifyfs-0.1.0*/_build/sub/t/sys/sysio-static.t | grep -i getcwd
                 U __getcwd_chk@@GLIBC_2.4
                 U getcwd@@GLIBC_2.2.5
000000000001a760 T __wrap_getcwd
000000000001ab10 T __wrap___getcwd

# when strlen is commented out
$ nm unifyfs-0.1.0*/_build/sub/t/sys/sysio-static.t | grep -i getcwd
                 U getcwd@@GLIBC_2.2.5
000000000001a6f0 T __wrap_getcwd
000000000001aaa0 T __wrap___getcwd
```
Seems like its picked up a __getcwd_chk from a different glibc library.

adammoody:
I think adding a wrapper for __getcwd_chk helps cover more cases.  While testing another interesting thing showed up.  At one point, I was trying to print the address of the getcwd() function.  I added a few lines like the following: ```printf("getcwd=%p\n", &getcwd);```.

This caused the gotcha tests to fail in a bunch of new places.  There were several tests that used to call the wrapper which stopped after adding that print.  It seems like that acted to lock in the real getcwd somehow.
https://travis-ci.org/github/LLNL/UnifyFS/builds/693637832

Dropping those printfs and adding the __getcwd_chk finally passes all tests on travis.
https://travis-ci.org/github/LLNL/UnifyFS/builds/693635851

CamStan:
Nice debugging @adammoody! Albeit potentially a bit tedious. 😅 Almost wanted to keep all your commits just to see the history of it.

adammoody:
Yeah, I counted up those commits.  Had about 40 different commits pushed to travis, spanned 3 different days and a total of 7 hours, sigh...

Thanks for your help, @CamStan !  You saved me some time banging my head on the wall on this one. :-)

adammoody:
In case we hit something like this again, a tip on running nm against our binaries within .travis, first add a line like the following to locate where the executable is:
```
- make distcheck
- find . -print | grep sysio-static.t
```
Then seeing what that finds after one push to travis, you can add a second line to run nm like.
```
- nm unifyfs-0.1.0*/_build/sub/t/sys/sysio-static.t | grep -i getcwd
- nm unifyfs-0.1.0*/_build/sub/t/sys/.libs/sysio-gotcha.t | grep -i getcwd
```

adammoody:
Thanks for the tip on __getcwd, @MichaelBrim .  I added a wrapper for that, but gotcha didn't find it and so I took it back out for now.  There was a third function we needed in __getcwd_chk.

It's still a mystery to me on how the compiler picks which getcwd calls in the source code get mapped to getcwd() and which ones get mapped to __getcwd_chk().

adammoody:
Oh, I may have a theory...

The prototype for this other function is:

```__getcwd_chk(char* buf, size_t size, size_t buflen)```

It is supposed to check that size <= buflen and abort with an exit() if not to prevent buffer overruns:
http://refspecs.linux-foundation.org/LSB_4.0.0/LSB-Core-generic/LSB-Core-generic/libc---getcwd-chk-1.html

For my buffer in the test code, I am using a hard coded ```char path[64]```.  Thus, the compiler knows I have a 64-char buffer.  When I then have code like the following:
```
size_t len = 25;
getcwd(path, len);
```
The compiler can determine at compile time that there can be no buffer overrun, since it knows the value of len (25) is less than the size of my buffer (64).  In this case, it knows it's fine to call getcwd.

In the case of:
```
size_t len = strlen(str);
getcwd(path, len)
```
then the value of len is not determined until run time.  The compiler still knows I have a buffer of 64 characters, though, so it can map this call to ```__getcwd_chk(path, len, 64)```.  That's a safe thing to do since it prevents a potential buffer overrun.

File: t/sys/chdir.c [hunk 1]
```diff
@@ -0,0 +1,361 @@
...
+    ok(rc == 0, "%s:%d chdir(%s): %s",
+       __FILE__, __LINE__, "/", strerror(errno));
+
+    /* check that we're in root directory */
+    str = getcwd(path, sizeof(path));
+    ok(str != NULL, "%s:%d getcwd: %s",
+       __FILE__, __LINE__, strerror(errno));
+    ok(str != NULL && strcmp(str, "/") == 0,
+       "%s:%d getcwd returned %s expected %s",
+       __FILE__, __LINE__, str, "/");
```
CamStan, line 71: More just an FYI, but liptap does have a string comparison test. Not necessary though.
```suggestion
    str = getcwd(path, sizeof(path));
    ok(str != NULL, "%s:%d getcwd: %s",
       __FILE__, __LINE__, strerror(errno));
    is(str, "/",
       "%s:%d getcwd returned %s expected %s",
       __FILE__, __LINE__, str, "/");
```

t/sys/chdir.c
```diff
@@ -0,0 +1,361 @@
...
+    ok(str != NULL, "%s:%d getcwd: %s",
+       __FILE__, __LINE__, strerror(errno));
+    ok(str != NULL && strcmp(str, buf2) == 0,
+       "%s:%d getcwd returned %s expected %s",
+       __FILE__, __LINE__, str, buf2);
+
+
+#if 0
+    /* change to root directory */
+    rc = chdir("/");
```
CamStan, line 188: Need this commented out section?

t/sys/chdir.c
```diff
@@ -0,0 +1,361 @@
...
+       "%s:%d get_current_dir_name returned %s expected %s",
+       __FILE__, __LINE__, str, buf2);
+    if (str != NULL) {
+        free(str);
+    }
+
+
+#if 0
+    /* change to root directory */
+    rc = chdir("/");
```
CamStan, line 323: This one as well?

client/src/unifyfs-sysio.c
```diff
@@ -1440,6 +1787,65 @@ ssize_t UNIFYFS_WRAP(pwrite64)(int fd, const void* buf, size_t count,
...
+            free(unifyfs_cwd);
+        }
+        unifyfs_cwd = strdup(path);
+        return 0;
+    } else {
+        MAP_OR_FAIL(fchdir);
+        int ret = UNIFYFS_REAL(fchdir)(fd);
+
+        /* if the change dir was successful,
+         * update our current working direcotry */
```
CamStan, line 1818: Oops, same typo here also:
```suggestion
         * update our current working directory */
```

examples/src/Makefile.am
```diff
@@ -83,20 +83,20 @@ test_ftn_flags   = $(AM_FCFLAGS) $(MPI_FFLAGS) \
    -I$(top_srcdir)/client/src -I$(top_srcdir)/common/src
 test_ftn_ldadd   = $(top_builddir)/client/src/libunifyfsf.la -lrt -lm $(FCLIBS)
 test_ftn_ldflags = $(AM_LDFLAGS) $(MPI_FLDFLAGS) \
-    $(FLATCC_LDFLAGS) $(FLATCC_LIBS)
+    $(FLATCC_LDFLAGS) $(FLATCC_LIBS) $(SPATH_LDFLAGS) $(SPATH_LIBS)
```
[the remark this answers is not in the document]
adammoody: Yeah, let me clean things up.  Actually, I ended up making SPATH required after all.  As an aside, we may later drop spath as a dependency later as @tonyhutter might have a function we can use instead.

client/src/unifyfs.c
```diff
@@ -2233,6 +2242,17 @@ static int unifyfs_init(void)
                 free(unifyfs_cwd);
                 unifyfs_cwd = NULL;
             }
+        } else {
+            /* user did not specify a CWD, so initialize with the actual
+             * current working dir */
+            //MAP_OR_FAIL(getcwd);
+            //int cwd_rc = UNIFYFS_REAL(getcwd)(cwdpath, sizeof(cwdpath));
```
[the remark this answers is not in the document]
adammoody: Nope, dropped them.

client/src/unifyfs-sysio.c
```diff
@@ -155,6 +155,353 @@ int UNIFYFS_WRAP(rmdir)(const char* path)
...
+            free(unifyfs_cwd);
+        }
+        unifyfs_cwd = strdup(upath);
+        return 0;
+    } else {
+        MAP_OR_FAIL(chdir);
+        int ret = UNIFYFS_REAL(chdir)(path);
+
+        /* if the change dir was successful,
+         * update our current working direcotry */
```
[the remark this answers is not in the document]
adammoody: Got it.

client/src/unifyfs-sysio.c
adammoody: Thanks.

t/sys/chdir.c
adammoody: Nice, I'll switch since that's cleaner.  Can we also pass a variable name in place of a hard-coded string, like ```is(str, unifyfs_root, ...);```?

t/sys/chdir.c
adammoody: Ideally, it'd be nice to include these tests so long as the compiler doesn't complain about getwd.  I've added a TODO to explain why these are commented.  I'm hoping we might figure out how to conditionally allow them for the compilers that allow it.

t/sys/chdir.c
adammoody: I'd like to activate these tests once our directory wrappers are working.  Added a TODO.

t/sys/chdir.c
CamStan: Yeah, should be able to. I haven't tested it myself using two variables though. Tony used it here, but only with one variable name
https://github.com/LLNL/UnifyFS/blob/1d12d24a7710dfd5ba7bc096993fffb1891f3ff2/t/sys/write-read.c#L152

t/sys/chdir.c
CamStan: Ah, yeah. Didn't look at what tests they were, just saw they were commented out. Sounds good.

t/sys/chdir.c
CamStan: Sounds good.

examples/src/Makefile.am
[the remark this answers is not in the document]
CamStan: I'll hold off on updating our spackage then, unless we get any user complaints.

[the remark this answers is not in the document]
adammoody: Dropped the SPATH flags, but SPATH is still optional actually.  We may want to make that required.

adammoody:
We probably need to watch out for other _chk() versions of various wrappers.  Here are some others I see:
```
>>: nm /usr/lib/libc.so.6 | grep _chk
001179b0 t __GI___vdprintf_chk
00115bf0 t __GI___vfprintf_chk
001170d0 t __GI___vfwprintf_chk
001159a0 t ___fprintf_chk
00115bf0 t ___vfprintf_chk
00117980 T __dprintf_chk
001160e0 T __fgets_chk
00116260 T __fgets_unlocked_chk
001171f0 T __fgetws_chk
00117370 T __fgetws_unlocked_chk
001159a0 T __fprintf_chk
00116640 T __fread_chk
001167b0 T __fread_unlocked_chk
00116e80 T __fwprintf_chk
00116580 T __getcwd_chk
00115d10 T __gets_chk
00116530 T __getwd_chk
00117df0 T __poll_chk
00117e30 T __ppoll_chk
001163b0 T __pread64_chk
00116360 T __pread_chk
00116300 T __read_chk
00116490 T __readlink_chk
001164f0 T __readlinkat_chk
001165c0 T __realpath_chk
001179b0 T __vdprintf_chk
00115bf0 T __vfprintf_chk
001170d0 T __vfwprintf_chk
```

t/sys/chdir.c
adammoody: Yep, variable names are fine.

examples/src/Makefile.am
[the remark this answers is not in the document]
CamStan: If there's no reason not to make it required (aside from another dependency users have to deal with), might as well.

Alternatively, with this implementation, we could always add it to our bootstrap.sh and to our spackage as an enabled variant by default. That way, users will always get it if installing the standard way, but still leaves them the option of opting out of it.

PR #504, opened 2020-05-14
placement nomination for stagein/stageout

craigsteffen:
Beginning of the PR to integrate file stage-in and stage-out
capability to the unifyfs command.  

### Description

This branch/PR will add calls in the unifyfs command to the 
transfer API to bring files into the UnifyFS file space during 
unifyfs launch, and stage files out when calling teardown but
before the servers are actually stopped.  

All the check lists will be filled in when this becomes a PR 
that's ready to integrate.  Until then this PR is a draft (as 
noted by the "WIP" tag).  

### Motivation and Context

This PR is to fulfill the stage-in/stage-out ECP Milestone for the 
end of May 2020.  

### How Has This Been Tested?

This commit is just comments where the hooks will go.  

Testing will be implemented and documented as the patch gets put together.  

### Types of changes

### Checklist:

### DIFF
--- a/util/unifyfs/src/unifyfs-rm.c
+++ b/util/unifyfs/src/unifyfs-rm.c
@@ -554,14 +554,20 @@ static size_t construct_server_argv(unifyfs_args_t* args,
 	        return -1;
             }
 
-	    //	    if (args->server_path != NULL) {
+	    //  The commented-out code lines will be uncommented once the
+	    //  transfer-stage infrastructure is stable enough that we can
+	    //  count on it being in the right place in the install.  For
+	    //  now, while we're rolling it out, we're requiring the user
+	    //  to explicitly supply the location of the transfer-stage
+	    //  binary.
+	    //	    if (args->transfer_exe != NULL) {
 	    server_argv[0] = strdup(args->transfer_exe);
 	    //    } else {
-	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	        server_argv[0] = strdup(LIBEXECDIR "/transfer-stage");
 	    //	    }
 	    argc = 1;
 
-	    // if we need to add arguments betwen the command and the
+	    // if we need to add arguments between the command and the
 	    // manifest file name they get added here.
 
 	    server_argv[argc] = strdup(args->stage_in);
@@ -581,14 +587,20 @@ static size_t construct_server_argv(unifyfs_args_t* args,
 	        return -1;
             }
 
-	    //	    if (args->server_path != NULL) {
+	    //  The commented-out code lines will be uncommented once the
+	    //  transfer-stage infrastructure is stable enough that we can
+	    //  count on it being in the right place in the install.  For
+	    //  now, while we're rolling it out, we're requiring the user
+	    //  to explicitly supply the location of the transfer-stage
+	    //  binary.
+	    //	    if (args->transfer_exe != NULL) {
 	    server_argv[0] = strdup(args->transfer_exe);
 	    //    } else {
-	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	        server_argv[0] = strdup(LIBEXECDIR "/transfer-stage");
 	    //	    }
 	    argc = 1;
 
-	    // if we need to add arguments betwen the command and the
+	    // if we need to add arguments between the command and the
 	    // manifest file name they get added here.
 
 	    server_argv[argc] = strdup(args->stage_out);
--- a/util/unifyfs/src/unifyfs-rm.c
+++ b/util/unifyfs/src/unifyfs-rm.c
@@ -236,6 +236,8 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
         sleep(interval);
 
         if (wait_time > args->timeout) {
+	    fprintf(stderr,
+		    "Exceeding timeout while waiting for servers to start.\n");
             ret = UNIFYFS_FAILURE;
             break;
         }
@@ -483,58 +485,120 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
 static size_t construct_server_argv(unifyfs_args_t* args,
                                     char** server_argv)
 {
-    size_t argc;
+    size_t argc = 0;
     char number[16];
 
-    if (server_argv != NULL) {
-        if (args->server_path != NULL) {
-            server_argv[0] = strdup(args->server_path);
-        } else {
-            server_argv[0] = strdup(BINDIR "/unifyfsd");
-        }
-    }
-    argc = 1;
-
-    if (args->debug) {
+    switch (args->arg_type) {
+    case UNIFYFS_CONSTRUCT_ARGS_UNIFYFSD:
         if (server_argv != NULL) {
-            server_argv[argc] = strdup("-v");
-            snprintf(number, sizeof(number), "%d", args->debug);
-            server_argv[argc + 1] = strdup(number);
+            if (args->server_path != NULL) {
+                server_argv[0] = strdup(args->server_path);
+            } else {
+                server_argv[0] = strdup(BINDIR "/unifyfsd");
+            }
         }
-        argc += 2;
-    }
+        argc = 1;
 
-    if (args->cleanup) {
+	if (args->debug) {
+            if (server_argv != NULL) {
+	        server_argv[argc] = strdup("-v");
+		snprintf(number, sizeof(number), "%d", args->debug);
+		server_argv[argc + 1] = strdup(number);
+            }
+	    argc += 2;
+	}
+
+	if (args->cleanup) {
+            if (server_argv != NULL) {
+	        server_argv[argc] = strdup("-C");
+            }
+	    argc++;
+	}
+
+	if (args->consistency != UNIFYFS_CM_LAMINATED) {
+            if (server_argv != NULL) {
+	        server_argv[argc] = strdup("-c");
+		server_argv[argc + 1] =
+		    strdup(unifyfs_cm_enum_str(args->consistency));
+            }
+	    argc += 2;
+	}
+
+	if (args->mountpoint != NULL) {
+            if (server_argv != NULL) {
+	        server_argv[argc] = strdup("-m");
+                server_argv[argc + 1] = strdup(args->mountpoint);
+            }
+	    argc += 2;
+	}
+
+	if (server_argv != NULL) {
+	    server_argv[argc] = strdup("-S");
+	    server_argv[argc + 1] = strdup(args->share_dir);
+	    server_argv[argc + 2] = strdup("-H");
+	    server_argv[argc + 3] = strdup(args->share_hostfile);
+	    argc += 4;
+	}
+	break;
+    case UNIFYFS_CONSTRUCT_ARGS_STAGE_IN:
         if (server_argv != NULL) {
-            server_argv[argc] = strdup("-C");
-        }
-        argc++;
-    }
+            if (args->transfer_exe == NULL) {
+	        fprintf(stderr,
+			"%s: args->transfer_exe is NULL!\n",
+			__func__);
+                return -1;
+            }
+            if (args->stage_in == NULL) {
+                fprintf(stderr, "%s: args->stage_in is NULL!\n",
+			__func__);
+	        return -1;
+            }
 
-    if (args->consistency != UNIFYFS_CM_LAMINATED) {
+	    //	    if (args->server_path != NULL) {
+	    server_argv[0] = strdup(args->transfer_exe);
+	    //    } else {
+	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	    }
+	    argc = 1;
+
+	    // if we need to add arguments betwen the command and the
+	    // manifest file name they get added here.
+
+	    server_argv[argc] = strdup(args->stage_in);
+	    argc++;
+	}
+	break;
+    case UNIFYFS_CONSTRUCT_ARGS_STAGE_OUT:
         if (server_argv != NULL) {
-            server_argv[argc] = strdup("-c");
-            server_argv[argc + 1] =
-                strdup(unifyfs_cm_enum_str(args->consistency));
-        }
-        argc += 2;
-    }
+            if (args->transfer_exe == NULL) {
+	        fprintf(stderr, "%s: args->transfer_exe is NULL!\n",
+			__func__);
+	        return -1;
+            }
+            if (args->stage_in == NULL) {
+	        fprintf(stderr, "%s: args->stage_out is NULL!\n",
+			__func__);
+	        return -1;
+            }
 
-    if (args->mountpoint != NULL) {
-        if (server_argv != NULL) {
-            server_argv[argc] = strdup("-m");
-            server_argv[argc + 1] = strdup(args->mountpoint);
-        }
-        argc += 2;
-    }
+	    //	    if (args->server_path != NULL) {
+	    server_argv[0] = strdup(args->transfer_exe);
+	    //    } else {
+	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	    }
+	    argc = 1;
 
-    if (server_argv != NULL) {
-        server_argv[argc] = strdup("-S");
-        server_argv[argc + 1] = strdup(args->share_dir);
-        server_argv[argc + 2] = strdup("-H");
-        server_argv[argc + 3] = strdup(args->share_hostfile);
+	    // if we need to add arguments betwen the command and the
+	    // manifest file name they get added here.
+
+	    server_argv[argc] = strdup(args->stage_out);
+	    argc++;
+	}
+	break;
+    default:
+        fprintf(stderr, "%s: Invalid args->arg_type!\n", __func__);
+        return -1;
     }
-    argc += 4;
 
     return argc;
 }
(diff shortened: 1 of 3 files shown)


craigsteffen:
> My suggestion would be to integrate stage-in functionality within `unifyfs_start_servers()` and stage-out within `unifyfs_stop_servers()`. The manifest file passed on the command line will be available from cli_args.

Yup.  I'm wending my way through the source code now, and I now see that the comments I put in are, as you say, too early in the logic flow.  

Creating the branch and the PR was more a placeholder than anything to make sure that Hyogi and I both have a place to start working on it, but now I'm looking more carefully about where to put stuff.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -807,11 +808,27 @@ int unifyfs_start_servers(unifyfs_resource_t* resource,
...
     }
+
+    // here we will check the status file that all the servers are
+    // up and talking to each other and are happy.  
+
+    // file stage-in goes here.  
+    // grab the name of the stage-in manifest file.  
+    // check that file for existence, parse it for proper structure
+    // check the files that it refers to
+    // then call the transfer API to transfer all the files
```
MichaelBrim, line 826: Note that you can't use the transfer API in the `unifyfs` utility. It may run on a node where the server is not running (for current LSF systems, it runs on a 'batch' node that is not a 'compute' node, which is where the servers are launched). Talk to Hyogi about my suggestion for this.

craigsteffen: Right, that's a good point.  We'll probably do the parsing locally, read the manifest, check it for validity, and create a set of marching orders corresponding to *each* valid file in the manifest that we want to transfer in.  

For each such file, we'll need to construct a proper jsrun (or appropriate equivalent) command that will fire off the transfer-static (or whatever) binary that, as you say, will then be launched on the *compute* nodes, where the servers live.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -808,10 +810,50 @@ int unifyfs_start_servers(unifyfs_resource_t* resource,
...
+        return_val=resource_managers[resource->rm].launch(resource, args);
     }
+
+    // here we will check the status file that all the servers are
+    // up and talking to each other and are happy.
+
+    // check for a --stage-out= argument.  If there is one, throw
+    // a warning that a --stage-out argument with a server launch
+    // doesn't make any sense and is not supported at this time.
+    if(args->stage_out != NULL){
```
MichaelBrim, line 824: All CLI arg checking should be done in `unifyfs.c:parse_cmd_arguments()`. That includes checking for existence of the provided manifest file.

craigsteffen:
Ok.  I'll change it tomorrow.

util/unifyfs/src/unifyfs.c
```diff
@@ -99,8 +99,8 @@ static char* usage_str =
     "  -s, --script=<path>       [OPTIONAL] <path> to custom launch script\n"
     "  -S, --share-dir=<path>    [REQUIRED] shared file system <path> for use by servers\n"
     "  -c, --cleanup             [OPTIONAL] clean up the UnifyFS storage upon server exit\n"
-    "  -i, --stage-in=<path>     [OPTIONAL, NOT YET SUPPORTED] stage in file(s) at <path>\n"
-    "  -o, --stage-out=<path>    [OPTIONAL, NOT YET SUPPORTED] stage out file(s) to <path> on termination\n"
+    "  -i, --stage-in=<path>     [OPTIONAL] stage in manifest file(s) at <path>\n"
+    "  -o, --stage-out=<path>    [OPTIONAL] stage out manifest file(s) at <path>\n"
```
sandrain, line 103: I think the --stage-out option has to go to options for the terminate command below.

craigsteffen: Oh right.  Good point.

util/unifyfs/src/unifyfs.c
```diff
@@ -92,6 +96,9 @@ static char* usage_str =
     "Common options:\n"
     "  -d, --debug               enable debug output\n"
     "  -h, --help                print usage\n"
+    "  -t, --timeout=<time>      [OPTIONAL] timeout value\n"
```
CamStan, line 99: This option is already on line 108 as an option for the start command. We need it twice?

craigsteffen:
Oops.  No, I must have just missed that.  I'll remove that change when I
squash all these together, which will happen as soon as I get some things
tested.

craigsteffen:
This is still WIP.  

My part is done (until we test it) but we're going to add Hyogi's piece to this branch so that we can test them together, then this will be made into the PR for the actual merge and milestone item.

util/unifyfs/src/unifyfs.c
```diff
@@ -76,6 +79,7 @@ static struct option const long_opts[] = {
     { "stage-in", required_argument, NULL, 'i' },
     { "stage-out", required_argument, NULL, 'o' },
     { "timeout", required_argument, NULL, 't' },
+    { "transfer-exe", required_argument, NULL, 'x' },
```
MichaelBrim, line 82: @craigsteffen you forgot to add 'x:' to short_opts

util/unifyfs/src/unifyfs.c
```diff
@@ -258,12 +279,176 @@ int main(int argc, char** argv)
...
         fprintf(stderr, "INTERNAL ERROR: unhandled action %d\n", (int)action);
         return -1;
     }
+} // int main(...
+
+
+// validates that the named manifest file exists and is readable
+// and returns the number of lines in it.
+// negative return value for error.
+int unifyfs_lines_in_manifest_file(char* manifest_filename)
```
MichaelBrim, line 391: I have no clue why this is necessary. Please explain the motivation.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,120 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
+    switch (args->arg_type) {
+    case UNIFYFS_CONSTRUCT_ARGS_UNIFYFSD:
         if (server_argv != NULL) {
-            server_argv[argc] = strdup("-v");
-            snprintf(number, sizeof(number), "%d", args->debug);
-            server_argv[argc + 1] = strdup(number);
+            if (args->server_path != NULL) {
+                server_argv[0] = strdup(args->server_path);
+            } else {
+                server_argv[0] = strdup(BINDIR "/unifyfsd");
```
tonyhutter, line 497: Should we be hard coding the path to the `unifyfsd` here?

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,120 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
 
-    if (args->mountpoint != NULL) {
+	    //	    if (args->server_path != NULL) {
+	    server_argv[0] = strdup(args->transfer_exe);
+	    //    } else {
+	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	    }
+	    argc = 1;
+
+	    // if we need to add arguments betwen the command and the
```
tonyhutter, line 564: "between"

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,120 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
+			__func__);
+	        return -1;
+            }
 
-    if (args->mountpoint != NULL) {
+	    //	    if (args->server_path != NULL) {
+	    server_argv[0] = strdup(args->transfer_exe);
+	    //    } else {
+	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	    }
```
tonyhutter, line 561: Remove debug comments

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,120 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
-        server_argv[argc] = strdup("-S");
-        server_argv[argc + 1] = strdup(args->share_dir);
-        server_argv[argc + 2] = strdup("-H");
-        server_argv[argc + 3] = strdup(args->share_hostfile);
+	    //	    if (args->server_path != NULL) {
+	    server_argv[0] = strdup(args->transfer_exe);
+	    //    } else {
+	    //	        server_argv[0] = strdup(BINDIR "/unifyfsd");
+	    //	    }
+	    argc = 1;
```
tonyhutter, line 589: Remove debug comments

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,120 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
+
+	    // if we need to add arguments betwen the command and the
```
tonyhutter, line 591: "between"

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -922,30 +988,79 @@ int unifyfs_start_servers(unifyfs_resource_t* resource,
...
+        transfer_return_val =
+	  resource_managers[resource->rm].launch(resource, args);
+	// If there's any possibility that the stage-in happens asynchronously,
+	// here's where we would put the check to wait until it was all done.
     }
 
-    return rc;
+    // Possibly we should combine the two return values in the future?
+    transfer_return_val++;
+    transfer_return_val--;
```
tonyhutter, line 1031: Can these be removed?

util/unifyfs/src/unifyfs.c
craigsteffen: Indeed.  Thanks.  I'll roll that in with my next updated.

util/unifyfs/src/unifyfs.c
craigsteffen: I think it's prudent to check the manifest file for existence and to check its structure as much as is reasonable early on in the flow of unifyfs so that if it's clearly pathological, the user will have some indication that something's wrong.  If, for instance, the transfer command that does the stage-in finds something bad in the manifest file, I don't know that it has as good error reporting path to the user.  

Personally I think the positive confirmation that the front end found the manifest file and that it contains X lines is a nice confirmation that the unifyfs command got the --stage-in (or --stage-out) flag, found everything it was expecting, and reports that everything is fine.  This is a single-line, relatively unobtrusive way of reporting all of that along with a quantitative diagnostic value to match against what was expected.  

If you think we should wrap this in a if(debug) statement unless it find something pathological, then I suppose we could do that.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: No idea.  I didn't do that.  It was that way.  Mike talked about this on the call on Monday, that if the unifyfsd isn't specified by flag, there's a default location for it.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Fixed in the next update.

util/unifyfs/src/unifyfs.c
MichaelBrim: I think we should just check for file existence at this point. Calling out to 'wc -l' only tells you that we got a file with some number of lines in it, it doesn't validate the contents in any way. I could pass a file full of '\n' characters and it would be happy. The format checking/parsing should be left to the transfer utility. If it detects a problem, it will generate a message to stderr, which is the same as what the unifyfs utility does for errors.

util/unifyfs/src/unifyfs-rm.c
MichaelBrim: Yes, the BINDIR path to the server is only used as a default. If the user gives a path on the command-line, that's the 'if' case.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Yes.  Sorry.  I put them in to get the compiler to stop whining about an unused variable.  I meant to fix that before I uploaded.  I'll fix that in the next commit.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Fixed both in next commit.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: These aren't debug comments.  They're how the code will work once the stage-in/stage-out infrastructure is stable enough that we can reasonably count on a transfer-stage binary being in the right place.  Admittedly I didn't change the names in them for this, which I should have.  

My next commit has the names change as to how it will be implemented, and has explanatory accompanying comments explaining what it is and why it's there.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Explanation in the other thread.  Same reason.

util/unifyfs/src/unifyfs.c
craigsteffen: Ok.  I wanted to do as much validation as possible at the front end, and you asked for all the validation to be done in unifyfs.c, not farther down, so I did as much as I could without actually parsing the file.  

As a user, if I have, say, a script that generates a manifest file from a directory for stage-in, and I'm expecting it to have 10 files, then it would be useful to me to have an initial debugging output that tells me "found 0 lines" or "found 5000 lines".  But if you think that manifest file validation should only be done in the transfer application, then I'll take this out once we've tested the unifyfs command and the transfer-stage command together.  I won't be in the next commit but it will be in the commit when we put in the final PR.

craigsteffen:
Oops.  I forgot to fix the ++/-- thing.  I'll do that now.

util/unifyfs/src/unifyfs-rm.c
tonyhutter: What if the user has `unifyfsd` in their `$PATH`?

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -922,30 +1000,76 @@ int unifyfs_start_servers(unifyfs_resource_t* resource,
...
+	      resource_managers[resource->rm].launch(resource, args);
+        }
+    } else {
+        unifyfsd_return_val = wait_server_initialization(resource, args);
+        if (unifyfsd_return_val) {
+	    fprintf(stderr,
+		    "ERROR: failed to wait for server initialization: (%d s)\n",
+		    unifyfsd_return_val);
         }
     }
```
tonyhutter, line 1021: We should be checking for a valid `unifyfsd_return_val` value here, and bailing if it is bad.  That will also allow you to get rid of the summation of return values at the end of the function:
```
return unifyfsd_return_val + transfer_return_val;
```

util/unifyfs/src/unifyfs.c
```diff
@@ -170,15 +177,20 @@ static void parse_cmd_arguments(int argc, char** argv)
...
 
         case 'o':
-            printf("WARNING: stage-out not yet supported!\n");
+	  //            printf("WARNING: stage-out not yet supported!\n");
             stage_out = strdup(optarg);
             break;
 
+	case 'x':
+            transfer_exe = strdup(optarg);
+	    //	    printf("specified transfer exec: >%s<\n",transfer_exe);
```
tonyhutter, line 191: Remove commented out printfs.

util/unifyfs/src/unifyfs.c
```diff
@@ -258,12 +279,175 @@ int main(int argc, char** argv)
...
+	        return -EINVAL;
+            }
+	    trans_exe_fp = fopen(cli_args.transfer_exe, "r");
+            if (trans_exe_fp == NULL) {
+	        fprintf(stderr,
+			"ERROR!  The trans exe file you specified >%s<doesn't exist!\n",
+			cli_args.transfer_exe);
+		return -ENOENT;
+            }
+	    fclose(trans_exe_fp);
```
tonyhutter, line 310: Remove the fopen()/fclose() and replace with:
```C
if (access(cli_args.transfer_exe, R_OK) == 0)
```

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,132 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
+	    argc++;
+	}
+
+	if (args->consistency != UNIFYFS_CM_LAMINATED) {
+            if (server_argv != NULL) {
+	        server_argv[argc] = strdup("-c");
+		server_argv[argc + 1] =
+		    strdup(unifyfs_cm_enum_str(args->consistency));
+            }
+	    argc += 2;
```
tonyhutter, line 524: Check you indents.  There's a couple of places where you have mixed tabs/spaces.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -483,58 +485,132 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
 static size_t construct_server_argv(unifyfs_args_t* args,
                                     char** server_argv)
 {
-    size_t argc;
+    size_t argc = 0;
     char number[16];
```
tonyhutter, line 490: Add a single `if (server_argv != NULL)` check here so you don't have to do the check so many times within the function.

util/unifyfs/src/unifyfs-rm.c
CamStan: Not sure this is related as I'm still lacking some experience here, but is what @tonyhutter is saying similar to what Spack does? Doing the commands:
```
spack install unifyfs
spack load unifyfs
```
adds the appropriate directories to `PATH`, `MANPATH`, `CPATH`, and LD_LIBRARY_PATH`.

As a user, after doing `spack load unifyfs`, the commands `unifyfs` and `unifyfsd` are recognized (i.e., can simply type `unifyfs start -h` or `unifyfsd -h`).

craigsteffen:
This work is being instead folded into PR 511.

PR #511, opened 2020-05-22
unifyfs-stage utility for data staging. 

sandrain:
this utility is expected to be used by unifyfs cli to handle --stage-in and --stage-out options.

- util/unifyfs-stage: unifyfs-stage utility
- unifyfs.c: some fix on unifyfs_transfer_parallel
- configure update for checking openssl-devel package (for md5sum)

### Motivation and Context
`unifyfs-stage` is a new utility that will be internally used for supporting data stage-in and stage-out. #504 depends on this.

EDIT: I've added a commit which modifies the `unifyfs` utility.

### How Has This Been Tested?
Tested on Summit (GPFS) and some x86 cluster with NFS. The serial transfer (file per process, each process transfers a whole file) works fine (including checksum verification).

### Types of changes

### DIFF
--- a/t/0700-unifyfs-stage-full.t
+++ b/t/0700-unifyfs-stage-full.t
@@ -1,0 +1,64 @@
+#!/bin/bash
+#
+# Test unifyfs-stage executable for basic functionality
+#
+
+test_description="Test basic functionality of unifyfs-stage executable"
+
+. $(dirname $0)/sharness.sh
+
+test_expect_success "unifyfs-stage exists" '
+    test_path_is_file ${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage
+'
+test_expect_success "testing temp dir exists" '
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}
+'
+
+mkdir -p ${UNIFYFS_TEST_TMPDIR}/config_0700
+mkdir -p ${UNIFYFS_TEST_TMPDIR}/stage_source
+mkdir -p ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700
+
+test_expect_success "stage testing dirs exist" '
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}/config_0700
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}/stage_source
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700
+'
+
+dd if=/dev/urandom bs=4M count=1 of=${UNIFYFS_TEST_TMPDIR}/stage_source/source_0700.file
+
+test_expect_success "source.file exists" '
+    test_path_is_file ${UNIFYFS_TEST_TMPDIR}/stage_source/source_0700.file
+'
+
+rm -f ${UNIFYFS_TEST_TMPDIR}/config_0700/*
+rm -f ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/*
+
+test_expect_success "config_0700 directory is empty" '
+    test_dir_is_empty ${UNIFYFS_TEST_TMPDIR}/config_0700
+'
+
+echo "\"${UNIFYFS_TEST_TMPDIR}/stage_source/source_0700.file\" \"${UNIFYFS_TEST_MOUNT}/intermediate.file\""  > ${UNIFYFS_TEST_TMPDIR}/config_0700/test_IN.manifest
+echo "\"${UNIFYFS_TEST_MOUNT}/intermediate.file\" \"${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/destination_0700.file\"" > ${UNIFYFS_TEST_TMPDIR}/config_0700/test_OUT.manifest
+
+test_expect_success "config_0700 directory now has manifest files" '
+    test_path_is_file  ${UNIFYFS_TEST_TMPDIR}/config_0700/test_IN.manifest
+    test_path_is_file  ${UNIFYFS_TEST_TMPDIR}/config_0700/test_OUT.manifest
+'
+
+test_expect_success "target directory is empty" '
+    test_dir_is_empty ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700
+'
+
+${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage -m ${UNIFYFS_TEST_MOUNT} ${UNIFYFS_TEST_TMPDIR}/config_0700/test_IN.manifest > ${UNIFYFS_TEST_TMPDIR}/config_0700/stage_IN_output.OUT 2>&1
+
+${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage -m ${UNIFYFS_TEST_MOUNT} ${UNIFYFS_TEST_TMPDIR}/config_0700/test_OUT.manifest > ${UNIFYFS_TEST_TMPDIR}/config_0700/stage_OUT_output.OUT 2>&1
+
+test_expect_success "input file has been staged to output" '
+    test_path_is_file ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/destination_0700.file
+'
+
+test_expect_success "final output is identical to initial input" '
+    test_might_fail test_cmp ${UNIFYFS_TEST_TMPDIR}/stage_source/source_0700.file ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/destination_0700.file
+'
+
+test_done
--- a/t/9300-unifyfs-stage-isolated.t
+++ b/t/9300-unifyfs-stage-isolated.t
@@ -1,0 +1,62 @@
+#!/bin/bash
+#
+# Test unifyfs-stage executable for basic functionality
+#
+
+test_description="Test basic functionality of unifyfs-stage executable"
+
+. $(dirname $0)/sharness.sh
+
+test_expect_success "unifyfs-stage exists" '
+    test_path_is_file ${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage
+'
+test_expect_success "testing temp dir exists" '
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}
+'
+
+mkdir -p ${UNIFYFS_TEST_TMPDIR}/config_9300
+mkdir -p ${UNIFYFS_TEST_TMPDIR}/stage_source
+mkdir -p ${UNIFYFS_TEST_TMPDIR}/stage_destination_9300
+
+test_expect_success "stage testing dirs exist" '
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}/config_9300
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}/stage_source
+    test_path_is_dir  ${UNIFYFS_TEST_TMPDIR}/stage_destination_9300
+'
+
+# NOTE: we're using the unifyfs-stage binary as its own transfer data target
+# because we know it's there and it's filled with non-zero data.
+cp ${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage ${UNIFYFS_TEST_TMPDIR}/stage_source/source_9300.file
+
+test_expect_success "source.file exists" '
+    test_path_is_file ${UNIFYFS_TEST_TMPDIR}/stage_source/source_9300.file
+'
+
+rm - f $ {UNIFYFS_TEST_TMPDIR} / config_9300/*
+rm -f ${UNIFYFS_TEST_TMPDIR}/stage_destination_9300/*
+
+test_expect_success "config_9300 directory is empty" '
+    test_dir_is_empty ${UNIFYFS_TEST_TMPDIR}/config_9300
+'
+
+echo "\"${UNIFYFS_TEST_TMPDIR}/stage_source/source_9300.file\" \"${UNIFYFS_TEST_TMPDIR}/stage_destination_9300/destination_9300.file\""  > ${UNIFYFS_TEST_TMPDIR}/config_9300/test_INOUT.manifest
+
+test_expect_success "config_9300 directory now has manifest files" '
+    test_path_is_file  ${UNIFYFS_TEST_TMPDIR}/config_9300/test_INOUT.manifest
+'
+
+test_expect_success "target directory is empty" '
+    test_dir_is_empty ${UNIFYFS_TEST_TMPDIR}/stage_destination_9300
+'
+
+${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage -N ${UNIFYFS_TEST_TMPDIR}/config_9300/test_INOUT.manifest > ${UNIFYFS_TEST_TMPDIR}/config_9300/stage_INOUT_output.OUT 2>&1
+
+test_expect_success "input file has been staged to output" '
+    test_path_is_file ${UNIFYFS_TEST_TMPDIR}/stage_destination_9300/destination_9300.file
+'
+
+test_expect_success "final output is identical to initial input" '
+    test_cmp ${UNIFYFS_TEST_TMPDIR}/stage_source/source_9300.file ${UNIFYFS_TEST_TMPDIR}/stage_destination_9300/destination_9300.file
+'
+
+test_done
--- a/util/unifyfs-stage/src/unifyfs-stage-transfer.c
+++ b/util/unifyfs-stage/src/unifyfs-stage-transfer.c
@@ -1,0 +1,403 @@
+/*
+ * Copyright (c) 2020, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2020, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyFS.
+ * For details, see https://github.com/LLNL/UnifyFS.
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
+#include <config.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <errno.h>
+#include <ctype.h>
+#include <limits.h>
+#include <sys/time.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <dirent.h>
+#include <libgen.h>
+#include <getopt.h>
+#include <time.h>
+#include <mpi.h>
+#include <openssl/md5.h>
+
+#include "unifyfs-stage.h"
+
+/**
+ * @brief Run md5 checksum on specified file, send back
+ *        digest.
+ *
+ * @param path      path to the target file
+ * @param digest    hash of the file
+ *
+ * @return 0 on success, errno otherwise
+ */
+static int md5_checksum(const char* path, unsigned char* digest)
+{
+    int ret = 0;
+    size_t len = 0;
+    int fd = -1;
+    unsigned char data[UNIFYFS_STAGE_MD5_BLOCKSIZE] = { 0, };
+    MD5_CTX md5;
+
+    fd = open(path, O_RDONLY);
+    if (fd < 0) {
+        perror("open");
+        return errno;
+    }
+
+    ret = MD5_Init(&md5);
+    if (!ret) {
+        fprintf(stderr, "failed to create md5 context\n");
+        goto out;
+    }
+
+    while ((len = read(fd, (void*) data, UNIFYFS_STAGE_MD5_BLOCKSIZE)) != 0) {
+        ret = MD5_Update(&md5, data, len);
+        if (!ret) {
+            fprintf(stderr, "failed to update checksum\n");
+            goto out;
+        }
+    }
+
+    ret = MD5_Final(digest, &md5);
+    if (!ret) {
+        fprintf(stderr, "failed to finalize md5\n");
+    }
+
+out:
+    /* MD5_xx returns 1 for success */
+    ret = ret == 1 ? 0 : EIO;
+    close(fd);
+
+    return ret;
+}
+
+/**
+ * @brief prints md5 checksum into string
+ *
+ * @param buf       buffer to print into
+ * @param digest    hash of the file
+ *
+ * @return buffer that has been printed to
+ */
+static char* checksum_str(char* buf, unsigned char* digest)
+{
+    int i = 0;
+    char* pos = buf;
+
+    for (i = 0; i < MD5_DIGEST_LENGTH; i++) {
+        pos += sprintf(pos, "%02x", digest[i]);
+    }
+
+    pos[0] = '\0';
+
+    return buf;
+}
+
+/**
+ * @brief takes check sums of two files and compares
+ *
+ * @param src     path to one file
+ * @param dst     path to the other file
+ *
+ * @return 0 if files are identical, non-zero if not, or other error
+ */
+static int verify_checksum(const char* src, const char* dst)
+{
+    int ret = 0;
+    int i = 0;
+    char md5src[2 * MD5_DIGEST_LENGTH + 1] = { 0, };
+    char md5dst[2 * MD5_DIGEST_LENGTH + 1] = { 0, };
+    unsigned char src_digest[MD5_DIGEST_LENGTH + 1] = { 0, };
+    unsigned char dst_digest[MD5_DIGEST_LENGTH + 1] = { 0, };
+
+    src_digest[MD5_DIGEST_LENGTH] = '\0';
+    dst_digest[MD5_DIGEST_LENGTH] = '\0';
+
+    ret = md5_checksum(src, src_digest);
+    if (ret) {
+        fprintf(stderr, "failed to calculate checksum for %s (%s)\n",
+                        src, strerror(ret));
+        return ret;
+    }
+
+    ret = md5_checksum(dst, dst_digest);
+    if (ret) {
+        fprintf(stderr, "failed to calculate checksum for %s (%s)\n",
+                        dst, strerror(ret));
+        return ret;
+    }
+
+    if (verbose) {
+        printf("[%d] src: %s, dst: %s\n", rank,
+               checksum_str(md5src, src_digest),
+               checksum_str(md5dst, dst_digest));
+    }
+
+    for (i = 0; i < MD5_DIGEST_LENGTH; i++) {
+        if (src_digest[i] != dst_digest[i]) {
+            fprintf(stderr, "[%d] checksum verification failed: "
+                            "(src=%s, dst=%s)\n", rank,
+                            checksum_str(md5src, src_digest),
+                            checksum_str(md5dst, dst_digest));
+            ret = EIO;
+        }
+    }
+
+    return ret;
+}
+
+/*
+ * Parse a line from the manifest in the form of:
+ *
+ * <src path> <whitespace separator> <dest path>
+ *
+ * If the paths have spaces, they must be quoted.
+ *
+ * On success, return 0 along with allocated src and dest strings.  These
+ * must be freed when you're finished with them.  On failure return non-zero,
+ * and set src and dest to NULL.
+ *
+ * Note, leading and tailing whitespace are ok.  They just get ignored.
+ * Lines with only whitespace are ignored.  A line of all whitespace will
+ * return 0, with src and dest being NULL, so users should not check for
+ * 'if (*src == NULL)' to see if the function failed.  They should be looking
+ * at the return code.
+ */
+/**
+ * @brief parses manifest file line, passes back src and dst strings
+ *
+ * @param line    input manifest file line
+ * @param src     return val of src filename
+ * @param dst     return val of dst filename
+ *
+ * @return 0 if all was well, or there was nothing; non-zero on error
+ */
+int
+unifyfs_parse_manifest_line(char* line, char** src, char** dest)
+{
+    char* new_src = NULL;
+    char* new_dest = NULL;
+    char* copy;
+    char* tmp;
+    unsigned long copy_len;
+    int i;
+    unsigned int tmp_count;
+    int in_quotes = 0;
+    int rc = 0;
+
+    copy = strdup(line);
+    copy_len = strlen(copy) + 1;/* +1 for '\0' */
+
+    /* Replace quotes and separator with '\0' */
+    for (i = 0; i < copy_len; i++) {
+        if (copy[i] == '"') {
+            in_quotes ^= 1;/* toggle */
+            copy[i] = '\0';
+        } else if (isspace(copy[i]) && !in_quotes) {
+            /*
+             * Allow any whitespace for our separator
+             */
+            copy[i] = '\0';
+        }
+    }
+
+    /*
+     * copy[] now contains a series of strings, one after the other
+     * (possibly containing some NULL strings, which we ignore)
+     */
+    tmp = copy;
+    while (tmp < copy + copy_len) {
+        tmp_count = strlen(tmp);
+        if (tmp_count > 0) {
+            /* We have a real string */
+            if (!new_src) {
+                new_src = strdup(tmp);
+            } else {
+                if (!new_dest) {
+                    new_dest = strdup(tmp);
+                } else {
+                    /* Error: a third file name */
+                    rc = 1;
+                    break;
+                }
+            }
+        }
+        tmp += tmp_count + 1;
+    }
+
+    /* Some kind of error parsing a line */
+    if (rc != 0 || (new_src && !new_dest)) {
+        fprintf(stderr, "manifest file line >>%s<< is invalid!\n",
+                line);
+        free(new_src);
+        free(new_dest);
+        new_src = NULL;
+        new_dest = NULL;
+        if (rc == 0) {
+            rc = 1;
+        }
+    }
+
+    *src = new_src;
+    *dest = new_dest;
+
+    free(copy);
+    return rc;
+}
+
+/**
+ * @brief controls the action of the stage-in or stage-out.  Opens up
+ *        the manifest file, sends each line to be parsed, and fires
+ *        each source/destination to be staged.
+ *
+ * @param ctx     stage context and instructions
+ *
+ * @return 0 indicates success, non-zero is error
+ */
+int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
+{
+    int ret = 0;
+    int count = 0;
+    FILE* fp = NULL;
+    char* src = NULL;
+    char* dst = NULL;
+    char linebuf[LINE_MAX] = { 0, };
+    struct stat sb = { 0, };
+
+    if (!ctx) {
+        return EINVAL;
+    }
+
+    fp = fopen(ctx->manifest_file, "r");
+    if (!fp) {
+        fprintf(stderr, "failed to open file %s: %s\n",
+		ctx->manifest_file, strerror(errno));
+        ret = errno;
+        goto out;
+    }
+
+    while (NULL != fgets(linebuf, LINE_MAX-1, fp)) {
+        if (strlen(linebuf) < 5) {
+	    // the manifest file perhaps ends with a couple of characters
+	    // and/or a newline not meant to be a transfer spec.
+            if (linebuf[0] == '\n') {
+	        goto out;
+            } else{
+	        fprintf(stderr, "Short (bad) manifest file line: >%s<\n",
+			linebuf);
+		ret = -EINVAL;
+                goto out;
+            }
+	}
+	ret = unifyfs_parse_manifest_line(linebuf, &src, &dst);
+	if (ret < 0) {
+	    fprintf(stderr, "failed to parse %s (%s)\n",
+		    linebuf, strerror(ret));
+            goto out;
+	}
+        if (ctx->mode == UNIFYFS_STAGE_SERIAL) {
+            if (count % total_ranks == rank) {
+                if (verbose) {
+                    fprintf(stdout, "[%d] serial transfer: src=%s, dst=%s\n",
+                                    rank, src, dst);
+                }
+
+                ret = unifyfs_transfer_file_serial(src, dst);
+                if (ret) {
+                    goto out;
+                }
+
+                if (ret < 0) {
+                    fprintf(stderr, "stat on %s failed (err=%d, %s)\n",
+                                    dst, errno, strerror(errno));
+                    ret = errno;
+                    goto out;
+                }
+
+                if (ctx->checksum) {
+                    ret = verify_checksum(src, dst);
+                    if (ret) {
+                        fprintf(stderr, "checksums for >%s< and >%s< differ!\n",
+                                src, dst);
+                        goto out;
+                    }
+                }
+            }
+        } else {
+            if (0 == rank) {
+                int fd = -1;
+
+                if (verbose) {
+                    fprintf(stdout, "[%d] parallel transfer: src=%s, dst=%s\n",
+                                    rank, src, dst);
+                }
+
+                /* FIXME: Since we cannot use the mpi barrier inside the
+                 * unifyfs_transfer_file_parallel(), we need to create the file
+                 * here before others start to access. It would be better if we
+                 * can skip this step.
+                 */
+                fd = open(dst, O_WRONLY|O_CREAT|O_TRUNC, 0600);
+                if (fd < 0) {
+                    fprintf(stderr, "[%d] failed to create the file %s\n",
+                                    rank, dst);
+                    goto out;
+                }
+
+                close(fd);
+            }
+
+            MPI_Barrier(MPI_COMM_WORLD);
+
+            ret = unifyfs_transfer_file_parallel(src, dst);
+            if (ret) {
+                goto out;
+            }
+
+            MPI_Barrier(MPI_COMM_WORLD);
+
+            ret = stat(dst, &sb);
+            if (ret < 0) {
+                fprintf(stderr, "stat on %s failed (err=%d, %s)\n",
+                                dst, errno, strerror(errno));
+                ret = errno;
+                goto out;
+            }
+
+            if (ctx->checksum && 0 == rank) {
+                ret = verify_checksum(src, dst);
+                if (ret) {
+                    goto out;
+                }
+            }
+        }
+
+        count++;
+    }
+out:
+    if (ret) {
+        fprintf(stderr, "failed to transfer file (src=%s, dst=%s): %s\n",
+                        src, dst, strerror(ret));
+    }
+
+    if (fp) {
+        fclose(fp);
+        fp = NULL;
+    }
+
+    return ret;
+}
+
--- a/util/unifyfs-stage/src/unifyfs-stage-transfer.c
+++ b/util/unifyfs-stage/src/unifyfs-stage-transfer.c
@@ -129,14 +129,14 @@ static int verify_checksum(const char* src, const char* dst)
     ret = md5_checksum(src, src_digest);
     if (ret) {
         fprintf(stderr, "failed to calculate checksum for %s (%s)\n",
-                        src, strerror(ret));
+                src, strerror(ret));
         return ret;
     }
 
     ret = md5_checksum(dst, dst_digest);
     if (ret) {
         fprintf(stderr, "failed to calculate checksum for %s (%s)\n",
-                        dst, strerror(ret));
+                dst, strerror(ret));
         return ret;
     }
 
@@ -149,9 +149,9 @@ static int verify_checksum(const char* src, const char* dst)
     for (i = 0; i < MD5_DIGEST_LENGTH; i++) {
         if (src_digest[i] != dst_digest[i]) {
             fprintf(stderr, "[%d] checksum verification failed: "
-                            "(src=%s, dst=%s)\n", rank,
-                            checksum_str(md5src, src_digest),
-                            checksum_str(md5dst, dst_digest));
+                    "(src=%s, dst=%s)\n", rank,
+                    checksum_str(md5src, src_digest),
+                    checksum_str(md5dst, dst_digest));
             ret = EIO;
         }
     }
@@ -275,7 +275,6 @@ int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
     char* src = NULL;
     char* dst = NULL;
     char linebuf[LINE_MAX] = { 0, };
-    struct stat sb = { 0, };
 
     if (!ctx) {
         return EINVAL;
@@ -284,35 +283,34 @@ int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
     fp = fopen(ctx->manifest_file, "r");
     if (!fp) {
         fprintf(stderr, "failed to open file %s: %s\n",
-		ctx->manifest_file, strerror(errno));
+                ctx->manifest_file, strerror(errno));
         ret = errno;
         goto out;
     }
 
-    while (NULL != fgets(linebuf, LINE_MAX-1, fp)) {
+    while (NULL != fgets(linebuf, LINE_MAX - 1, fp)) {
         if (strlen(linebuf) < 5) {
-	    // the manifest file perhaps ends with a couple of characters
-	    // and/or a newline not meant to be a transfer spec.
             if (linebuf[0] == '\n') {
-	        goto out;
-            } else{
-	        fprintf(stderr, "Short (bad) manifest file line: >%s<\n",
-			linebuf);
-		ret = -EINVAL;
+                // manifest file ends in a blank line
+                goto out;
+            } else {
+                fprintf(stderr, "Short (bad) manifest file line: >%s<\n",
+                        linebuf);
+                ret = -EINVAL;
                 goto out;
             }
-	}
-	ret = unifyfs_parse_manifest_line(linebuf, &src, &dst);
-	if (ret < 0) {
-	    fprintf(stderr, "failed to parse %s (%s)\n",
-		    linebuf, strerror(ret));
+        }
+        ret = unifyfs_parse_manifest_line(linebuf, &src, &dst);
+        if (ret < 0) {
+            fprintf(stderr, "failed to parse %s (%s)\n",
+                    linebuf, strerror(ret));
             goto out;
-	}
+        }
         if (ctx->mode == UNIFYFS_STAGE_SERIAL) {
             if (count % total_ranks == rank) {
                 if (verbose) {
                     fprintf(stdout, "[%d] serial transfer: src=%s, dst=%s\n",
-                                    rank, src, dst);
+                            rank, src, dst);
                 }
 
                 ret = unifyfs_transfer_file_serial(src, dst);
@@ -322,7 +320,7 @@ int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
 
                 if (ret < 0) {
                     fprintf(stderr, "stat on %s failed (err=%d, %s)\n",
-                                    dst, errno, strerror(errno));
+                            dst, errno, strerror(errno));
                     ret = errno;
                     goto out;
                 }
@@ -342,21 +340,15 @@ int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
 
                 if (verbose) {
                     fprintf(stdout, "[%d] parallel transfer: src=%s, dst=%s\n",
-                                    rank, src, dst);
+                            rank, src, dst);
                 }
 
-                /* FIXME: Since we cannot use the mpi barrier inside the
-                 * unifyfs_transfer_file_parallel(), we need to create the file
-                 * here before others start to access. It would be better if we
-                 * can skip this step.
-                 */
-                fd = open(dst, O_WRONLY|O_CREAT|O_TRUNC, 0600);
+                fd = open(dst, O_WRONLY | O_CREAT | O_TRUNC, 0600);
                 if (fd < 0) {
                     fprintf(stderr, "[%d] failed to create the file %s\n",
-                                    rank, dst);
+                            rank, dst);
                     goto out;
                 }
-
                 close(fd);
             }
 
@@ -369,13 +361,8 @@ int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
 
             MPI_Barrier(MPI_COMM_WORLD);
 
-            ret = stat(dst, &sb);
-            if (ret < 0) {
-                fprintf(stderr, "stat on %s failed (err=%d, %s)\n",
-                                dst, errno, strerror(errno));
-                ret = errno;
-                goto out;
-            }
+            // possible lamination check or force lamination
+            // may need to go here
 
             if (ctx->checksum && 0 == rank) {
                 ret = verify_checksum(src, dst);
@@ -390,7 +377,7 @@ int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
 out:
     if (ret) {
         fprintf(stderr, "failed to transfer file (src=%s, dst=%s): %s\n",
-                        src, dst, strerror(ret));
+                src, dst, strerror(ret));
     }
 
     if (fp) {
--- a/util/unifyfs-stage/src/unifyfs-stage.c
+++ b/util/unifyfs-stage/src/unifyfs-stage.c
@@ -1,0 +1,300 @@
+/*
+ * Copyright (c) 2020, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2020, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyFS.
+ * For details, see https://github.com/LLNL/UnifyFS.
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
+/* unifyfs-stage: this application is supposed to excuted by the unifyfs
+ * command line utility for:
+ * - stage in: moving files in pfs to unifyfs volume before user starts
+ *   application,
+ *   e.g., unifyfs start --stage-in=<manifest file>
+ * - stage out: moving files in the unifyfs volume to parallel file system
+ *   after user application completes,
+ *   e.g., unifyfs terminate --stage-out=<manifest file>
+ *
+ * Currently, we request users to pass the <manifest file> to specify target
+ * files to be transferred. The <manifest file> should list all target files
+ * and their destinations, line by line.
+ *
+ * This supports two transfer modes (although both are technically parallel):
+ *
+ * - serial: Each process will transfer a file. Data of a single file will
+ *   reside in a single compute node.
+ * - parallel (-p, --parallel): Each file will be split and transferred by all
+ *   processes. Data of a single file will be spread evenly across all
+ *   available compute nodes.
+ *
+ * TODO:
+ * Maybe later on, it would be better to have a size threshold. Based on the
+ * threshold, we can determine whether a file needs to transferred serially (if
+ * smaller than threshold), or parallelly.
+ */
+#include <config.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <getopt.h>
+#include <errno.h>
+#include <limits.h>
+#include <libgen.h>
+#include <mpi.h>
+#include <unifyfs_misc.h>
+
+#include "unifyfs_const.h"
+#include "unifyfs-stage.h"
+
+int rank;
+int total_ranks;
+int verbose;
+
+static int debug;
+static int checksum;
+static int mode;
+static char* manifest_file;
+static char* mountpoint = "/unifyfs";
+static char* share_dir;
+
+static unifyfs_stage_t _ctx;
+
+/**
+ * @brief create a status (lock) file to notify the unifyfs executable
+ * when the staging is finished
+ *
+ * @param status 0 indicates success
+ *
+ * @return 0 on success, errno otherwise
+ */
+static int create_status_file(int status)
+{
+    char filename[PATH_MAX];
+    FILE* fp = NULL;
+    const char* msg = status ? "fail" : "success";
+    int return_val_from_scnprintf;
+
+    return_val_from_scnprintf =
+        scnprintf(filename, PATH_MAX,
+		  "%s/%s", share_dir, UNIFYFS_STAGE_STATUS_FILENAME);
+    if (return_val_from_scnprintf > (PATH_MAX-1)) {
+        fprintf(stderr, "Stage status file is too long!\n");
+        return -ENOMEM;
+    }
+
+    fp = fopen(filename, "w");
+    if (!fp) {
+        fprintf(stderr, "failed to create %s (%s)\n",
+                        filename, strerror(errno));
+        return errno;
+    }
+
+    fprintf(fp, "%s\n", msg);
+
+    fclose(fp);
+
+    return 0;
+}
+
+static struct option long_opts[] = {
+    { "checksum", 0, 0, 'c' },
+    { "debug", 0, 0, 'd' },
+    { "help", 0, 0, 'h' },
+    { "mountpoint", 1, 0, 'm' },
+    { "parallel", 0, 0, 'p' },
+    { "share-dir", 1, 0, 's' },
+    { "verbose", 0, 0, 'v' },
+    { 0, 0, 0, 0 },
+};
+
+static char* short_opts = "cdhm:ps:v";
+
+static const char* usage_str =
+    "\n"
+    "Usage: %s [OPTION]... <manifest file>\n"
+    "\n"
+    "Transfer files between unifyfs volume and external file system.\n"
+    "The <manifest file> should contain list of files to be transferred,\n"
+    "and each line should be formatted as\n"
+    "\n"
+    "  /source/file/path,/destination/file/path\n"
+    "\n"
+    "Specifying directories is not supported.\n"
+    "\n"
+    "Available options:\n"
+    "\n"
+    "  -c, --checksum           verify md5 checksum for each transfer\n"
+    "  -h, --help               print this usage\n"
+    "  -m, --mountpoint=<mnt>   use <mnt> as unifyfs mountpoint\n"
+    "                           (default: /unifyfs)\n"
+    "  -p, --parallel           transfer each file in parallel\n"
+    "                           (experimental)\n"
+    "  -s, --share-dir=<path>   directory path for creating status file\n"
+    "  -v, --verbose            print noisy outputs\n"
+    "\n"
+    "Without the '-p, --parallel' option, a file is transferred by a single\n"
+    "process. If the '-p, --parallel' option is specified, each file will be\n"
+    "divided by multiple processes and transferred in parallel.\n"
+    "\n";
+
+static char* program;
+
+static void print_usage(void)
+{
+    if (0 == rank) {
+        fprintf(stdout, usage_str, program);
+    }
+}
+
+static inline
+void debug_pause(int rank, const char* fmt, ...)
+{
+    if (rank == 0) {
+        va_list args;
+
+        va_start(args, fmt);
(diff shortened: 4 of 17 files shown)


util/unifyfs/src/unifyfs-rm.c
```diff
@@ -244,6 +248,98 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
...
+ *
+ * @return
+ */
+
+enum {
+    UNIFYFS_STAGE_IN = 0,
+    UNIFYFS_STAGE_OUT = 1,
+};
+
+static inline unsigned int estimate_timeout(const char* manifest_file, int op)
```
MichaelBrim, line 265: Should we allow the user to specify a good timeout for the file stage?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,294 @@
...
+        return errno;
+    }
+
+    ret = MD5_Init(&md5);
+    if (!ret) {
+        fprintf(stderr, "failed to create md5 context\n");
+        goto out;
+    }
+
+    while ((len = read(fd, (void*) data, 1024)) != 0) {
```
tonyhutter, line 57: Did you try benchmarking this with a larger chunk size?  1024 is pretty small.  Maybe try 1MB?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,294 @@
...
+#include <dirent.h>
+#include <libgen.h>
+#include <getopt.h>
+#include <time.h>
+#include <mpi.h>
+#include <openssl/md5.h>
+
+#include "unifyfs-stage.h"
+
+static int md5_checksum(const char* path, unsigned char* digest)
```
tonyhutter, line 37: None of the functions in this file have documentation.  Please add a quick comment above each one describing what they do and what the args/return values mean.

craigsteffen:
Tony,

I think Hyogi is turning further work on this over to me; I'll talk to him
on the phone tomorrow about this.  I'll take care of documenting the
functions.

Craig

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,292 @@
...
+
+static const char* usage_str =
+    "\n"
+    "Usage: %s [OPTION]... <manifest file>\n"
+    "\n"
+    "Transfer files between unifyfs volume and external file system.\n"
+    "The <manifest file> should contain list of files to be transferred,\n"
+    "and each line should be formatted as\n"
+    "\n"
+    "  /source/file/path,/destination/file/path\n"
```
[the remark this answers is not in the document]
sandrain: @MichaelBrim Nothing wrong with a space. I did this for avoiding test errors (mistakenly putting two spaces). We can switch to a space or any number of spaces and put extra code for parsing those tokens correctly.

util/unifyfs/src/unifyfs-rm.c
sandrain: I have actually no idea on this. It's relatively easy for stage-in, because all files already exist. But for stage-out, maybe even some users won't have clear idea about how long it may take.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,294 @@
...
+                    goto out;
+                }
+
+                ret = stat(dst, &sb);
+                if (ret < 0) {
+                    fprintf(stderr, "stat on %s failed (err=%d, %s)\n",
+                                    dst, errno, strerror(errno));
+                    ret = errno;
+                    goto out;
+                }
```
tonyhutter, line 221: This `stat()` isn't necessary.  If the file isn't there, it's going to be caught in `verify_checksum()`.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: In fact, this is really slow for large files (tested up to 32GB). In addition to the buffer size, currently it's done by a single process/thread. I will change this to use a larger buffer, at least 1MB as you suggested.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: @tonyhutter Correct for regular file systems. But for unifyfs it seems necessary to have the file correctly laminated. Someone actually needs to confirm this.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: Thanks, @tonyhutter. I will include comments.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: I'm a little confused.  Are you saying the `stat()` helps to laminate the file after it's transferred?

util/unifyfs-stage/src/unifyfs-stage.c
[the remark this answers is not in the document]
tonyhutter: I like the spaces better too.  You will need to add addition logic to parse filenames with spaces though (either escaping the spaces `'\ '` or quoting the filename).

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: Honestly, I am a bit confused as well. I've had some problems with laminating files in my test setup. It seems like we need to do this for N-1 files, according to the write example:

https://github.com/LLNL/UnifyFS/blob/0c9eb9ab10a9008ddb1789efebe62e10ab7a6eac/examples/src/testutil_rdwr.h#L236

```

    if (cfg->io_pattern == IO_PATTERN_N1) {
        test_barrier(cfg);
        if (cfg->rank != 0) {
            /* call stat() to update global metadata */
            struct stat st;
            int stat_rc = stat(filepath, &st);
            if (-1 == stat_rc) {
                /* lamination failed */
                test_print(cfg, "stat() update during lamination failed");
                rc = -1;
            }
        }

```

I haven't had enough time for figuring out this.

craigsteffen:
We had to add lamination for the transfer application a while back by using
chmod.  Would that work here?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: You shouldn't need to do anything special.  Your current callpath will laminate the file:
```
unifyfs_transfer_file_serial()
    unifyfs_transfer_file()
        chmod(0222)
```
If for some reason the files aren't laminated, you could try calling `sync()` after they're all transferred.

sandrain: I looked the current client code, and it looks like the `unifyfs_fid_sync` function is the one that sends all cached extent information down to the server daemon.

https://github.com/LLNL/UnifyFS/blob/0c9eb9ab10a9008ddb1789efebe62e10ab7a6eac/client/src/unifyfs.c#L1700

This function is called from multiple system calls including `stat(2)`. Technically, `stat(2)` here doesn't look necessary, because `unifyfs_fid_sync` should be invoked by `close(2)` and `fsync(2)` as well.

I will remove the `stat(2)` here and verify if it still works. Also, one thing we can do here is we can see if the dest file size (by stat) is identical to the source file size. If the size is different, we don't have to call the expensive checksum calculation which will obviously fail.

sandrain: And, the lamination process still confuses me. Basically, the file should be laminated before accessing (reading data) from all nodes. But simply closing a file will not laminate the file. Also, laminating a file won't push the unwritten metadata to the server. Further, you can laminate a currently opened file with `fchmod(2)`. Besides the fact we laminate a file with chmod (instead of having a dedicated API), I don't know if all those things are intuitive enough.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,294 @@
...
+    pos[0] = '\0';
+
+    return buf;
+}
+
+static int verify_checksum(const char* src, const char* dst)
+{
+    int ret = 0;
+    int i = 0;
+    char md5src[36] = { 0, };
```
MichaelBrim, line 96: hard-coded 36 here is bound to cause problems. use UNIFYFS_MAX_FILENAME for all path buffers

MichaelBrim, line 96: Why hard-coded 36? Shouldn't this be (2 * MD5_DIGEST_LEN + 1) based on the code in `checksum_str()`

util/unifyfs-stage/src/unifyfs-stage.c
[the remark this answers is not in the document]
MichaelBrim: Technically, POSIX portable pathnames (https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282) don't allow spaces, and we could forbid them inside the UnifyFS namespace. But, we would still need to accommodate the external filenames that include spaces.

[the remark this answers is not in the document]
MichaelBrim: My suggestion would be to require the use of quotes (either ' or ") surrounding filenames with spaces.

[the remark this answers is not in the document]
sandrain: We will follow the suggestions of using space. I used comma because it was simpler; I think there would are filenames with spaces and quotes than with comma (maybe I am wrong). From user's perspective, I am not sure if putting a space and required surrounding quotes would be more convenient than just putting a comma in between.

craigsteffen:
I need to clean up the argument parsing and change the manifest file format, and then I'll take the WIP tag off.

util/unifyfs/src/unifyfs.c
```diff
@@ -180,9 +182,11 @@ static void parse_cmd_arguments(int argc, char** argv)
 
         case 'h':
         default:
+	    printf("\n\nArgument %d is invalid!\n", argument_count);
```
tonyhutter, line 185: `-h` should not print this error.

craigsteffen: Oooh--good catch.  I'll fix that in the next commit.  Thanks.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -651,6 +821,50 @@ static int jsrun_terminate(unifyfs_resource_t* resource,
...
 }
 
+/**
+ * @brief Launch data stage using IBM jsrun
+ *
+ * @param resource The job resource record
+ * @param args     The command-line options
+ *
+ * @return
+ */
```
sandrain, line 831: @craigsteffen I forgot to include a note in the comment, saying that "currently we are launching a process per node, but we could launch more processes based on the available cores per node in future."

util/unifyfs/src/unifyfs.c
```diff
@@ -170,20 +173,23 @@ static void parse_cmd_arguments(int argc, char** argv)
...
             usage(0);
             break;
+
+        default:
+	    printf("\n\nArgument %d is invalid!\n", argument_count);
+            usage(-EINVAL);
+            break;
         }
+        argument_count++;
     }
```
sandrain, line 193: @craigsteffen Also forgot here: if --stage-out is given, we need to verify if '--shared-dir' is given together.

craigsteffen:
Argh.  Please disregard the latest push to the branch.  I pushed to a separate testing branch and for some reason git pushed to this one too.  I don't know why and I will remove it as soon as I figure out how. 

Hyogi and I are testing these changes.  When we're finished testing, we'll squash the commits, rebase them against dev, and then push that.

util/unifyfs/src/unifyfs.c
craigsteffen: Yep.  I changed that in the commits that I'm working on.  Thanks for the cross-check.

craigsteffen:
> Argh. Please disregard the latest push to the branch. I pushed to a separate testing branch and for some reason git pushed to this one too. I don't know why and I will remove it as soon as I figure out how.
> 
> Hyogi and I are testing these changes. When we're finished testing, we'll squash the commits, rebase them against dev, and then push that.

Ok, I think I fixed this again.  

I got caught by git's default behavior for push.default changing to "matching" which seems to me like it's a dumb change.  Anyway, I've changed it to current for mine and I"ll try to remember to run it with --dry-run in the future.

craigsteffen:
This is ready to go.  I've tested it on summitdev and Hyogi has tested it on summit.  

I'll put in a separate branch and PR for the documentation.

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,292 @@
...
+static int checksum;
+static int mode;
+static char* manifest_file;
+static char* mountpoint = "/unifyfs";
+static char* share_dir;
+
+static unifyfs_stage_t _ctx;
+
+/**
+ * @brief create a status file for unifyfs cli to get niotified with the data
```
tonyhutter, line 68: I don't understand this sentence

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,292 @@
...
+ * stage result
+ *
+ * @param status 0 indicates success
+ *
+ * @return 0 on success, errno otherwise
+ */
+static int create_status_file(int status)
+{
+    char filename[PATH_MAX] = { 0, };
```
tonyhutter, line 77: This doesn't need to be initialized to zero.

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,292 @@
...
+    FILE* fp = NULL;
+    const char* msg = status ? "fail" : "success";
+
+    sprintf(filename, "%s/%s", share_dir, UNIFYFS_STAGE_STATUS_FILENAME);
```
tonyhutter, line 81: It's safer to use `scnprintf()`

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -244,6 +248,99 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
...
 }
 
+/**
+ * @brief wait until data stage operation finishes
+ *
+ * @param resource
+ * @param args
+ *
+ * @return
+ */
```
tonyhutter, line 258: What is this comment for?

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -244,6 +248,99 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
...
+    /* FIXME: we just wait for 20 mins.
+     * In fact, we can roughly estimate time for stage in by looking at the
+     * total file size. However, such a method is not possible for stage out
+     * because we have no idea on what the total transfer size is.
+     */
+    return 20*60;
+}
+
+static
+int wait_stage(unifyfs_resource_t* resource, unifyfs_args_t* args, int op)
```
tonyhutter, line 276: Document this function

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -271,6 +368,27 @@ static int remove_server_pid_file(unifyfs_args_t* args)
     return ret;
 }
 
+static int remove_stage_status_file(unifyfs_args_t* args)
```
tonyhutter, line 371: Document this function

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -271,6 +368,27 @@ static int remove_server_pid_file(unifyfs_args_t* args)
...
+{
+    int ret = 0;
+    char filename[PATH_MAX] = { 0, };
+
+    sprintf(filename, "%s/%s", args->share_dir, UNIFYFS_STAGE_STATUS_FILENAME);
```
tonyhutter, line 376: use `scnprintf()`

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -271,6 +368,27 @@ static int remove_server_pid_file(unifyfs_args_t* args)
     return ret;
 }
 
+static int remove_stage_status_file(unifyfs_args_t* args)
+{
+    int ret = 0;
+    char filename[PATH_MAX] = { 0, };
```
tonyhutter, line 374: No need to zero this array

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -539,6 +657,45 @@ static size_t construct_server_argv(unifyfs_args_t* args,
...
 
+/**
+ * @brief Default data stage launch routine
+ *
+ * @param args        The command-line options
+ * @param stage_args  unifyfs-stage argument vector to be filled
+ *
+ * @return number of server arguments
+ */
+static size_t construct_stage_argv(unifyfs_args_t* args,
```
tonyhutter, line 668: Returning a `size_t` from this is weird.  `size_t` is only used for returning the size of an object.  Use an unsigned int.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -539,6 +657,45 @@ static size_t construct_server_argv(unifyfs_args_t* args,
     return argc;
 }
 
+/**
+ * @brief Default data stage launch routine
```
tonyhutter, line 661: "Default data stage launch routine" doesn't really tell me what this function is doing.  What about, "Set default values for args[] and stage_argv[]"?

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -539,6 +657,45 @@ static size_t construct_server_argv(unifyfs_args_t* args,
...
+ *
+ * @param args        The command-line options
+ * @param stage_args  unifyfs-stage argument vector to be filled
```
tonyhutter, line 664: rename "vector" to "array"

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -539,6 +657,45 @@ static size_t construct_server_argv(unifyfs_args_t* args,
...
+        stage_argv[0] = strdup(LIBEXECDIR "/unifyfs-stage");
+    }
+    argc = 1;
+
+    if (args->mountpoint != NULL) {
+        if (stage_argv != NULL) {
+            stage_argv[argc] = strdup("-m");
+            stage_argv[argc + 1] = strdup(args->mountpoint);
+        }
+        argc += 2;
```
tonyhutter, line 683: If `args->mountpoint != NULL && stage_argv == NULL`, should we still be doing `argc += 2`?

util/unifyfs/src/unifyfs-rm.c
tonyhutter: You should also document the minimum size `stage_args[]` should be (6 entries?)

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -567,6 +724,20 @@ static int invalid_terminate(unifyfs_resource_t* resource,
...
+ * @param args      Not used
+ *
+ * @return -ENOSYS
+ */
+static int invalid_stage(unifyfs_resource_t* resource,
+                         unifyfs_args_t* args)
+{
+    return -ENOSYS;
+}
+
```
tonyhutter, line 740: This function is unnecessary, and the return code is wrong for this context:
```
ENOSYS

“Function not implemented.” This indicates that the function called is not implemented at all,
either in the C library itself or in the operating system. When you get this error, you can be
sure that this particular function will always fail with ENOSYS unless you install a new version
of the C library or the operating system. 
```
Replace all callers with another error code.

util/unifyfs/src/unifyfs.c
```diff
@@ -130,31 +131,33 @@ static void parse_cmd_arguments(int argc, char** argv)
...
-        case 'e':
-            srvr_exe = strdup(optarg);
+	case 'e':
+	    srvr_exe = strdup(optarg);
             break;
 
-        case 'm':
-            mountpoint = strdup(optarg);
+	case 'm':
+	    mountpoint = strdup(optarg);
```
tonyhutter, line 160: Fix indents.  It looks like you may be mixing tabs with spaces.

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +265,47 @@ int main(int argc, char** argv)
...
+	        fprintf(stderr,
+			"WARNING! You are using a script with --stage-out.\n");
+		fprintf(stderr, "This won't work.\n");
+		return -EINVAL;
+            }
+            if (stat(cli_args.stage_out, &manifest_file_stat)) {
+	        fprintf(stderr,
+			"Cannot read stageout manifest file:%s\n",
+			cli_args.stage_out);
+		return -ENOENT;
```
tonyhutter, line 306: Fix your indents

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +265,47 @@ int main(int argc, char** argv)
...
+		usage(1);
+		return -EINVAL;
+            }
+            if (cli_args.script) {
+	        fprintf(stderr,
+			"WARNING! You are using a script with --stage-out.\n");
+		fprintf(stderr, "This won't work.\n");
+		return -EINVAL;
+            }
+            if (stat(cli_args.stage_out, &manifest_file_stat)) {
```
tonyhutter, line 302: Use `access()` to check for the existence of a file instead of `stat()`.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+#include <openssl/md5.h>
+
+#include "unifyfs-stage.h"
+
+static int md5_checksum(const char* path, unsigned char* digest)
+{
+    int ret = 0;
+    size_t len = 0;
+    int fd = -1;
+    unsigned char data[1024] = { 0, };
```
tonyhutter, line 42: This doesn't need to be initialized to zero

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+
+out:
+    /* MD5_xx returns 1 for success */
+    ret = ret == 1 ? 0 : EIO;
+    close(fd);
+
+    return ret;
+}
+
+static inline char* checksum_str(char* buf, unsigned char* digest)
```
tonyhutter, line 78: Why is this inlined?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+    pos[0] = '\0';
+
+    return buf;
+}
+
+static int verify_checksum(const char* src, const char* dst)
+{
+    int ret = 0;
+    int i = 0;
+    char md5src[2 * MD5_DIGEST_LENGTH + 1] = { 0, };
```
tonyhutter, line 96: Question: why is this `2 * MD5_DIGEST_LENGTH + 1`?  I'm not saying it's wrong, it's just not clear to me what it is.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+                            checksum_str(md5src, src_digest),
+                            checksum_str(md5dst, dst_digest));
+            ret = EIO;
+        }
+    }
+
+    return ret;
+}
+
+static inline char* str_rtrim(char* str)
```
tonyhutter, line 137: Why inline?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+
+    return str;
+}
+
+// format of the manifest file lines (for the moment) is as
+// follows: each line consists of the source filename including
+// full path, then one or more spaces, and then the destination
+// filename including full path.  If neither of the filenames
+// contain spaces, the filenames can lack quotes around them.
+// If the filenames contain spaces, then they may be surrounded
```
tonyhutter, line 156: "may be" -> "must be"

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+// with double quote symbols (").  The only characters that are
+// disallowed with this format are the double-quote symbol
+// and the newline character.
+static inline int unifyfs_parse_manifest_line
+                  (char* line, char** src, char** dst)
+{
+    char* ptr_doublequotes[4];
+    char* CSP = NULL; // "current search pointer"
```
tonyhutter, line 164: Please don't use all caps names for variables (only #defines)

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+    char* loc_of_space = NULL;
+    char* loc_of_newline = NULL;
+    char test_char;
+    size_t offset;
+
+    // does the first filename start with a doublequote character?
+    if ((*line) == '"') {
```
tonyhutter, line 171: Did you consider using `strtok()` to get the quoted filenames (if any)?

tonyhutter:
`jsrun_stage()`, `mpirun_stage()`, and `srun_stage()` all share the same basic code with different data values.  Why not refactor them so that they pass an arg string to a new `generic_stage()` function that tokenizes an arg string and assigns it to args[], and does the `execvp()`?

Psudocode:
```C
srun_stage() {
	sprintf(cmd, "srun -N %zu --ntasks-per-node 1", resource->n_nodes)
	generic_stage(cmd);
}

mpirun_stage()
{
	sprintf(cmd, "mpirun -np %zu --map-by ppr:1:node", resource->n_nodes);
	generic_stage(cmd);
}
...

**Comment by craigsteffen** [comment]
There are a bunch of comments.  I will try to get a patch that addresses as many of them as I can pushed at least a couple of hours before the afternoon meeting so we can discuss any of the stickier ones then.

File: util/unifyfs-stage/src/unifyfs-stage.c (hunk 10 above)
Comment (craigsteffen) [reply util/unifyfs-stage/src/unifyfs-stage.c:68]: Clarified in patch.

File: util/unifyfs-stage/src/unifyfs-stage.c (hunk 12 above)
Comment (craigsteffen) [reply util/unifyfs-stage/src/unifyfs-stage.c:81]: Good point.  Changed in upcoming patch.

Comment (craigsteffen) [reply util/unifyfs-stage/src/unifyfs-stage.c:81]: If someone has any ideas as to what file to #include to get this function prototype, that would be helpful.  I've been looking for 20 minutes and can't find it.

Comment (craigsteffen) [reply util/unifyfs-stage/src/unifyfs-stage.c:81]: Never mind.  I found where the scnprintf kernel function is re-implemented in the unifyfs source tree.

File: util/unifyfs/src/unifyfs-rm.c (hunk 13 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:258]: I think it was meant to be attached to a function farther down.  Moved.

File: util/unifyfs/src/unifyfs-rm.c (hunk 14 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:276]: This was the floating comment above

File: util/unifyfs/src/unifyfs-rm.c (hunk 15 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:371]: Done.

File: util/unifyfs/src/unifyfs-rm.c (hunk 17 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:374]: I've removed the zero initalizations for all the buffers in unifyfs-rm.c that defined buffers that were written directly to in that function by string (and thus NULL terminating) functions.

File: util/unifyfs/src/unifyfs-rm.c (hunk 19 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:661]: This was just copied from the server_argv function.  I've changed them both to clarify and differentiate.

File: util/unifyfs-stage/src/unifyfs-stage.c (hunk 11 above)
Comment (craigsteffen) [reply util/unifyfs-stage/src/unifyfs-stage.c:77]: Removed.

File: util/unifyfs/src/unifyfs-rm.c (hunk 16 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:376]: Done for all occurrences of sprintf() in this file.  

I will point out that sprintf appears at least 126 times elsewhere in the unifyfs source code.

File: util/unifyfs/src/unifyfs-rm.c (hunk 18 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:668]: The new function construct_stage_argv uses the same overall logic flow as the function construct_server_argv already existing in the current dev branch, which also returns size_t.  

If you'd like to overhaul how the functions in unifyfs-rm are written and/or structured, by all means let's open a new issue to overhaul that file, but let's do it as a separately from adding stage-in/stage-out functionality, which we've written to mirror the existing function templates in that file.

File: util/unifyfs/src/unifyfs-rm.c (hunk 20 above)
Comment (craigsteffen) [reply util/unifyfs/src/unifyfs-rm.c:664]: RE vector/array: That's the way the description for the existing function construct_server_argv was written.  Also I think vector is at least as correct as array in this instance.  

Let's update the documentation of all the functions together as a separate issue/branch/PR (unless construct_server_args is documented somewhere that I missed and that construct_stage_argv doesn't match that).

File: util/unifyfs-stage/src/unifyfs-stage-transfer.c [hunk 33]
```diff
@@ -0,0 +1,376 @@
...
+
+int unifyfs_stage_transfer(unifyfs_stage_t* ctx)
+{
+    int ret = 0;
+    int count = 0;
+    FILE* fp = NULL;
+    char* src = NULL;
+    char* dst = NULL;
+    char linebuf[LINE_MAX] = { 0, };
+    struct stat sb = { 0, };
```
tonyhutter, line 257: These don't need to be zeroed

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,376 @@
...
+        ret = errno;
+        goto out;
+    }
+
+    while (NULL != fgets(linebuf, LINE_MAX-1, fp)) {
+	if (strlen(linebuf) < 5) {
+	    // the manifest file perhaps ends with a couple of characters
+	    // and/or a newline not meant to be a transfer spec.
+            goto out;
+	}
```
tonyhutter, line 276: Allowing stray newlines is ok, but any other stray characters should be flagged as an error.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: That's a good question.   I don't really know.  But again, that mirrors the structure of construct_server_argv, which is constructed similarly.  

I have a counter question to this.  The last block in both construct_server_argv and construct_stage_argv has the argc+= stage OUTSIDE the whole block.  I can't think that's actually correct or what was intended.  Here's the one at the bottom of (existing function) construct_server_argv:
    if (server_argv != NULL) {
        server_argv[argc] = strdup("-S");
        server_argv[argc + 1] = strdup(args->share_dir);
        server_argv[argc + 2] = strdup("-H");
        server_argv[argc + 3] = strdup(args->share_hostfile);
    }
    argc += 4;

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Fair point.  Fixed in next patch.

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,292 @@
...
+        fprintf(stderr, " ENTER to continue ... ");
+
+        (void) getchar();
+    }
+
+    MPI_Barrier(MPI_COMM_WORLD);
+
+    /* internal accept() call from mpi may set errno */
+    errno = 0;
+}
```
tonyhutter, line 166: Can we get rid of this?

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,292 @@
...
+    ctx->mountpoint = mountpoint;
+    ctx->manifest_file = manifest_file;
+
+    if (verbose) {
+        unifyfs_stage_print(ctx);
+    }
+
+    if (debug) {
+        debug_pause(rank, "About to mount unifyfs.. ");
+    }
```
tonyhutter, line 259: Is this really needed?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: I hadn't.  It might be suitable, but I find the man page a bit murky to understand completely.  

I will need to do a bit of experimentation before I'm confident I can invoke strtok() correctly and consistently.  I'm sure we'll want to do a cleanup of the manifest line parsing in the near future.  We can add strtok() then.

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +265,47 @@ int main(int argc, char** argv)
...
+            return -EINVAL;
         }
+        if (cli_args.stage_in != NULL) {
+            if (cli_args.script) {
+	        fprintf(stderr,
+			"WARNING! You are using a script with --stage-in.\n");
+		fprintf(stderr, "This is won't work.\n");
+		return -EINVAL;
+            }
+            if (stat(cli_args.stage_in, &manifest_file_stat)) {
```
tonyhutter, line 277: Use `access()` to check if file exists

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +265,47 @@ int main(int argc, char** argv)
...
+		return -ENOENT;
+            }
+	}
         return unifyfs_start_servers(&resource, &cli_args);
     } else if (action == ACT_TERMINATE) {
+        if (cli_args.stage_out != NULL) {
+	    // status directory isn't required just to terminate the servers
+	    // but it IS required if we're calling for stage-out, so that
+	    // the stage-out can be started, then the server terminate waits
+	    // until the stage-out is done.
```
tonyhutter, line 290: Indents

util/unifyfs/src/unifyfs.h
```diff
@@ -43,6 +43,7 @@
 #endif
 
 #include <sys/types.h>
+#include <sys/stat.h>
```
tonyhutter, line 46: Why is this needed?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Yes.  I got rid of all the zero initializations in these files in the patch that I will be uploading probably within the next half hour.

util/unifyfs/src/unifyfs.c
craigsteffen: Done in next patch.

util/unifyfs/src/unifyfs.h
craigsteffen: For stat(), which has been replaced.

util/unifyfs-stage/src/unifyfs-stage.c
craigsteffen: Don't know.  I'll defer to Hyogi on this.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Ok.  I'll re-do this.

util/unifyfs/src/unifyfs-rm.c
MichaelBrim: @craigsteffen @tonyhutter `construct_server_argv()` is used in two modes. The first mode is when `server_argv == NULL`, in which it just returns the count of arguments that will reside in `server_argv`. That is why `argc` is incremented outside the checks for NULL. The returned argc count is then used to allocate the actual server arguments array, and the function is called again in the second mode, which fills in the argv array. The reason the return value is `size_t` is because that is the appropriate type for the size of an array (and also the size parameter to malloc/calloc).

util/unifyfs-stage/src/unifyfs-stage-transfer.c
MichaelBrim: @tonyhutter see my above comment where I asked hyogi to change it to this. It is determined by how many characters will be written to the string in `checksum_str()`

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Fixed in upcoming patch.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
MichaelBrim: `strtok()` is the way to go. Learn it, love it, and stop reinventing libc functions.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Which function should that comment go in?

util/unifyfs-stage/src/unifyfs-stage.c
MichaelBrim: This is really a developer convenience method (used by Hyogi mostly) to make it so that he can attach a debugger at the beginning of the program. It can be useful.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: This mirrors the functions invalid_terminate, and invalid_stage that are in the dev branch.  If we need to change all of them let's open a new issue and create a new branch+PR.

util/unifyfs/src/unifyfs.c
craigsteffen: This probably comes from me fixing the "code should start on a tabstop" in the patch checker.  Yes, emacs uses tabs/spaces relatively indiscriminately.  

I'm very sorry the diff looks ugly in the github display.  The source, however, looks fine as far as I can see and the extremely persnickety syntax checker is fine with it.  So unless there's a source file that's a problem when they check it out and use it in their local tools, I'm going to just leave this.

util/unifyfs/src/unifyfs.c
craigsteffen: I've fixed this in the upcoming patch.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: I removed all these.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Agreed.  Fixed.

util/unifyfs/src/unifyfs.h
craigsteffen: So this is removed.

util/unifyfs/src/unifyfs.c
MichaelBrim: @craigsteffen It is your responsibility to match the existing code style, which is 4-space indents (no tabs). You can easily set up your emacs config to do this (if you need help, just ask and I can provide the appropriate .emacsrc settings).

MichaelBrim: I just want to clarify, I'm not suggesting you introduced all the tabs in this file. But, when you edit the lines, you should make sure the result only has 4-space indents. As a project, we really should be using a coding convention tool that automatically applies the desired style to all source files, so we can stop worrying about this in code reviews.

util/unifyfs-stage/src/unifyfs-stage.c
sandrain: @tonyhutter I am mostly using it because it's convenient to attach from gdb. I don't think it's a big problem to keep the code.

util/unifyfs-stage/src/unifyfs-stage.c
sandrain: this was my habit of initializing stack vars. technically, we don't have to.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: I hadn't used strtok() before.  Whenever possible I parse lines using regex in perl, and I was in a hurry.  

If you still want this change before approving I'll make this change last.

util/unifyfs/src/unifyfs.c
craigsteffen: I'll run astyle on this and check it in with the next patch.

util/unifyfs/src/unifyfs.c
craigsteffen: I'll run astyle on this and repush it.

tonyhutter:
Can you add test cases for this?

craigsteffen:
Ok, it looks like the integration test is failing because it can't find this:
#include <common/src/unifyfs_misc.h>
which is necessary for scnprintf(), because that's apparently a re-implemented kernel function. I may be able to fix this by re-implementing it with snprintf(); is that acceptable?

craigsteffen:
I can't tell if the Travis failure has anything to do with removing the path from unifyfs_misc.h.  The failure seems to be that it can't run some "spack install", which I can't image could be caused by that change in the source file.  

By the way, I did spell the name of the file unifyfs_misc.h correctly in the include in the source, just not in the commit message.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: Please use this simplified manifest parser:
https://gist.github.com/tonyhutter/135155fc185f568d61c046875d39699c
It allows quoted filenames, a whitespace separator (not just space), and ignores trailing/leading whitespace.  This allows the user wants to indent certain lines, or tab-align source/destination filenames if they want.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: I added a comment to the source indicating that.  Thanks Mike.

util/unifyfs/src/unifyfs.c
craigsteffen: I patched this in the copy that will be pushed soon.

util/unifyfs/src/unifyfs.c
craigsteffen: Cleaned up in a future update.

util/unifyfs-stage/src/unifyfs-stage.c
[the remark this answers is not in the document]
craigsteffen: The new version that we're testing now uses spaces, and in fact, the new one that Tony has sent me does this too.  So this is resolved.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: This has been fixed.

util/unifyfs/src/unifyfs.c
craigsteffen: Fixed.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: This has now been added to the new patch that will go in.

util/unifyfs/src/unifyfs.c
craigsteffen: This has been fixed and will be pushed to the branch soon.

util/unifyfs/src/unifyfs-rm.c
sandrain: Sorry for this late comment. I've been thinking a bit about this. The case is different here from spawning unifyfsd because the data transfer will be terminated, while unifyfsd will continue to run. So, how about simply launching the unifyfs-stage using popen instead of using exec? Maybe, we can address this later after this PR is merged.

craigsteffen:
So this is for the stage-in?

So what you're saying is just wait for it to return from the stage-in
before it returns from the unifyfs comment?  That makes sense to me.  It's
only a one or two-line change, so I think it's fine to go in with this PR.

Craig

sandrain:
@craigsteffen I mean for both stage-in and stage-out. In both cases, the `unifyfs` utility should wait until the `unifyfs-stage` completes. The current way is just to use the status file, which is suitable for launching `unifyfsd` daemon because `unifyfsd` needs to be running when we `unifyfs start`. But, the `unifyfs-stage` should terminate anyway before the `unifyfs` daemon resume its execution. So, I think it would be fine to use `popen()`, and we don't have to introduce any status file or wait time.

util/unifyfs-stage/src/unifyfs-stage.c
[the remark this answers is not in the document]
MichaelBrim: @craigsteffen the `usage_str` still needs to be updated to reflect the new manifest line format

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +274,47 @@ int main(int argc, char** argv)
         if (NULL == cli_args.share_dir) {
             printf("USAGE ERROR: shared directory (-S) is required!\n");
             usage(1);
+            return -EINVAL;
+        }
+        if (cli_args.stage_in != NULL) {
+            if (cli_args.script) {
+                fprintf(stderr,
+                        "WARNING! You are using a script with --stage-in.\n");
+                fprintf(stderr, "This is won't work.\n");
```
MichaelBrim, line 283: "This is won't" -> "This will not"

adammoody: I like "This won't is work"

craigsteffen:
Oops.  Indeed; I'll fix that.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,403 @@
...
+                goto out;
+            }
+	}
+	ret = unifyfs_parse_manifest_line(linebuf, &src, &dst);
+	if (ret < 0) {
+	    fprintf(stderr, "failed to parse %s (%s)\n",
+		    linebuf, strerror(ret));
+            goto out;
+	}
+        if (ctx->mode == UNIFYFS_STAGE_SERIAL) {
```
tonyhutter, line 311: indents

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,403 @@
...
+                 * can skip this step.
+                 */
+                fd = open(dst, O_WRONLY|O_CREAT|O_TRUNC, 0600);
+                if (fd < 0) {
+                    fprintf(stderr, "[%d] failed to create the file %s\n",
+                                    rank, dst);
+                    goto out;
+                }
+
+                close(fd);
```
tonyhutter, line 360: I'm not clear on why this empty file needs to be created

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,403 @@
...
+            }
+
+            MPI_Barrier(MPI_COMM_WORLD);
+
+            ret = unifyfs_transfer_file_parallel(src, dst);
+            if (ret) {
+                goto out;
+            }
+
+            MPI_Barrier(MPI_COMM_WORLD);
```
tonyhutter, line 370: What is the reason for these MPI_Barriers()?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
```diff
@@ -0,0 +1,403 @@
...
+
+            ret = stat(dst, &sb);
+            if (ret < 0) {
+                fprintf(stderr, "stat on %s failed (err=%d, %s)\n",
+                                dst, errno, strerror(errno));
+                ret = errno;
+                goto out;
+            }
```
tonyhutter, line 378: What is this stat here for?  Is it necessary?

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -244,6 +256,111 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
...
+    while (1) {
+        fp = fopen(filename, "r");
+        if (fp) {
+            char* line = fgets(linebuf, 15, fp);
+            if (0 == strncmp("success", line, strlen("success"))) {
+                ret = 0;
+                break;      // transfer completed
+            } else if (0 == strncmp("fail", line, strlen("fail"))) {
+                ret = -EIO;
+                break;      // transfer failed
```
tonyhutter, line 326: Do you need to `fclose(fp)` for these cases?

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -244,6 +256,111 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
...
+    return ret;
+}
+
+static inline int wait_stage_in(unifyfs_resource_t* resource,
+                                unifyfs_args_t* args)
+{
+    return wait_stage(resource, args, UNIFYFS_STAGE_IN);
+}
+
+static inline int wait_stage_out(unifyfs_resource_t* resource,
```
tonyhutter, line 358: These should not be inlined

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: @tonyhutter Otherwise, all ranks will try to open the target file with O_CREAT inside the unifyfs_trasnfer function. This leads to an incomplete/corrupted file in PFS.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: The first barrier is to ensure the shared file is created by the rank 0. Second one is to wait until all processes finish transfer of the file, before checking the result (checksum).

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: You previously pointed this out, and it seems unnecessary. I am not 100% sure on this.

util/unifyfs/src/unifyfs-rm.c
sandrain: Would you mind explaining why they should not?

util/unifyfs/src/unifyfs-rm.c
sandrain: Correct, we need to.

util/unifyfs/src/unifyfs.c
craigsteffen: Now fixed in an upcoming release.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: I assume this is resolved then?

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: I've removed this in a future push.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Fixed in future commit with astyle.

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Fixed in upcoming commit.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: I guess so, because we cannot synchronize mpi processes within the transfer api function, unless I missed something, or there's other simple fix. @tonyhutter

util/unifyfs/src/unifyfs-rm.c
craigsteffen: Tony?  Any thoughts on this?  Why shouldn't these functions be inlined?  
I don't care either way, but it seems to be a reasonable use of inlining.  It lowers the duplication of code by having the lower function called with two different function wrappers but without the added context switch induced by calling one function with another one.

tonyhutter: https://www.kernel.org/doc/html/latest/process/coding-style.html#the-inline-disease
In this case the function is static, so the compiler will already optimize it.

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -0,0 +1,300 @@
...
+
+static void print_usage(void)
+{
+    if (0 == rank) {
+        fprintf(stdout, usage_str, program);
+    }
+}
+
+static inline
+void debug_pause(int rank, const char* fmt, ...)
```
tonyhutter, line 156: This should not be inlined.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: Did you try having `do_transfer_file_parallel()` `open(O_CREAT|O_EXCL...)`, and if that doesn't work, `open(O_WRONLY)`?

util/unifyfs/src/unifyfs-rm.c
sandrain: @tonyhutter I see your point there, but isn't it more like it "doesn't have to" be inlined? Like Craig mentioned, it is not critical whether these functions become inline or not, but I don't think it makes much sense to say "inline keyword is there to give a hint to the compiler to inline the function when possible, but you shouldn't be using it because the compiler is going to do it anyway when possible." The document you linked basically says "use it appropriately."  These functions have a single line, which one could go with macros instead. So, I don't see any problems with specifying the inline keyword there. Also, the compiler would not inline a function when not appropriate even with the inline keyword.

util/unifyfs-stage/src/unifyfs-stage.c
craigsteffen: Removed in next upload.

craigsteffen:
This passes all checks now.  

I've created unit tests as asked.  I created one that uses the unifyfsd and one that doesn't, so both can be automatically tested.  

I think I've removed most of the inline functions, except for the one-line function shim that I don't think matters.  

There are some other simplifications that Hyogi and I agree are better left to a separate issue/PM. 

I believe the concensus on the only other issues above (the MPI barrier and the waits) are that they are fine as is for now.  

Let's try to talk about any remaining issues on this PR today during the calls so we can get this committed.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
sandrain: @tonyhutter I think that will work, so something like:

```
do _transfer_file_parallel()
{
    fd = open(O_CREAT|O_EXCL ...); // possibly only single process will succeed

    if (fd < 0 && errno = EEXIST) { // for others
        fd = open(O_WRONLY );
    }
}
```

although we would still need one of the barriers afterwards.

craigsteffen:
Also I've kept this up to date against dev, so merging shouldn't be a problem.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Do we need to do anything with this before the PR is ready to upload?  

If the only thing we're saving is not creating an empty (lock?) file, then that doesn't seem like it's worth the effort to stop and do the same thing with a different method.

util/unifyfs-stage/src/unifyfs-stage.c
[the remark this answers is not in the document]
craigsteffen: Oops.  This was so far up on the PR page I missed it today when I was cleaning up.  I'll do this now.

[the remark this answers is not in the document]
craigsteffen: This is now fixed in the force push I just pushed in.  It has a correct and slightly more elaborate usage message about the manifest file. 

I also put the MPI_Init() after the argument parsing, so that if it's invoked without an MPI environment with --help or no arguments, it produces sensible output.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: @sandrain yep, that's what I was thinking.  @craigsteffen given that the PR adds a comment saying "FIXME:  ...it would be better if we could fix this step", I think we should fix it in this PR.  Then we could delete that comment.

craigsteffen: Ok, I can change this and re-push it. 

As long as I'm doing that, anything else you'd like me to work on Tony, or can the rest be pushed the way it is now?

util/unifyfs/src/unifyfs-rm.c
tonyhutter: Ok, I think that's fair then if you want to inline it to signify "this is a function that's basically a macro".

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: Got it, thanks for the info.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: Ok, I've read this discussion more carefully this time, but I'm confused.  
The comment says "Since we cannot use the mpi barrier inside the unifyfs_transfer_file_parallel()".  Why can't we?  Why can't we just MPI_File_Open inside the unifyfs_transfer_file_paralle() using the GLOBAL communicator?  We know we're in an MPI context.  That's the natural thing to do.  
If for some reason we can't do that, then I propose we leave the code like it is and remove the FIXME comment.  I don't see any particular reason not to code it that way.  One rank will make a metadata call to create the file, then lots of ranks will open it, and they're all opening it expecting it to be there so they aren't all hitting errors and then having to open it up again.

util/unifyfs-stage/src/unifyfs-stage.c
```diff
@@ -284,16 +298,19 @@ int main(int argc, char** argv)
...
+    if (should_we_mount_unifyfs) {
+        ret = unifyfs_unmount();
+        if (ret) {
+            fprintf(stderr, "unmounting unifyfs failed (ret=%d)\n", ret);
+        }
     }
-out:
+ out:
     MPI_Finalize();
+ preMPIout:
```
tonyhutter: Indent

util/unifyfs-stage/src/unifyfs-stage-transfer.c
tonyhutter: @craigsteffen that's fine.  The main reason I was suggesting using O_EXCL was because the comment made it sound like the "rank 0 creates the file first" way was undesirable, but if you guys are ok with it then leave it as it.

t/0700-unifyfs-stage-full.t
```diff
@@ -0,0 +1,64 @@
...
+# Test unifyfs-stage executable for basic functionality
+#
+
+test_description="Test basic functionality of unifyfs-stage executable"
+
+. $(dirname $0)/sharness.sh
+
+#test_expect_success "Intercepted mount point $UNIFYFS_MOUNTPOINT is empty" '
+#    test_dir_is_empty $UNIFYFS_MOUNTPOINT
+#'
```
CamStan, line 12: Can remove this. Assuming it's leftover from copying the file from a previous test.

t/9300-unifyfs-stage-isolated.t
```diff
@@ -0,0 +1,61 @@
...
+# Test unifyfs-stage executable for basic functionality
+#
+
+test_description="Test basic functionality of unifyfs-stage executable"
+
+. $(dirname $0)/sharness.sh
+
+#test_expect_success "Intercepted mount point $UNIFYFS_MOUNTPOINT is empty" '
+#    test_dir_is_empty $UNIFYFS_MOUNTPOINT
+#'
```
CamStan, line 12: Same here.

t/9300-unifyfs-stage-isolated.t
```diff
@@ -0,0 +1,61 @@
...
+
+test_expect_success "config_9300 directory is empty" '
+    test_dir_is_empty ${UNIFYFS_TEST_TMPDIR}/config_9300
+'
+
+echo "\"${UNIFYFS_TEST_TMPDIR}/stage_source/source.file\" \"${UNIFYFS_TEST_TMPDIR}/stage_destination_9300/destination_9300.file\""  > ${UNIFYFS_TEST_TMPDIR}/config_9300/test_INOUT.manifest
+
+test_expect_success "config_9300 directory now has manifest files" '
+    test_path_is_file  ${UNIFYFS_TEST_TMPDIR}/config_9300/test_INOUT.manifest
+
```
CamStan, line 48: Empty line here inside test can probably be removed.

t/0700-unifyfs-stage-full.t
craigsteffen: Oops.  I thought I'd cleaned out all the cruft; looks like I missed a few.

craigsteffen:
Cameron, thanks for the pointer for the file compare test.  Yes, that would make it much better.  I will implement that today.

util/unifyfs-stage/src/unifyfs-stage-transfer.c
craigsteffen: comment removed in upcoming patch.

util/unifyfs-stage/src/unifyfs-stage.c
craigsteffen: Ok.  One space before tags is how emacs formats them by default.  Neither checkpatch, make stylecheck nor Travis said anything about the last several checkins.  Astyle apparent has a different convention.  I take it that astyle is the definitive guide on what's correct, not checkpatch?

t/0700-unifyfs-stage-full.t
craigsteffen: removed in patch.

t/9300-unifyfs-stage-isolated.t
craigsteffen: removed

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -244,6 +256,115 @@ static int wait_server_initialization(unifyfs_resource_t* resource,
...
+    return_val_from_scnprintf =
+        scnprintf(filename, PATH_MAX,
+                  "%s/%s", args->share_dir, UNIFYFS_STAGE_STATUS_FILENAME);
+    if (return_val_from_scnprintf > (PATH_MAX - 2)) {
+        fprintf(stderr, "Unifyfs status filename is too long!\n");
+        return -ENOMEM;
+    }
+
+
+    if (op == UNIFYFS_STAGE_IN) {
```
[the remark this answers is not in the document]
craigsteffen: HUH.  You're right.  I know we used it at some point, but obviously something changed and we didn't progogate it through all the way.  

Fixed in upcoming patch.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -935,17 +1311,65 @@ int unifyfs_start_servers(unifyfs_resource_t* resource,
...
+    pid_t pid;
 
     if ((resource == NULL) || (args == NULL)) {
         return -EINVAL;
     }
 
+    if (args->stage_out) {
+        rc = remove_stage_status_file(args);
+        if (rc) {
+            fprintf(stderr, "ERROR: failed to remove stage status file\n");
```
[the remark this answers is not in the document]
craigsteffen: Got rid of the ERROR enunciations.

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +274,47 @@ int main(int argc, char** argv)
         if (NULL == cli_args.share_dir) {
             printf("USAGE ERROR: shared directory (-S) is required!\n");
             usage(1);
+            return -EINVAL;
+        }
+        if (cli_args.stage_in != NULL) {
+            if (cli_args.script) {
+                fprintf(stderr,
+                        "WARNING! You are using a script with --stage-in.\n");
+                fprintf(stderr, "This will not work.\n");
```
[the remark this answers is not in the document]
craigsteffen: Done in upcoming patch.

util/unifyfs/src/unifyfs.c
```diff
@@ -258,9 +274,47 @@ int main(int argc, char** argv)
...
+            // until the stage-out is done.
+            if (NULL == cli_args.share_dir) {
+                printf("USAGE ERROR: shared directory (-S) is required!\n");
+                usage(1);
+                return -EINVAL;
+            }
+            if (cli_args.script) {
+                fprintf(stderr,
+                        "WARNING! You are using a script with --stage-out.\n");
+                fprintf(stderr, "This will not work.\n");
```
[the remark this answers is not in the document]
craigsteffen: Changed.

t/9300-unifyfs-stage-isolated.t
craigsteffen: Done in upcoming.

craigsteffen:
Ok, everything I've promised above, as far as I know, is in this branch now.  

Everything works on summitdev, but now it apparently doesn't work in the Travis environment.  

Several previous versions of these tests have worked fine, both on summitdev and in Travis.  

Cameron suggested added a check to test the initial input file against the final output file.  When the initial input file was sourced from /dev/zero, that works fine on summitdev for both test 0700 (moving the file through unifyfs) and 9300 (moving the file directly from directory to directory).  However, the last thing I did was to use the unifyfs-stage executable itself as the file to move.  This will make the test somewhat able to detect the file being scrambled during transfer.  

Doing that test on summitdev works fine.  0700 and 9300 tests all pass.  But in the Travis check triggered by the push, the final compare test fails on test 0700 (passes in 9300).  

Since I can't see inside Travis I don't know how to triage this, much less find the problem or fix it.  Anyone have any ideas?

craigsteffen:
Thanks Mike and Tony for the approvals.  

I'm working with Cameron to try to fix the test.  If I get it fixed, I'll upload that final version (which will only different from this in the test files).  Worst case, I'll mark the problematic test as "expect to fail" if I'm confident that the testing environment is the likely contributor to the problem.  

I'll add the "ready to merge" flag when I'm done.

craigsteffen:
Tests set up so that the final step in test 0700 is "expect_failure", which is true of Travis but not on real systems. 

Rebased against dev just now. 

All tests pass.  Ready to merge as far as I know.

t/0700-unifyfs-stage-full.t
```diff
@@ -0,0 +1,64 @@
...
+
+${SHARNESS_BUILD_DIRECTORY}/util/unifyfs-stage/src/unifyfs-stage -m ${UNIFYFS_TEST_MOUNT} ${UNIFYFS_TEST_TMPDIR}/config_0700/test_OUT.manifest > ${UNIFYFS_TEST_TMPDIR}/config_0700/stage_OUT_output.OUT 2>&1
+
+test_expect_success "input file has been staged to output" '
+    test_path_is_file ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/destination_0700.file
+'
+
+test_expect_failure "final output is identical to initial input" '
+    test_cmp ${UNIFYFS_TEST_TMPDIR}/stage_source/source_0700.file ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/destination_0700.file
+'
```
CamStan, line 62: Hmm. Well. I realize I suggested this, but now that I think about it, let's make one more change to this since you have an issue to track it:
```suggestion
test_expect_failure "final output is identical to initial input" '
    test_might_fail test_cmp ${UNIFYFS_TEST_TMPDIR}/stage_source/source_0700.file ${UNIFYFS_TEST_TMPDIR}/stage_destination_0700/destination_0700.file
'
```
I say this because as `test_expect_failure` will make this test actually fail when run locally and through GitLab CI.
Could always comment it out or just remove it and eventually add it to integration test suite, but just adding the `test_might_fail` before it will still let it run but essentially label it as "unstable."

craigsteffen: I was a bit squeamish about that myself, but I didn't know about the test_might_fail directive.  I completely agree.  

Checked uploaded and checked in, and Travis is happy with it.

CamStan: Nice! I'd forgotten it was a thing as I hadn't used it before.

PR #514, opened 2020-06-04
fix client id assignment

MichaelBrim:
### Description

Occasionally, when running many clients per server, two (or more) clients could mount at the exact same time, leading to a duplicate client id assignment. This results in the duplicate clients
both trying to use the same logio and shmem contexts, which leads to various problems, including read hangs.

This PR fixes the problem by synchronizing updates to application state using a pthread mutex. Also included is a testutil fix to make sure unifyfs_mount() failures in the example programs are
properly handled.

### How Has This Been Tested?

Tested on Summitdev with 8 ranks per node on up to 48 nodes.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/margo_client.c
+++ b/client/src/margo_client.c
@@ -290,6 +290,7 @@ int invoke_client_mount_rpc(void)
         LOGWARN("mismatch on app_id - using %d, server returned %d",
                 unifyfs_app_id, srvr_app_id);
     }
+    LOGDBG("My client id is %d", unifyfs_client_id);
 
     /* free resources */
     margo_free_output(handle, &out);
--- a/common/src/seg_tree.c
+++ b/common/src/seg_tree.c
@@ -27,12 +27,14 @@
   * segments in the tree are non-overlapping.  Added segments overwrite the old
   * segments in the tree.  This is used to coalesce writes before an fsync.
   */
+
 #include <stdio.h>
 #include <errno.h>
 #include <string.h>
 #include <stdlib.h>
 #include <assert.h>
 #include <pthread.h>
+
 #include "seg_tree.h"
 #include "tree.h"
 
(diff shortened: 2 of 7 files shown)


MichaelBrim:
I found another case where this broke. I will push the fix soon.

tonyhutter:
Nice work tracking this down :+1:

MichaelBrim:
Well, this is a fun one. It appears that pthread mutexes don't provide mutual exclusion in argobots ULTs. I guess I will try the argobots-specific mutexes, and if those work, I'll see if I can encapsulate them better for use in other margo rpc handlers where we might need mutual exclusion.

MichaelBrim:
This is now ready for review. I'm running the performance tests on Summitdev using this and it seems to be going well.

client/src/margo_client.c
```diff
@@ -290,6 +290,7 @@ int invoke_client_mount_rpc(void)
         LOGWARN("mismatch on app_id - using %d, server returned %d",
                 unifyfs_app_id, srvr_app_id);
     }
+    LOGERR("My client id is %d", unifyfs_client_id);
```
tonyhutter, line 293: Should this be a  `LOGDBG`?

MichaelBrim: yes, it should. looks like I missed converting one back to LOGDBG. I'll update and repush.

common/src/seg_tree.c
```diff
@@ -277,7 +279,7 @@ int seg_tree_add(struct seg_tree* seg_tree, unsigned long start,
 
     /* Check whether we can coalesce new extent with any preceding extent. */
     prev = RB_PREV(inttree, &seg_tree->head, target);
-    if (prev != NULL && prev->end + 1 == target->start) {
+    if ((prev != NULL) && (prev->end + 1 == target->start)) {
```
adammoody, line 282: One more pair?
```suggestion
    if ((prev != NULL) && ((prev->end + 1) == target->start)) {
```

adammoody:
Good sleuthing, @MichaelBrim

common/src/seg_tree.c
MichaelBrim: I'm in favor - anybody oppose?

tonyhutter: I'm fine with it

adammoody:
Thanks @MichaelBrim and @tonyhutter

PR #524, opened 2020-06-18
support per-file sync

MichaelBrim:
### Description

This updates the `unifyfs_sync()` method in the client to support sync of either all outstanding writes or writes to a specific file. The server rpc handler did not require any modifications to accommodate this change.

The biggest change to how the client operates is that we no longer put writes in the shmem index as they happen. This avoids any interleaving of writes to different files. Instead, each sync rpc only handles the writes for a single file, and we always generate the contents of the shmem write index from each file's `extents_sync` segment tree at the time of the sync. This behavior matches what we were doing before when `UNIFYFS_FLATTEN_WRITES` was enabled (which was the default mode of operation).

Also included were fixes for several places where the wrong in/out structure types were used in the rpc invokers and handlers.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-fixed.c
+++ b/client/src/unifyfs-fixed.c
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2017, Lawrence Livermore National Security, LLC.
+ * Copyright (c) 2020, Lawrence Livermore National Security, LLC.
  * Produced at the Lawrence Livermore National Laboratory.
  *
- * Copyright 2017, UT-Battelle, LLC.
+ * Copyright 2020, UT-Battelle, LLC.
  *
  * LLNL-CODE-741539
  * All rights reserved.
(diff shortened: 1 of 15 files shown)


adammoody:
Thanks, @MichaelBrim .  This will be a nice addition.

client/src/unifyfs-fixed.c
```diff
@@ -270,84 +120,100 @@ void unifyfs_rewrite_index_from_seg_tree(void)
...
+        /* is this the target file? */
+        if ((target_fid != -1) && (fid != target_fid)) {
+            continue;
+        }
 
-    /* flushed, clear buffer and refresh number of entries
-     * and number remaining */
-    clear_index();
+        unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
+        if ((NULL == meta) || (meta->fid != fid)) {
```
tonyhutter, line 177: `(meta->fid != fid)` <- is this check needed?

MichaelBrim: When fds/fids are re-used, the fid field is reset to -1. This is just a paranoid check to make sure the meta is for the current fid.

MichaelBrim: I wanted to put the check in the `unifyfs_get_meta_from_fid()` function itself, but it's used to do get the filemeta pointer when a fid is first allocated (before fid is set).

MichaelBrim:
I just finished testing the various combination of configs on Summitdev. Everything seems to be working well (and write might even be faster). This is ready for review.

common/src/unifyfs_client_rpcs.h
```diff
@@ -102,9 +116,9 @@ DECLARE_MARGO_RPC_HANDLER(unifyfs_metaget_rpc)
...
  *
- * given app_id and client_id as input, read all write extents
- * from client index in shared memory and insert corresponding
- * key/value pairs into our global metadata */
+ * given a client identified by (app_id, client_id) as input, read the write
+ * extents for one or more of the client's files from the shared memory index
+ * and update the global metadata for the file(s) */
 MERCURY_GEN_PROC(unifyfs_sync_in_t,
                  ((int32_t)(app_id))
                  ((int32_t)(client_id)))
```
sandrain, line 124: @MichaelBrim because the sync rpc is per-file basis, can we add a gfid as an additional parameter here?

MichaelBrim: We could, but the gfid can easily be retrieved from the shmem data, so I'm not sure what is the benefit.

sandrain: We can possibly remove the repeated gfid from `unifyfs_index_t`, but maybe we can address later. We might need it back for different sync operations, if ever.

MichaelBrim: I avoided changing the server rpc handler  (`rm_cmd_sync`) mostly to help avoid any conflicts with work going on in margotreenew.

client/src/unifyfs-fixed.c
tonyhutter: Got it, thanks for the info

client/src/unifyfs-fixed.c
```diff
@@ -129,126 +59,46 @@ static void clear_index(void)
...
+                 file_pos + length - 1,
+                 log_pos);
 
     return UNIFYFS_SUCCESS;
 }
 
 /*
  * Remove all entries in the current index and re-write it using the write
- * metadata stored in all the file's seg_trees.  This only re-writes the
+ * metadata stored in target files' seg_trees.  This only re-writes the
```
tonyhutter, line 101: "file's"

MichaelBrim: Thanks, I fixed the comment.

client/src/unifyfs-fixed.c
```diff
@@ -50,76 +50,6 @@
...
-    off_t next_offset = next_idx->file_pos;
-    if (prev_offset != next_offset) {
-        /* index values are not contiguous, can't combine */
-        return UNIFYFS_SUCCESS;
-    }
-
-    /* got contiguous bytes in the same file,
-     * check whether both index values fall in the same slice */
-    off_t prev_slice = prev_idx->file_pos / slice_size;
-    off_t next_slice = next_idx->file_pos / slice_size;
```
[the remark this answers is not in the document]
MichaelBrim: latest push removes this

client/src/unifyfs-fixed.c
```diff
@@ -270,84 +120,100 @@ void unifyfs_rewrite_index_from_seg_tree(void)
...
-     * if we're using that optimization */
-    if (unifyfs_flatten_writes) {
-        unifyfs_rewrite_index_from_seg_tree();
-    }
+    int ret = UNIFYFS_SUCCESS;
 
-    /* if there are no index entries, we've got nothing to sync */
-    if (*unifyfs_indices.ptr_num_entries == 0) {
-        return UNIFYFS_SUCCESS;
+    if (target_fid == -1) {
```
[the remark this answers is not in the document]
MichaelBrim: latest push figures out if any of the extents wrote to spillover, and does a logio sync if so

MichaelBrim:
@adammoody I addressed your comments

Issue #525, opened 2020-06-22
Failure exporting data file created in unifyfs by multi-node execution


clmendes:
@adammoody and @MichaelBrim ,

As I had reported in our latest call, I'm having trouble exporting the datafile that is produced under /unifyfs in a multi-node execution on Catalyst of an MPI program. This error also happens, for example, when I run the example at /examples/src/app-mpiio.c and try to export file testfile using sysio-cp-gotcha: this copy utility hangs, and the exported file is left incomplete.

Logs from the client and from the server show that the client was expecting to read more data, whereas the server on that same node was waiting for work. Thus, it looks like a deadlock scenario.

This problem occurs even running with just two nodes, and one processor per node. I reproduced the problem with the code below, which simply tries to create a 64-byte file using function MPI_File_write_at. This works fine with multiple processors on a single node, but running with two processors on two nodes (i.e. one processor per node) results in a file test.dat that cannot be exported by sysio-cp-gotcha.

Note: the ufs:/unifyfs naming is used such that we can run with Mvapich, according to the hint received from Rob Latham.

Example program:

#define FILESIZE 64
#include <stdio.h>
#include "mpi.h"
#include <unifyfs.h>
int main(int argc, char *argv[])
{
        int ret=0;
        MPI_File fh;
        MPI_Status status;
        MPI_Offset offset;
        int buf [FILESIZE/sizeof(int)], rank, nprocs, nints, count, i;
        MPI_Init (0,0);
        MPI_Comm_rank (MPI_COMM_WORLD, &rank);
        MPI_Comm_size (MPI_COMM_WORLD, &nprocs);
        ret = unifyfs_mount("/unifyfs", rank, nprocs, 0);
        if (ret) {
                fprintf(stderr,"unifyfs_mount failed, rank %d (return = %d)\n",rank,ret);
                exit(-1);
        }
        for (i=0; i<FILESIZE/sizeof(int); i++) buf[i]=rank;
        nints = FILESIZE/(nprocs*sizeof(int));
        offset = rank * nints * sizeof(int);
        /* OPen File collectively */
        MPI_File_open(MPI_COMM_WORLD, "ufs:/unifyfs/test.dat",
                MPI_MODE_CREATE|MPI_MODE_RDWR, MPI_INFO_NULL, &fh);
        if (fh==NULL) fprintf(stderr,"[%d] MPI_File_open ERROR\n",rank);
        /* Write to File */
[...]

PR #529, opened 2020-06-25
Fixes the problem of read becoming unresponsive

sandrain:
Fixes the problem of request manager becoming unresponsive, when handling a read request that needs to fetch data from local and remote servers:

- Fix the request manager
- An example to examine the read operations

### Description

The request manager does not properly process when a read request from a client needs to fetch data both from local and remote servers. The problem does not happen if the data is fully fetched either locally or remotely. Therefore if an application reads data using the same block size while it's written (like most of our example do), the problem does not occur.

Specifically, in the `rm_process_remote_chunk_responses`:

https://github.com/LLNL/UnifyFS/blob/9ed0f1480882fd8a74e7277eaa17446c06c9049e/server/src/unifyfs_request_manager.c#L1732

```
 /* iterate over each active read request */
    for (i = 0; i < RM_MAX_ACTIVE_REQUESTS; i++) {
        server_read_req_t* req = thrd_ctrl->read_reqs + i;
        if ((req->num_remote_reads > 0) &&
            (req->status == READREQ_STARTED)) {
            /* iterate over each delegator we need to send requests to */
            remote_chunk_reads_t* rcr;
            for (j = 0; j < req->num_remote_reads; j++) {
                rcr = req->remote_reads + j;
                if (NULL == rcr->resp) {
                    continue;
                }
                LOGDBG("found read req %d responses from delegator %d",
                       i, rcr->rank);
                rc = rm_handle_chunk_read_responses(thrd_ctrl, req, rcr);
                if (rc != (int)UNIFYFS_SUCCESS) {
                    LOGERR("failed to handle chunk read responses");
                    ret = rc;
                }
            }
```

If the `req` status becomes `READREQ_PARTIAL_COMPLETE`, the request is not further processed by this function. This may fix the issue #525. 

### Types of changes

### DIFF
--- a/examples/src/read-data.c
+++ b/examples/src/read-data.c
@@ -1,0 +1,269 @@
+/*
+ * Copyright (c) 2020, Lawrence Livermore National Security, LLC.
+ * Produced at the Lawrence Livermore National Laboratory.
+ *
+ * Copyright 2019, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyFS.
+ * For details, see https://github.com/LLNL/UnifyFS.
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
+/* read-data: This program aims to test reading data from a file with arbitrary
+ * offset and length. This program can run either interactively (only
+ * specifying filename) or non-interactively (specifying filename with offset
+ * and length).
+ */
+#include <config.h>
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include <errno.h>
+#include <limits.h>
+#include <sys/time.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <fcntl.h>
+#include <libgen.h>
+#include <getopt.h>
+#include <unifyfs.h>
+
+#include "testutil.h"
+
+#define DEFAULT_MOUNTPOINT "/unifyfs"
+
+static char filename[PATH_MAX];
+static char mountpoint[PATH_MAX];
+
+static char* buf;
+static unsigned int bufsize;
+
+static int parse_line(char* line, unsigned long* offset, unsigned long* length)
+{
+    char* pos = NULL;
+
+    line[strlen(line)-1] = '\0';
+
+    if (strncmp("quit", line, strlen("quit")) == 0) {
+        return 1;
+    }
+
+    pos = strchr(line, ',');
+    if (!pos) {
+        return -1;
+    }
+
+    *pos = '\0';
+    pos++;
+
+    *offset = strtoul(line, NULL, 0);
+    *length = strtoul(pos, NULL, 0);
+
+    return 0;
+}
+
+static void alloc_buf(unsigned long length)
+{
+    if (!buf) {
+        bufsize = length;
+        buf = malloc(bufsize);
+    } else {
+        if (bufsize < length) {
+            buf = realloc(buf, length);
+        }
+    }
+
+    if (!buf) {
+        perror("failed to allocate buffer");
+        exit(1);
+    }
+}
+
+static void do_pread(int fd, size_t length, off_t offset)
+{
+    int ret = 0;
+
+    alloc_buf(length);
+
+    errno = 0;
+
+    ret = pread(fd, buf, length, offset);
+
+    printf(" -> pread(off=%lu, len=%lu) = %d", offset, length, ret);
+    if (errno) {
+        printf("  (err=%d, %s)\n", errno, strerror(errno));
+    } else {
+        printf("\n");
+    }
+}
+
+static void run_interactive(int fd)
+{
+    int ret = 0;
+    unsigned long offset = 0;
+    unsigned long length = 0;
+    char* line = NULL;
+    char linebuf[LINE_MAX];
+
+    while (1) {
+        printf("\nread (offset,length)> ");
+        fflush(stdout);
+        line = fgets(linebuf, LINE_MAX-1, stdin);
+        if (!line) {
+            continue;
+        }
+
+        ret = parse_line(line, &offset, &length);
+        if (ret < 0) {
+            continue;
+        } else if (1 == ret) {
+            printf("terminating..\n");
+            break;
+        }
+
+        do_pread(fd, length, offset);
+    }
+}
+
+static struct option long_opts[] = {
+    { "help", 0, 0, 'h' },
+    { "mount", 1, 0, 'm' },
+    { "offset", 1, 0, 'o' },
+    { "length", 1, 0, 'l' },
+    { 0, 0, 0, 0},
+};
+
+static char* short_opts = "hm:o:l:";
+
+static const char* usage_str =
+"\n"
+"Usage: %s [options...] <pathname>\n"
+"\n"
+"Test reading data from a file <pathname>. <pathname> should be a full\n"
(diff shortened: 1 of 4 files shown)


adammoody:
Nice job tracking this down, @sandrain .  Instead of adding a test for PARTIAL_COMPLETE, I was wondering if the PARTIAL_COMPLETE state is actually used.  This moves the request to something other than STARTED but not COMPLETE.

Is this state needed?

If not, we could drop the PARTIAL_COMPLETE state and just go with STARTED.  If we do need it, are there other places where it should be added?

sandrain:
@adammoody In fact, I was wondering if we could just remove it.  The request manager code in this part was a bit complicated, and I didn't look it further.

The `PARTIAL_COMPLETE` status is used when a request (`server_read_req_t`) consists of multiple read requests (`remote_chunk_read_t`), and some of `remote_chunk_read_t` are complete. I think it should be just fine to leave the status as STARTED when partially complete. I will test and see if we can remove the PARTIAL status.

adammoody:
Let's tag @MichaelBrim on this, too.

MichaelBrim:
I honestly can't remember if PARTIAL_COMPLETE is absolutely necessary. I doubt it since we're not actually checking for that exact state in any code. Go ahead and try removing it.

MichaelBrim:
I think we probably need to update the examples (particularly writeread) to support different block sizes for writes vs reads. We would have hit this a long time ago if that case was tested.

adammoody:
> 
> 
> I honestly can't remember if PARTIAL_COMPLETE is absolutely necessary. I doubt it since we're not actually checking for that exact state in any code. Go ahead and try removing it.

Right, @sandrain also double check any code that is comparing to ```== STARTED``` since previous PARTIAL_COMPLETE states will now fall into this set and look for ```!=STARTED``` since that might have been an implicit way to check for PARTIAL_COMPLETE state.

sandrain:
I have removed the PARTIAL status. The following test job goes fine:

1. `write-static`
2. `read-static`
3. `read-data-static` with non-aligned offset/length
4. `sysio-cp-static`, copying out the testfile to an external location

I guess that I've checked all the status checking if statements, please let me know if I missed anything.

sandrain:
> I think we probably need to update the examples (particularly writeread) to support different block sizes for writes vs reads. We would have hit this a long time ago if that case was tested.

@MichaelBrim I think it would be good if we have a test to read with random offset and length (not block or page aligned) within the file length, like (off=15678, len=9976), in a loop.

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+static char* mountpoint = "/unifyfs";
+
+static char linebuf[LINE_MAX];
+static char* buf;
+static unsigned int bufsize;
+
+static int fd;
+static struct stat sb;
+static unsigned int offset;
+static unsigned int length;
```
tonyhutter, line 43: These should be `unsigned long` to match the `%lu` identifiers in `printf(" -> pread(off=%lu, len=%lu) = %d"`

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+    "Available options:\n"
+    "                              (handy for attaching in debugger)\n"
+    " -h, --help                   help message\n"
+    " -m, --mount=<mountpoint>     use <mountpoint> for unifyfs\n"
+    "                              (default: /unifyfs)\n"
+    " -o, --offset=<offset>        read from <offset>\n"
+    " -l, --length=<length>        read <length> bytes\n"
+    "\n";
+
+static char* program;
```
tonyhutter, line 152: `program` needs to be set

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+};
+
+static char* short_opts = "hm:o:l:";
+
+static const char* usage_str =
+    "\n"
+    "Usage: %s [options...] <pathname>\n"
+    "\n"
+    "Available options:\n"
+    "                              (handy for attaching in debugger)\n"
```
tonyhutter, line 144: "(handy for attaching in debugger)" ?

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+
+    ret = pread(fd, buf, length, offset);
+
+    printf(" -> pread(off=%lu, len=%lu) = %d", offset, length, ret);
+    if (errno) {
+        printf("  (err=%d, %s)\n", errno, strerror(errno));
+    } else {
+        printf("\n");
+    }
+}
```
tonyhutter, line 102: `no return statement in function returning non-void`

examples/src/read-data.c
tonyhutter, line 43: Can't the majority of these be local variables?

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * Copyright 2019, UT-Battelle, LLC.
+ *
+ * LLNL-CODE-741539
+ * All rights reserved.
+ *
+ * This is the license for UnifyFS.
+ * For details, see https://github.com/LLNL/UnifyFS.
+ * Please read https://github.com/LLNL/UnifyFS/LICENSE for full license text.
+ */
```
tonyhutter, line 13: Add a comment describing what this program does
"This utility allows you to tests reads on a file.  It can run either interactively or non-interactively"

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+{
+    int ret = 0;
+    int ch = 0;
+    int optidx = 0;
+
+    while ((ch = getopt_long(argc, argv,
+                             short_opts, long_opts, &optidx)) >= 0) {
+        switch (ch) {
+        case 'm':
+            mountpoint = strdup(optarg);
```
tonyhutter, line 170: You need to `free(mountpoint)` somewhere.

examples/src/read-data.c
```diff
@@ -0,0 +1,241 @@
...
+
+        case 'h':
+        default:
+            print_usage();
+            break;
+        }
+    }
+
+    if (argc - optind != 1) {
+        fprintf(stderr, "Usage: %s <filename>\n", argv[0]);
```
tonyhutter, line 189: Maybe call `print_usage()` here instead to convey more information?

sandrain:
@tonyhutter Thanks a lot. I was in a hurry in the afternoon. I've addressed your comments and pushed again.

examples/src/read-data.c
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+
+    if (!buf) {
+        perror("failed to allocate buffer");
+        exit(1);
+    }
+}
+
+static void do_pread(int fd, unsigned long length, unsigned long offset)
```
MichaelBrim, line 85: why aren't you using the actual types for `pread()` parameters (i.e., `off_t offset` and `size_t length`)?

sandrain: Thanks, @MichaelBrim. I've updated it.

examples/src/read-data.c
```diff
@@ -0,0 +1,258 @@
...
+    close(fd);
+
+    if (buf) {
+        free(buf);
+    }
+
+    if (unifyfs) {
+        unifyfs_unmount();
+    }
+
```
tonyhutter, line 256: Add:
`free(program)`

sandrain: @tonyhutter Thanks. I have updated. `man 3 basename` says not to pass the pointer to `free(3)`. So, instead, I used another pointer to keep the string returned from `strdup` to free it later.

tonyhutter: :+1:

PR #536, opened 2020-07-09
Handling large transfer between servers. 

### LINKED ISSUE #532: Handling large I/O request
### System information
Summitdev

The example programs fail with large I/O block/chunk sizes (e.g., blocksize/chunksize 64MB). This appears to be a problem that UnifyFS does not check if the size of the data from a given IO system call (e.g., `read(2)`, `write(2)`) exceeds the client shm buffer. Based on our profiling, there exist lots of scientific applications that perform I/O with large buffers:

(analysis by @fwang2):

> I did this for you, over the last year, we have  15,335 jobs doing read larger than 1G at posix. putting this into perspective though, over summit life time so far, we have 15,512 jobs doing that kind of large read calls; so majority of them starts only this year; and that majority is a very minority of it, out of 9 million job records.

So, properly handling the large request will be critical.

### Describe how to reproduce the problem
Run the `write` and `read` examples with large block/chunk sizes (e.g., `--blocksize=$((128*(1<<20))) --chunksize=$((64*(1<<20)))`). 

The following is the test from summitdev (4 nodes, ppn=1):

`write` successfully writes the data, but the final size becomes 0 (`write-static -b 134217728 -c 67108864 -n 32 -f testfile`):
```
[0] ERROR - file size check failed - actual size is 0 B, expected 17179869184 B

I/O pattern:               N-to-1
I/O block size:            131072.00 KiB
I/O request size:          65536.00 KiB
Number of processes:       4
Each process wrote:        4096.00 MiB
Total data written:        16384.00 MiB
File create time:          0.055255 sec
Maximum write time:        1.891449 sec
Global write time:         1.891488 sec
Maximum sync time:         2.504927 sec
Global sync time:          2.504969 sec
File close time:           0.011243 sec
File laminate time:        0.065025 sec
Full write time:           4.543681 sec
File stat time:            0.029394 sec
Aggregate write bandwidth: 9086.981 MiB/s
Effective write bandwidth: 8662.142 MiB/s
```

`read` simply fails (`read-static -b 134217728 -c 67108864 -n 32 -f testfile`):
```
[0] ERROR - file size check failed - actual size is 0 B, expected 17179869184 B
[0] read() EOF
[0] read req 0 failed
[0] read() EOF
[0] read req 1 failed
[0] read() EOF
[0] read req 2 failed
[0] read() EOF
[0] read req 3 failed
[0] read() EOF
[0] read req 4 failed
[0] read() EOF
[0] read req 5 failed
[0] read() EOF
[0] read req 6 failed
[0] read() EOF
[0] read req 7 failed
[0] read() EOF
[0] read req 8 failed
[0] read() EOF
[0] read req 9 failed
[0] read() EOF
[0] read req 10 failed
[0] read() EOF
[0] read req 11 failed
[0] read() EOF
[0] read req 12 failed
[0] read() EOF
[0] read req 13 failed
[0] read() EOF
[0] read req 14 failed
[0] read() EOF
[0] read req 15 failed
[0] read() EOF
[0] read req 16 failed
[0] read() EOF
[0] read req 17 failed
[0] read() EOF
[0] read req 18 failed
[0] read() EOF
[0] read req 19 failed
[0] read() EOF
[0] read req 20 failed
[0] read() EOF
[0] read req 21 failed
[0] read() EOF
[0] read req 22 failed
[0] read() EOF
[0] read req 23 failed
[0] read() EOF
[0] read req 24 failed
[0] read() EOF
[0] read req 25 failed
[0] read() EOF
[0] read req 26 failed
[0] read() EOF
[0] read req 27 failed
[0] read() EOF
[0] read req 28 failed
[0] read() EOF
[0] read req 29 failed
[0] read() EOF
[0] read req 30 failed
[0] read() EOF
[0] read req 31 failed
[0] read() EOF
[0] read req 32 failed
[0] read() EOF
[0] read req 33 failed
[0] read() EOF
[0] read req 34 failed
[0] read() EOF
[0] read req 35 failed
[0] read() EOF
[0] read req 36 failed
[0] read() EOF
[0] read req 37 failed
[0] read() EOF
[0] read req 38 failed
[0] read() EOF
[0] read req 39 failed
[0] read() EOF
[0] read req 40 failed
[0] read() EOF
[0] read req 41 failed
[0] read() EOF
[0] read req 42 failed
[0] read() EOF
[0] read req 43 failed
[0] read() EOF
[0] read req 44 failed
[0] read() EOF
[0] read req 45 failed
[0] read() EOF
[0] read req 46 failed
[0] read() EOF
[0] read req 47 failed
[0] read() EOF
[0] read req 48 failed
[0] read() EOF
[0] read req 49 failed
[0] read() EOF
[0] read req 50 failed
[0] read() EOF
[0] read req 51 failed
[0] read() EOF
[0] read req 52 failed
[0] read() EOF
[0] read req 53 failed
[0] read() EOF
[0] read req 54 failed
[0] read() EOF
[0] read req 55 failed
[0] read() EOF
[0] read req 56 failed
[0] read() EOF
[0] read req 57 failed
[0] read() EOF
[0] read req 58 failed
[0] read() EOF
[0] read req 59 failed
[0] read() EOF
[0] read req 60 failed
```

The read output is from the rank 0, but other ranks also fail with the same error message.

I think somehow we need to breakdown such large IO requests into smaller requests, perhaps before sending them to the server.

sandrain:
Handling large transfer between servers. This fixes the large data transfer failures (in the mercury tranport layer) between servers.

* `common/unifyfs_const.h`: defined `MAX_BULK_TX_SIZE` (8MB) for data transfer size.
* `src/server/unifyfs_request_manager.c`: fixed request manager to initiate multiple transfers for large bulk data.
* type fix in `examples/src/read-data.c`
* some relevant debugging messages.

### Description
One of the (so far identified) problems in handling a large I/O request is in the bulk data transfer in the RPC layer. The `margo_bulk_transfer` does not consider the size limitation of the underlying transport layer, and fails when the bulk transfer size exceeds the size the underlying transport layer can handle. Specifically, the following bulk transfer in the request manager fails when the transfer size (`in.bulk_size`) becomes larger than the transport layer can handle:

https://github.com/LLNL/UnifyFS/blob/cd0c9b45ea843b5af3e403526057488953bd092b/server/src/unifyfs_request_manager.c#L2313

```
            /* pass along address of buffer we want to transfer
             * data into to prepare it for a bulk write,
             * get resulting margo handle */
            hg_bulk_t bulk_handle;
            hret = margo_bulk_create(mid, 1, (void**)&resp_buf, &in.bulk_size,
                HG_BULK_WRITE_ONLY, &bulk_handle);
            assert(hret == HG_SUCCESS);

            /* execute the transfer to pull data from remote side
             * into our local bulk transfer buffer */
            hret = margo_bulk_transfer(mid, HG_BULK_PULL, hgi->addr,
                in.bulk_handle, 0, bulk_handle, 0, in.bulk_size);
            assert(hret == HG_SUCCESS);
```

Also, the use of `assert(3)` leads to the immediate termination of the server daemon (maybe we need to start remove `assert(3)` calls).

With this fix, the large file read works (e.g., `read(fd, buf, 2GB)`), as long as the request manager can allocate buffer for handling the request (`UNIFYFS_MAX_SPLIT_CNT`). It also looks like there are other problems when many clients (per server, i.,e., large `ppn`) are concurrently performing large read operations.

This patch partially fixes #532.

### Types of changes

### DIFF
--- a/examples/src/read-data.c
+++ b/examples/src/read-data.c
@@ -85,7 +85,7 @@ static void alloc_buf(unsigned long length)
 
 static void do_pread(int fd, size_t length, off_t offset)
 {
-    int ret = 0;
+    ssize_t ret = 0;
 
     alloc_buf(length);
 
@@ -93,7 +93,7 @@ static void do_pread(int fd, size_t length, off_t offset)
 
     ret = pread(fd, buf, length, offset);
 
-    printf(" -> pread(off=%lu, len=%lu) = %d", offset, length, ret);
+    printf(" -> pread(off=%lu, len=%lu) = %zd", offset, length, ret);
     if (errno) {
         printf("  (err=%d, %s)\n", errno, strerror(errno));
     } else {
--- a/server/src/unifyfs_request_manager.c
+++ b/server/src/unifyfs_request_manager.c
@@ -2255,6 +2255,7 @@ static void chunk_read_response_rpc(hg_handle_t handle)
 {
     int32_t ret;
     hg_return_t hret;
+    chunk_read_response_out_t out;
 
     /* get input params */
     chunk_read_response_in_t in;
@@ -2268,6 +2269,9 @@ static void chunk_read_response_rpc(hg_handle_t handle)
     int req_id     = (int)in.req_id;
     int num_chks   = (int)in.num_chks;
     size_t bulk_sz = (size_t)in.bulk_size;
+
+    LOGDBG("received chunk read response from server %d (%d chunks)",
+           src_rank, num_chks);
 
     /* The input parameters specify the info for a bulk transfer
      * buffer on the sending process.  We use that info to pull data
@@ -2306,21 +2310,49 @@ static void chunk_read_response_rpc(hg_handle_t handle)
             hg_bulk_t bulk_handle;
             hret = margo_bulk_create(mid, 1, (void**)&resp_buf, &in.bulk_size,
                 HG_BULK_WRITE_ONLY, &bulk_handle);
-            assert(hret == HG_SUCCESS);
+            if (hret != HG_SUCCESS) {
+                LOGERR("failed to prepare bulk transfer");
+                ret = UNIFYFS_ERROR_MARGO;
+                goto out_respond;
+            }
 
             /* execute the transfer to pull data from remote side
-             * into our local bulk transfer buffer */
-            hret = margo_bulk_transfer(mid, HG_BULK_PULL, hgi->addr,
-                in.bulk_handle, 0, bulk_handle, 0, in.bulk_size);
-            assert(hret == HG_SUCCESS);
+             * into our local bulk transfer buffer.
+             * NOTE: mercury/margo bulk transfer does not check the maximum
+             * transfer size that the underlying transport supports, and a
+             * large bulk transfer may result in failure. */
+            int i = 0;
+            hg_size_t remain = in.bulk_size;
 
-            /* process read replies (headers and data) we just
-             * received */
-            rc = rm_post_chunk_read_responses(app_id, client_id,
-                src_rank, req_id, num_chks, bulk_sz, resp_buf);
-            if (rc != (int)UNIFYFS_SUCCESS) {
-                LOGERR("failed to handle chunk read responses")
-                ret = rc;
+            do {
+                hg_size_t offset = i * MAX_BULK_TX_SIZE;
+                hg_size_t len = remain < MAX_BULK_TX_SIZE
+                                ? remain : MAX_BULK_TX_SIZE;
+
+                hret = margo_bulk_transfer(mid, HG_BULK_PULL, hgi->addr,
+                                           in.bulk_handle, offset,
+                                           bulk_handle, offset, len);
+                if (hret != HG_SUCCESS) {
+                    break;
+                }
+
+                remain -= len;
+                i++;
+            } while (remain > 0);
+
+            if (hret == HG_SUCCESS) {
+                LOGDBG("transferred bulk data (%lu bytes)", in.bulk_size);
+
+                /* process read replies (headers and data) we just received */
+                rc = rm_post_chunk_read_responses(app_id, client_id,
+                        src_rank, req_id, num_chks, bulk_sz, resp_buf);
+                if (rc != (int)UNIFYFS_SUCCESS) {
+                    LOGERR("failed to handle chunk read responses")
+                        ret = rc;
+                }
+            } else {
+                LOGERR("failed to perform bulk transfer");
+                ret = UNIFYFS_ERROR_MARGO;
             }
 
             /* deregister our bulk transfer buffer */
@@ -2328,8 +2360,8 @@ static void chunk_read_response_rpc(hg_handle_t handle)
         }
     }
 
+out_respond:
     /* fill output structure */
-    chunk_read_response_out_t out;
     out.ret = ret;
 
     /* return to caller */
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -1107,7 +1107,7 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
     /*
      * ToDo: Exception handling when some of the requests
      * are missed
-     * */
+     */
 
     /* spin waiting for read data to come back from the server,
      * we process it in batches as it comes in, eventually the
@@ -1121,11 +1121,14 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
         } else {
             tmp_rc = process_read_data(read_reqs, count, &done);
             if (tmp_rc != UNIFYFS_SUCCESS) {
+                LOGERR("failed to process data from server");
                 rc = UNIFYFS_FAILURE;
             }
             delegator_signal();
         }
     }
+
+    LOGDBG("fetched all data from server for %d requests", count);
 
     /* got all of the data we'll get from the server,
      * check for short reads and whether those short
--- a/common/src/unifyfs_const.h
+++ b/common/src/unifyfs_const.h
@@ -54,6 +54,7 @@
 #define REQ_BUF_LEN (MAX_META_PER_SEND * 64) /* chunk read reqs buffer size */
 #define SHM_WAIT_INTERVAL 1000       /* unit: ns */
 #define RM_MAX_ACTIVE_REQUESTS 64    /* number of concurrent read requests */
+#define MAX_BULK_TX_SIZE (8 * MIB)   /* bulk transfer size */
 
 // Server - Service Manager
 #define LARGE_BURSTY_DATA (512 * MIB)


examples/src/read-data.c
```diff
@@ -85,15 +85,15 @@ static void alloc_buf(unsigned long length)
...
+    ssize_t ret = 0;
 
     alloc_buf(length);
 
     errno = 0;
 
     ret = pread(fd, buf, length, offset);
 
-    printf(" -> pread(off=%lu, len=%lu) = %d", offset, length, ret);
+    printf(" -> pread(off=%lu, len=%lu) = %ld", offset, length, ret);
```
MichaelBrim, line 96: ssize_t uses '%zd' for format string, size_t uses '%zu'

sandrain: Thanks, @MichaelBrim. I've updated.

server/src/unifyfs_request_manager.c
```diff
@@ -2306,30 +2310,58 @@ static void chunk_read_response_rpc(hg_handle_t handle)
...
+
+                /* process read replies (headers and data) we just received */
+                rc = rm_post_chunk_read_responses(app_id, client_id,
+                        src_rank, req_id, num_chks, bulk_sz, resp_buf);
+                if (rc != (int)UNIFYFS_SUCCESS) {
+                    LOGERR("failed to handle chunk read responses")
+                        ret = rc;
+                }
+            } else {
+                LOGERR("failed to perform bulk transfer");
```
adammoody, line 2354: Does it make sense to still post a read_response if the transfer fails, but indicate an error status value in the read request?  Would not posting anything lead to a hang while reading?

server/src/unifyfs_request_manager.c
```diff
@@ -2306,30 +2310,58 @@ static void chunk_read_response_rpc(hg_handle_t handle)
             hg_bulk_t bulk_handle;
             hret = margo_bulk_create(mid, 1, (void**)&resp_buf, &in.bulk_size,
                 HG_BULK_WRITE_ONLY, &bulk_handle);
-            assert(hret == HG_SUCCESS);
+            if (hret != HG_SUCCESS) {
+                LOGERR("failed to prepare bulk transfer");
```
adammoody, line 2314: if we should still post a read response as noted below, this might also need to do the same.

server/src/unifyfs_request_manager.c
sandrain: @adammoody Yes, it would be better to inform client that the read has failed. I will update accordingly. Maybe, we will fail the whole operation as EIO, instead of returning partial data.

sandrain: @adammoody I've been looking at the code for handling read operations between the client and server, and I seems like we do not handle any server-side read failures properly at this moment (correct me if I am wrong). The following is the data structures that we have for handling reads:

```
typedef struct {
    size_t nbytes;      /* size of data chunk */
    size_t offset;      /* file offset */
    size_t log_offset;  /* remote log offset */
    int log_app_id;     /* remote log application id */
    int log_client_id;  /* remote log client id */
} chunk_read_req_t;

typedef struct {
    size_t offset;    /* file offset */
    size_t nbytes;    /* requested read size */
    ssize_t read_rc;  /* bytes read (or negative error code) */
} chunk_read_resp_t;

typedef struct {
    int rank;                /* remote delegator rank */
    int rdreq_id;            /* read-request id */
    int app_id;              /* app id of requesting client process */
    int client_id;           /* client id of requesting client process */
    int num_chunks;          /* number of chunk requests/responses */
    readreq_status_e status; /* summary status for chunk reads */
    size_t total_sz;         /* total size of data requested */
    chunk_read_req_t* reqs;  /* @RM: subarray of server_read_req_t.chunks
                              * @SM: received requests buffer */
    chunk_read_resp_t* resp; /* @RM: received responses buffer
                              * @SM: allocated responses buffer */
} remote_chunk_reads_t;

typedef struct {
    size_t length;  /* length of data to read */
    size_t offset;  /* file offset */
    int gfid;       /* global file id */
    int errcode;    /* request completion status */
} client_read_req_t;
```

I think we would need to set the error code to the `chunk_read_resp_t`, but the failure of the bulk transfer means that we have failed to receive this data structure from a remote delegator. I was initially thinking that setting the whole data fetch operations (`remote_chunk_reads_t`) from a specific remote server as a failure, but we would need to add an additional field (or defining a new enum in `readreq_status_e`, like READREQ_FAILED) to do so.

In any cases, the client-side code won't handle the error properly (`unifyfs_gfid_read_reqs`):

https://github.com/LLNL/UnifyFS/blob/cd0c9b45ea843b5af3e403526057488953bd092b/client/src/unifyfs.c#L1107

```

    /*
     * ToDo: Exception handling when some of the requests
     * are missed
     * */

```

So, my opinion is how about separating this problem (propagating server-side read I/O errors correctly to client) for another PR, because it will require some changes in server as well as client. In addition to the margo bulk transfer errors, there are other errors that we would need to handle properly in the `chunk_read_response_rpc` function, e.g., the code path for the `malloc(3)` failure will also skip reporting client, causing the client to wait until the timeout.

I can make a quick/dirty workaround to fail the client read, but also think that it would be better to be done more properly. Let me know how you think.

adammoody:
@sandrain , here is where the client checks for errors on the read replies as they come back from the server:
https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/client/src/unifyfs.c#L702

Note that a single read request from the client can explode into many read replies that come back from the server.  If *any* of those read replies indicate an error, the client marks its request to be in an error state.

sandrain:
> @sandrain , here is where the client checks for errors on the read replies as they come back from the server:
> 
> https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/client/src/unifyfs.c#L702
> 
> Note that a single read request from the client can explode into many read replies that come back from the server. If _any_ of those read replies indicate an error, the client marks its request to be in an error state.

Correct, but we would not have `rep` data returned if the bulk transfer has failed, because it failed while transferring it. So, I actually tried to create the `rep` data structure just for reporting errors, but the client didn't seem to respect it. I will try again.

adammoody:
Just looking for other places where we are tracking error status to help narrow down the cases we're missing...

It looks like the service manager encodes read errors in the read_rc field of the response that it sends back to the request manager here:
https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/server/src/unifyfs_service_manager.c#L197

The request manager checks the read_rc field for errors here:
https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/server/src/unifyfs_request_manager.c#L1938

and then the request manager sets the error field in the shared memory reply that it sends back to the client here:
https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/server/src/unifyfs_request_manager.c#L1953

Maybe we can hook into that path?

Issue #538, opened 2020-07-15
Unresponsive `lio_listio` when multiple `aiocb` structures are given


sandrain:
### System information
Summitdev

### Describe the problem you're observing

`listio_lio` becomes unresponsive, when multiple `aiocb` is fed. I suspect that the resource manager encounters deadlock. Our existing example doesn't encounter this problem (possibly because only a single `aiocb` is passed at a time?).

### Describe how to reproduce the problem

Run the following example application, after creating `tt1` and `tt2` files (e.g., using `write-static` example):

```
#include <stdio.h>
#include <stdlib.h>
#include <stdint.h>
#include <string.h>
#include <errno.h>
#include <limits.h>
#include <sys/time.h>
#include <sys/types.h>
#include <aio.h>
#include <unistd.h>
#include <fcntl.h>
#include <libgen.h>
#include <assert.h>
#include <getopt.h>
#include <unifyfs.h>

static char buf[4*(1<<20)];

static int do_test_simple(void)
{
    int i;
    int ret;
    struct aiocb aio[2] = { 0, };
    struct aiocb *aio_list[2];
    int fd1, fd2;
    struct timeval t1, t2;
    double elapsed = 0.0;

    fprintf(stderr, "## == simple listio ==\n");

    fd1 = open("/unifyfs/tt1", O_RDONLY);
    fd2 = open("/unifyfs/tt2", O_RDONLY);
    assert(fd1 > 0 && fd2 > 0);

    for (i = 0; i < 2; i++) {
        int fd = i == 0 ? fd1 : fd2;
        unsigned long offset = i*(4*(1<<20));
        unsigned long count = 1<<20;
        struct aiocb *current = &aio[i];

        current->aio_fildes = fd;
        current->aio_offset = offset;
        current->aio_nbytes = count;
        current->aio_buf = &buf[i*(1<<20)];
        current->aio_lio_opcode = LIO_READ;

        aio_list[i] = current;

        fprintf(stderr, "## aio[%d] (fd=%d, offset=%lu, nbytes=%lu)\n",
                        i, current->aio_fildes,
                        current->aio_offset, current->aio_nbytes);
        fflush(stdout);
    }

    gettimeofday(&t1, NULL);
    fprintf(stderr, "## calling lio_listio..\n");

    ret = lio_listio(LIO_WAIT, aio_list, 2, 0);

    gettimeofday(&t2, NULL);
[...]

PR #539, opened 2020-07-17
Fixing the race condition between server and client during mread.

sandrain:
When multiple file descriptors/files are involved in a single lio_listio call,
UnifyFS hangs due to incorrect coordination for exchanging data between
server and client. This patch fixes it.

- client side (unifyfs-sysio.c and unifyfs.c): correctly identify the
request completion of individual files.
- server side (unifyfs_cmd_handler.c) : send mread rpc response before
handling the request.
- also removing some extra global lock statements in the request manager
(unifyfs_request_manager.c).
- some extra debugging messages

### Description
When `lio_listio(3)` is requested with multiple files, UnifyFS becomes unresponsive as described in #538. This happens primarily for two reasons:

1. upon receiving the mread RPC, the server thread may send data before it replies to the client:

https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/server/src/unifyfs_cmd_handler.c#L453

```
    /* get list of read requests */
    hret = margo_bulk_transfer(mid, HG_BULK_PULL, hgi->addr,
                               in.bulk_handle, 0, bulk_handle, 0, size);
    assert(hret == HG_SUCCESS);

    /* initiate read operations to fetch data for read requests */
    int ret = rm_cmd_mread(in.app_id, in.client_id,
                           in.read_count, buffer);

    /* build our output values */
    unifyfs_mread_out_t out;
    out.ret = ret;

    /* return to caller */
    hret = margo_respond(handle, &out);
    assert(hret == HG_SUCCESS);
```

Specifically, the execution of `rm_cmd_mread` may cause the request manager thread to send data (before margo handler thread responding to this rpc). This blocks the progress in the request manager (waiting for client to consume the shm data while holding the global lock, while the client is waiting for the server response).

As @MichaelBrim  suggested, a fundamental fix would be letting other thread (instead of this margo handler thread) execute the code of `rm_cmd_mread`, which is quite heavy/complex for a rpc handler thread.

2. the client does not correctly account the 'data completion' notice from the server:

The server sends the data completion message (via shmem buffer), when handling a request for a file completes. The current client code of fetching the data does not correctly detect the data completion. 

https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/client/src/unifyfs.c#L1115

```
    /* spin waiting for read data to come back from the server,
     * we process it in batches as it comes in, eventually the
     * server will tell us it's sent us everything it can */
    int done = 0;
    while (!done) {
        int tmp_rc = delegator_wait();
        if (tmp_rc != UNIFYFS_SUCCESS) {
            rc = UNIFYFS_FAILURE;
            done = 1;
        } else {
            tmp_rc = process_read_data(read_reqs, count, &done);
            if (tmp_rc != UNIFYFS_SUCCESS) {
                LOGERR("failed to process data from server");
                rc = UNIFYFS_FAILURE;
            }
            delegator_signal();
        }
    }
```

When multiples files are involved in a single `lio_listio`, this loop terminates before receiving all data.

This patch primarily addresses those issues.

### Types of changes

### DIFF
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -1581,6 +1581,7 @@ int UNIFYFS_WRAP(lio_listio)(int mode, struct aiocb* const aiocb_list[],
                     reqs[reqcnt].nread   = 0;
                     reqs[reqcnt].errcode = EINPROGRESS;
                     reqs[reqcnt].buf     = (char*)(cbp->aio_buf);
+                    reqs[reqcnt].aiocbp  = cbp;
                     reqcnt++;
                 }
             } else {
@@ -1609,12 +1610,10 @@ int UNIFYFS_WRAP(lio_listio)(int mode, struct aiocb* const aiocb_list[],
         }
 
         /* update aiocb fields to record error status and return value */
-        ndx = 0;
         for (i = 0; i < reqcnt; i++) {
-            char* buf = reqs[i].buf;
-            for (; ndx < nitems; ndx++) {
+            for (ndx = 0; ndx < nitems; ndx++) {
                 cbp = aiocb_list[ndx];
-                if ((char*)(cbp->aio_buf) == buf) {
+                if (cbp == reqs[i].aiocbp) {
                     AIOCB_ERROR_CODE(cbp) = reqs[i].errcode;
                     if (0 == reqs[i].errcode) {
                         AIOCB_RETURN_VAL(cbp) = reqs[i].length;
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -665,6 +665,10 @@ static int process_read_data(read_req_t* read_reqs, int count, int* done)
         char* rep_buf = shmptr;
         shmptr += rep->length;
 
+        LOGDBG("processing data response from server: "
+               "[%zu] (gfid=%d, offset=%lu, length=%lu, errcode=%d)",
+               i, rep->gfid, rep->offset, rep->length, rep->errcode);
+
         /* get start and end offset of reply */
         size_t rep_start = rep->offset;
         size_t rep_end   = rep->offset + rep->length;
@@ -746,6 +750,8 @@ static int process_read_data(read_req_t* read_reqs, int count, int* done)
             char* req_ptr = req->buf + req_offset;
             char* rep_ptr = rep_buf  + rep_offset;
             memcpy(req_ptr, rep_ptr, length);
+
+            LOGDBG("copied data to application buffer (%lu bytes)", length);
 
             /* update max number of bytes we have written to in the
              * request buffer */
@@ -1047,6 +1053,9 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
     /* prepare our shared memory buffer for delegator */
     delegator_signal();
 
+    /* for mread, we need to manually track the rpc progress */
+    unifyfs_mread_rpc_ctx_t mread_ctx = { 0, };
+
     /* we select different rpcs depending on the number of
      * read requests */
     if (count > 1) {
@@ -1078,7 +1087,7 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
                count, buffer, size);
 
         /* invoke multi-read rpc */
-        read_rc = invoke_client_mread_rpc(count, size, buffer);
+        read_rc = invoke_client_mread_rpc(count, size, buffer, &mread_ctx);
 
         /* free flat buffer resources */
         flatcc_builder_clear(&builder);
@@ -1113,6 +1122,7 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
      * we process it in batches as it comes in, eventually the
      * server will tell us it's sent us everything it can */
     int done = 0;
+    int rpc_done = 0;
     while (!done) {
         int tmp_rc = delegator_wait();
         if (tmp_rc != UNIFYFS_SUCCESS) {
@@ -1126,6 +1136,30 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
             }
             delegator_signal();
         }
+
+        /* if this was mread, track the progress */
+        if (count > 1 && !rpc_done) {
+            tmp_rc = unifyfs_mread_rpc_status_check(&mread_ctx);
+            if (tmp_rc < 0) {
+                LOGERR("failed to check the rpc progress");
+                continue;
+            }
+
+            /* if we received a response from the server, check any errors.
+             * for any errors, we do not have to wait for the data anymore. */
+            if (tmp_rc) {
+                LOGDBG("received rpc response from the server (ret=%d)",
+                       mread_ctx.rpc_ret);
+
+                if (mread_ctx.rpc_ret != UNIFYFS_SUCCESS) {
+                    LOGERR("mread rpc failed on server (ret=%d)",
+                           mread_ctx.rpc_ret);
+                    return UNIFYFS_FAILURE;
+                }
+
+                rpc_done = 1;
+            }
+        }
     }
 
     LOGDBG("fetched all data from server for %d requests", count);
@@ -1136,6 +1170,11 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
     for (i = 0; i < count; i++) {
         /* get pointer to next read request */
         read_req_t* req = &read_reqs[i];
+
+        /* no error message was received from server, set it success */
+        if (req->errcode == EINPROGRESS) {
+            req->errcode = UNIFYFS_SUCCESS;
+        }
 
         /* if we hit an error on our read, nothing else to do */
         if (req->errcode != UNIFYFS_SUCCESS) {
--- a/server/src/unifyfs_request_manager.c
+++ b/server/src/unifyfs_request_manager.c
@@ -228,9 +228,8 @@ void free_value_array(unifyfs_val_t** array)
 static void debug_print_read_req(server_read_req_t* req)
 {
     if (NULL != req) {
-        LOGDBG("server_read_req[%d] status=%d, gfid=%d, num_remote=%d",
-               req->req_ndx, req->status, req->extent.gfid,
-               req->num_remote_reads);
+        LOGDBG("server_read_req[%d] status=%d, num_remote=%d",
+               req->req_ndx, req->status, req->num_remote_reads);
     }
 }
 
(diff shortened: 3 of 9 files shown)


client/src/unifyfs.c
```diff
@@ -654,6 +654,8 @@ static int process_read_data(read_req_t* read_reqs, int count, int* done)
     /* get number of read replies in shared memory */
     size_t num = shm_hdr->meta_cnt;
 
+    LOGDBG("received %lu response from server", num);
```
MichaelBrim, line 657: "%zu" for size_t

sandrain: Thanks @MichaelBrim . I have fixed other places where I made the same error.

client/src/unifyfs.c
```diff
@@ -698,60 +704,54 @@ static int process_read_data(read_req_t* read_reqs, int count, int* done)
             }
 
             /* this reply overlaps with the request, check that
-             * we didn't get an error */
-            if (rep->errcode != UNIFYFS_SUCCESS) {
```
adammoody, line 702: This check on setting the read request errcode field is important, because a single read request could have two replies.  If the first reply has an error, but the second one succeeded, we still need to be sure the read request ends up marked with an error.  Basically, if *any* reply has an error for the read request, the read request should have an error.

client/src/unifyfs-sysio.c
```diff
@@ -1609,27 +1609,31 @@ int UNIFYFS_WRAP(lio_listio)(int mode, struct aiocb* const aiocb_list[],
...
-        ndx = 0;
         for (i = 0; i < reqcnt; i++) {
-            char* buf = reqs[i].buf;
-            for (; ndx < nitems; ndx++) {
+            read_req_t* req = &reqs[i];
+            char* buf = req->buf;
+
+            for (ndx = 0; ndx < nitems; ndx++) {
                 cbp = aiocb_list[ndx];
                 if ((char*)(cbp->aio_buf) == buf) {
```
adammoody, line 1618: I know you didn't change this line, but I feel like this check may not be a reliable way to match the request.  Is it allowed for two different requests to each provide the same user buffer address?  That wouldn't make a lot of sense, but if it is allowed, then we'll need to find some other way to match the read_request to the aio structure.  Just thought I'd raise this in passing.  If we deem it to be a problem, we can create an issue to fix it later.

server/src/unifyfs_cmd_handler.c
```diff
@@ -479,18 +479,25 @@ static void unifyfs_mread_rpc(hg_handle_t handle)
...
 
-    /* build our output values */
+    /* If we have successfully fetched requests, send the response first.
+     * Otherwise, the subsequent rm_cmd_mread() may try to send the data to the
+     * client, while the client is still waiting for our response.
+     * The read I/O results are reported via the shared memory (not through
+     * this rpc response).  */
     unifyfs_mread_out_t out;
-    out.ret = ret;
+    out.ret = 0;
```
adammoody, line 489: I haven't checked, but generally we want to return UNIFYFS_SUCCESS instead of 0.  In theory, UNIFYFS_SUCCESS might be redefined at some point to be some value other than 0.

However, the existing line was here to catch errors returned from the call to rm_cmd_mread to indicate to the caller that the request to the server failed to even get started, and so then the client knows that the server would not be sending any replies back at all.  Are we still accounting for that case?

adammoody: We could report those kind of "failed to launch" errors to the client through another mechanism.  Perhaps the client could always expect replies to come back, and have the server immediately send those replies with error status?

client/src/unifyfs.c
```diff
@@ -1109,21 +1149,31 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
...
+    int gfid_done = 0;
+    while (gfid_done < gfid_count) {
         int tmp_rc = delegator_wait();
         if (tmp_rc != UNIFYFS_SUCCESS) {
             rc = UNIFYFS_FAILURE;
-            done = 1;
+            break;
         } else {
+            int done = 0;
             tmp_rc = process_read_data(read_reqs, count, &done);
```
adammoody, line 1167: I was hoping that ```read_reqs``` could contain requests that could span over multiple files and then the server would signal its down with all requests.

I guess the server is only responding with one file at a time?  Can we fix things on the server side instead?  It sounds like the server is telling the client it is done too early.

client/src/unifyfs-sysio.c
adammoody: Oh just thought of a valid case.  This might be a reasonable thing to do if a process writes the same user buffer to two different files or even to two different offsets in the same file.

client/src/unifyfs.c
sandrain: @adammoody It seems like the server doesn't have a data structure to represent a `lio_listio` request consisting of multiple requests with multiple files (file descriptors). The server only tracks the request per gfid (`rm_cmd_mread` and `create_gfid_chunk_read`):

https://github.com/LLNL/UnifyFS/blob/b4ea3eeceac1c9bf4dffc95ac7479ae3003bd3e3/server/src/unifyfs_request_manager.c#L1313

I will take a 2nd look as you suggested, but it may cause more changes.

sandrain:
@CamStan also, I cannot tag a label here (wanted to put WIP here).

CamStan:
Hmm, yeah, those are some frustrating side effects. I'll get the mirror working as soon as I can to make sure things are set up properly. Then it might be worth considering giving everyone write access again until they work through making things more accessible.

adammoody:
Thanks for reporting these problems, @sandrain .  This is good feedback we can share with the team setting the rules for access for the gitlab test infrastructure.

server/src/unifyfs_cmd_handler.c
sandrain: @adammoody I see your point. Actually, I first tried with async rpcs (iforward/irespond) but later removed, because I thought the result of rm_cmd_mread was not too necessary. I will change back to the iforward, so we can also track the rm_cmd_mread.

sandrain:
@adammoody I've changed the data structures on the server side (instead of client tracking server behavior). Also, I've changed the mread rpc to use `margo_iforward`. Please review again. Also tagging @MichaelBrim for review.

server/src/unifyfs_request_manager.c
```diff
@@ -1914,26 +1903,24 @@ int rm_handle_chunk_read_responses(reqmgr_thrd_t* thrd_ctrl,
     client_shm = clnt->shmem_data;
     shm_hdr = (shm_data_header*) client_shm->addr;
 
-    RM_LOCK(thrd_ctrl);
-
     num_chks = del_reads->num_chunks;
-    gfid = rdreq->extent.gfid;
+    //gfid = rdreq->extent.gfid;
```
MichaelBrim, line 1907: delete this commented line

client/src/unifyfs-sysio.c
MichaelBrim: we should capture this defect in an issue and resolve separately.

sandrain:
@MichaelBrim @adammoody  I've added an extra member to the `read_req_t` to store the original pointer to the `struct aiodb`. I think this would be better than comparing the buffer pointers.

client/src/unifyfs-sysio.c
MichaelBrim: this seems to be resolved with the addition of the aiocb pointer

server/src/unifyfs_cmd_handler.c
sandrain: @adammoody I've resolved this by using iforward in client.

client/src/unifyfs.c
sandrain: @adammoody in the new commit, I've changed the server data structures so it can send a single complete message per client request.

client/src/unifyfs.c
sandrain: @adammoody I kept this code in the new commit.

adammoody:
Thanks @sandrain .  Proactively merged this.  I'll still take a look at this.  I'll just open an issue if I find anything.

client/src/unifyfs.c
```diff
@@ -1126,6 +1136,30 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
             }
             delegator_signal();
         }
+
+        /* if this was mread, track the progress */
+        if (count > 1 && !rpc_done) {
+            tmp_rc = unifyfs_mread_rpc_status_check(&mread_ctx);
```
adammoody, line 1142: Can there be a race condition in here where the ```done``` flag might get set (because the server responds with the data) before we get the return code from the mread rpc and set ```rpc_done```?  If so, we could add a check for rpc_done after this loop and check the status until we get the answer back.

client/src/unifyfs-sysio.c
```diff
@@ -1581,6 +1581,7 @@ int UNIFYFS_WRAP(lio_listio)(int mode, struct aiocb* const aiocb_list[],
                     reqs[reqcnt].nread   = 0;
                     reqs[reqcnt].errcode = EINPROGRESS;
                     reqs[reqcnt].buf     = (char*)(cbp->aio_buf);
+                    reqs[reqcnt].aiocbp  = cbp;
```
adammoody, line 1584: Thanks for adding this fix.

client/src/unifyfs.c
```diff
@@ -1113,6 +1122,7 @@ int unifyfs_gfid_read_reqs(read_req_t* in_reqs, int in_count)
      * we process it in batches as it comes in, eventually the
      * server will tell us it's sent us everything it can */
     int done = 0;
+    int rpc_done = 0;
     while (!done) {
         int tmp_rc = delegator_wait();
```
adammoody, line 1127: The client can get stuck in here in a tight loop waiting for the server to reply with data here.  If the rpc causes a server error, will the client exit this?  I haven't checked the code thoroughly yet, but that was the worry before.

adammoody: Ugh, that time out we do have in delegator_wait is sort of meant to detect a dead server.  Looking at that, it can introduce another race condition that we should fix up.  Consider this situation.

The client issues a read rpc to server.  The server receives the rpc from the client and starts the read operation (successfully).  But... the server is slow to get results back to client and that 5 sec timeout expires.  The client assumes the server is dead and returns from ```delegator_wait``` with an error, which causes the client to exit this while loop early.  However, the server doesn't know that the client gave up waiting, and in the meantime, the server gets the results back from the read and it starts writing the read replies into the shared memory region.  The server might then deadlock waiting on the client if it fills the buffer, or the client might process those read replies for another read operation it issued later, which could lead to corrupt read results on the client side.

We should perhaps treat that "dead server" timeout to be a fatal error.  Or we need to fix up the hand-shaking between the client/server some here.

client/src/unifyfs.c
sandrain: @adammoody I see your point. In fact, I did put the rpc checking routine after this loop in my first try. But, the problem was that there were times that the loop wouldn't finish if the server somehow failed to send all data as expected. The reason that I check rpc reply inside this loop is to release the client from checking the data that has failed anyway.

client/src/unifyfs.c
sandrain: There are probably many cases that we miss here in terms of handshaking between the server and client. Once the server starts fetching data, I do not think the server ever 'aborts/fails' a read request. And, like you said, the client cannot inform the server that 'I do not fetch data anymore.'  Basically, I do not think we have a clear way to handle exceptional cases gracefully between server and client. We need to fix that.

Also, I've tested reading random data back from clients, and the performance variance was huge. The following is the result from 8 compute nodes (each process executing 50 read operations of (random offset, 4MB)).

PPN | average | min | max | std
-- | -- | -- | -- | --
1 | 1.585 | 0.005 | 11.253 | 2.006
2 | 1.444 | 0.004 | 28.753 | 3.398
4 | 3.969 | 0.004 | 78.960 | 11.843

You can see a single 4MB read sometimes took more than 2mins when ppn was 4. I've also tested with ppn=8 (total 64 processes), and 55 processes couldn't even complete (50 times X 4MB read operations) in 30 minutes. I think we also need to rethink the server process model. It seems like a single delegator becomes a bottleneck when we have many processes per node, i.e., a large ppn.

I am bringing up this, because we may want to do something more than fixing the race between the server and client.

client/src/unifyfs.c
sandrain: I will probably register an issue about remaining issues on client/server coordination for handling read.

PR #545, opened 2020-07-28
Allow runtime configuration of max clients per application

MichaelBrim:
### Description

Previously, we had a compile-time limit of 64 clients per application (where application == mountpoint). This PR makes it a server configuration option that can be set at runtime via the environment variable UNIFYFS_SERVER_MAX_APP_CLIENTS. I also upped the default number of clients to 256.

### How Has This Been Tested?

Tested on Summitdev by running 3 back-to-back writeread examples using the new default and a low limit (25). In the latter case, the limit was properly enforced.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_server.c
+++ b/server/src/unifyfs_server.c
@@ -59,6 +59,7 @@ unifyfs_cfg_t server_cfg;
 
 static ABT_mutex app_configs_abt_sync;
 static app_config* app_configs[MAX_NUM_APPS]; /* list of apps */
+static size_t clients_per_app = MAX_APP_CLIENTS;
 
 /**
  * @brief create a ready status file to notify that all servers are ready for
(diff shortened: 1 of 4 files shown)


MichaelBrim:
@sandrain This should interest you, care to review?

server/src/unifyfs_server.c
```diff
@@ -640,6 +651,14 @@ app_config* new_application(int app_id)
     for (int i = 0; i < MAX_NUM_APPS; i++) {
```
sandrain, line 651: Overall, this looks good to me. Just one question in this loop. If we cannot find any empty slot for the new app client request, is there a reason that we need to keep the old app data structures, instead of victimizing the oldest one and allowing the new app to connect; e.g, is it necessary to locate data (in logio)?

MichaelBrim: MAX_NUM_APPS is still hard-coded, and represents the maximum number of application mountpoints. If you run k apps all with the same mountpoint (e.g. '/unifyfs'), it only counts as one app_config. The new configuration setting controls the maximum number of client processes that can mount at the same mountpoint, regardless of how many client applications you run. The new default is 256, so to exhaust that you could do 4 client applications each with 64 ppn, or 8 with 32ppn, or 16 with 16ppn.

sandrain: Now I understand. Thanks for explanation!

PR #566, opened 2020-11-09
Use tree-based group rpcs for file operations (aka margotree)

sandrain:
### Description
This commit squashes all changes from the margotreenew branch. A list of notable updates follows.

* The implementation of file operations within the server is now similar to the Linux VFS layer, where a new unifyfs_fops struct contains the function pointers for the various operations such as metaget, metaset, sync, truncate, laminate, read, etc. See `server/src/unifyfs_fops.h`

* The new distributed metadata implementation of file operations that leverages tree-based margo group rpcs is defined in `server/src/unifyfs_fops_rpc.c` . This implementation introduces the `unifyfs_inode` structure that keeps track of the metadata for a particular file identified by gfid. A `unifyfs_inode_tree` structure maps gfids to their corresponding inode. Currently, one server, the "owner", maintains the metadata for a given file until it is laminated, and all writes that have been synced are visible to readers. Upon lamination or truncation, updated file metadata is broadcast to the other servers.  After lamination, all servers can handle metadata queries locally. The tree-based rpc logic is defined in `unifyfs_tree.[ch]` and `unifyfs_group_rpc.[ch]`.

* The previous MDHIM-based support for file operations is maintained by `server/src/unifyfs_fops_mdhim.c`, which can be selected at configure time using `'--enable-mdhim'`. Note it may not be currently in a usable state.

* The example programs have been updated to enable additional functionality test capabilities:
  - testutil now supports options to truncate the target file before or after writes
  - new sysio examples were added to test create/open, truncate, and unlink

* Includes various bug fixes for problems encountered during testing:
  - single server startup hangs when using the unifyfs CLI tool
  - race condition on server pidfile creation
  - truncate filesize issues due to client caching of extents
  - `lio_listio` detection during configure
  - use argobots mutex for svcmgr state synchronization
  - use separate argobots mutex for reqmgr request locking
  - handle EINTR during `posix_fallocate`
  - `stat` file time fixes
  - segfault in remote read creation
  - remove logio spill files on server exit
  - avoid double free on `client->reqmgr`

* The support for available Mercury NA plugins for internode RPC and RMA has been improved to allow for both OFI and BMI, and the currently available plugin is determined by examining the NA configuration (`na_config.h`).

* Finally, the test suite has been updated to account for new test passes where things had previously been marked as skipped or todo

### Types of changes


MichaelBrim:
@sandrain Looks like we have some checkpatch cleanup to do, and some unit test failures to investigate.

sandrain:
@MichaelBrim yep. i will see what's happening.

client/src/unifyfs-dirops.h
```diff
@@ -1,8 +1,8 @@
 /*
- * Copyright (c) 2017, Lawrence Livermore National Security, LLC.
```
tonyhutter, line 2: In some commercial software I used to work on, and with ZFS code, the rule was to add the year to the copyright line (like `Copyright (c) 2017, 2020`) rather than update the year.  I'm not a lawyer, and I have no idea if that's what actually should be done, or even if it's relevant here, just wanted to mention it.

common/src/seg_tree.c
```diff
@@ -448,7 +511,10 @@ seg_tree_iter(struct seg_tree* seg_tree, struct seg_tree_node* start)
 void
 seg_tree_rdlock(struct seg_tree* seg_tree)
 {
-    assert(pthread_rwlock_rdlock(&seg_tree->rwlock) == 0);
+    int rc = pthread_rwlock_rdlock(&seg_tree->rwlock);
+    if (rc) {
+        LOGERR("pthread_rwlock_rdlock() failed - rc=%d", rc);
+    }
```
tonyhutter, line 517: I think the extra logging you added is good, but you should also keep the `assert()` so that it stops execution.  I'm worried you could get seg_tree corruption if the program continues without properly acquiring the lock.

tonyhutter: In fact, in a separate PR, we may want to consider changing the `asserts()` to an `exit(1)` just in case a user builds with NDEBUG (to disable the asserts()).

tonyhutter:
One thing that stands out is that the extent_tree and unifyfs_inode code looks suspiciously like copy-n-pasted seg_tree code.   I think we could just modify the existing seg_tree code to be more generic, and re-use it for the extent_tree and unifyfs_inode.  Specifically, we could add a a private data field to `struct seg_tree_node`:

```C
struct seg_tree_node {
    RB_ENTRY(seg_tree_node) entry;
    unsigned long start; /* starting logical offset of range */
    unsigned long end;   /* ending logical offset of range */
    unsigned long ptr;   /* physical offset of data in log */
    void *data;
};
```

Whereby you could assign `data` to whatever tree specfic data structures you needed, like:

```C
struct extent_tree_data {
    int svr_rank;        /* rank of server hosting data */
    int app_id;          /* application id (namespace) on server rank */
    int cli_id;          /* client rank on server rank */
} extent_tree_data;

seg_tree_node->data = &extent_tree_data;
```

So your extent_tree functions would be modified to use a seg_trees under the covers, like:

```C
/* Allocate a node for the range tree.  Free node with free() when finished */
static struct seg_tree_node* extent_tree_node_alloc(
    unsigned long start, /* logical starting offset of extent */
    unsigned long end,   /* logical ending offset of extent */
    int svr_rank,        /* rank of server hosting data */
    int app_id,          /* application id (namespace) on server rank */
    int cli_id,          /* client rank on server rank */
    unsigned long pos)   /* physical offset of data in log */
{
    /* allocate a new node structure */
    struct extent_tree_data* data = calloc(1, sizeof(*data));
    if (!node) {
        return NULL;
    }

    data->svr_rank = svr_rank
    data->app_id = app_id;
    data->cli_id = cli_id;

    /* Allocate a new seg_tree node with start, end, and data fields filled in */
    struct seg_tree_node *node = seg_tree_node_alloc(start, end, data);

    return node;
}
```

That way we can leverage all the existing seg_tree code (which we know works, and already has test cases).

tonyhutter:
There may be a bug in `seg_tree_remove()`.  I tried adding this test case:

```diff
diff --git a/t/common/seg_tree_test.c b/t/common/seg_tree_test.c
index 7176eda..74c08c1 100644
--- a/t/common/seg_tree_test.c
+++ b/t/common/seg_tree_test.c
@@ -197,6 +197,21 @@ int main(int argc, char** argv)
     ok(max == 150, "max is 150 (got %lu)", max);
     ok(count == 1, "count is 1 (got %lu)", count);
 
+    seg_tree_clear(&seg_tree);
+    seg_tree_add(&seg_tree, 0, 0, 0);
+    seg_tree_add(&seg_tree, 1, 10, 101);
+    seg_tree_add(&seg_tree, 20, 30, 20);
+    seg_tree_add(&seg_tree, 31, 40, 131);
+
+    /* Remove a single entry */
+    seg_tree_remove(&seg_tree, 0, 0);
+    ok(1==1, "removed a single entry, got %s", print_tree(tmp, &seg_tree));
+    is("[1-10:101][20-30:20][31-40:131]", print_tree(tmp, &seg_tree), "removed a single range, got %s", print_tree(tmp, &seg_tree));
+
+    /* Remove a range spanning the two bordering ranges [20-30] & [31-40]. */
+    seg_tree_remove(&seg_tree, 25, 31);
+    is("[1-10:101][20-24:20][32-40:132]", print_tree(tmp, &seg_tree), "removed a range that truncated two entries, got %s", print_tree(tmp, &seg_tree));
+
     seg_tree_clear(&seg_tree);
     seg_tree_destroy(&seg_tree);
```

The 2nd part of the test tries to truncate [25-31] from the [20-30][31-40] entries.  It correct changes the segment start/end points to [20-24][32-40], but doesn't update `node->ptr` for the [31-40] entry.

tonyhutter:
This seems to fix it for me
```diff
diff --git a/common/src/seg_tree.c b/common/src/seg_tree.c
index 7f1578a..c47e3e1 100644
--- a/common/src/seg_tree.c
+++ b/common/src/seg_tree.c
@@ -358,6 +358,8 @@ int seg_tree_remove(
                  * update node start */
                 LOGDBG("updating node start from %lu to %lu",
                        node->start, (end + 1));
+
+                node->ptr += (end + 1 - node->start);
                 node->start = end + 1;
             }
         } else if (node->start < start) {
```

client/src/unifyfs-dirops.h
sandrain: Thanks, I've addressed this.

common/src/seg_tree.c
sandrain: @tonyhutter @MichaelBrim How about letting these functions return `rc` and fixing callers to handle the error?

sandrain:
Thanks @tonyhutter. I've added your segment tree test and fix. I've moved on to refactoring the segment tree.

adammoody:
While they share some code, there are a lot of differences among the seg_tree, extent_tree, and inode_tree implementations.  Whether we refactor or not, we'll need to define a different set of tests for each tree.

For example, when adding a new node, seg_tree and extent_tree can both coalesce contiguous entries but inode does not.  Further, the logic used to decide whether to coalesce two entries in seg_tree is different from extent_tree.  That means that each tree needs a custom "add" function.  Similarly, each needs a custom "remove".

While I agree that cut-and-paste is not the best programming practice, by the time we refactor things, I suspect we'll end up with something that has a lot of macros or a lot of function pointers.  In my opinion, none of those options are great programming practices, but I find cut-and-pasted code to be easier to read, debug, and maintain.

sandrain:
@adammoody @tonyhutter I've looked the code more carefully. As @adammoody pointed out, we probably cannot refactor nicely but will end up having separate functions for each module. The core part is already extracted in the rbtree code (tree.h), so I think it should be okay to leave as they are for now.

tonyhutter:
> While I agree that cut-and-paste is not the best programming practice, by the time we refactor things, I suspect we'll end up with something that has a lot of macros or a lot of function pointers. In my opinion, none of those options are great programming practices, but I find cut-and-pasted code to be easier to read, debug, and maintain.

> As @adammoody pointed out, we probably cannot refactor nicely but will end up having separate functions for each module.

I respectfully disagree, but I'll bow to the majority on this.

The only thing I ask is that when you do copy-in-paste the seg_tree.c code, keep the comments intact.  There are a number of cases where the capitalization and punctuation were strangely removed.  Here's some examples:

```C
//
// seg_tree.c
//
    /*
     * This function is only called when we know that segment 1 and segment 2
     * overlap with each other. Find first portion of segment 1 that does not
     * overlap with segment 2, if any.
     */
...
            /*
             * If the non-overlapping part came from the front portion of the
             * existing range, then there is a trailing portion of the
             * existing range to add back to be considered again in the next
             * loop iteration.
             */


//
// extent_tree.c
//
    /* this function is only called when we know that segment 1
     * and segment 2 overlap with each other, find first portion
     * of segment 1 that does not overlap with segment 2, if any */
...
            /* if the non-overlapping part came from the front
             * portion of the existing range, then there is a
             * trailing portion of the existing range to add back
             * to be considered again in the next loop iteration */
```

adammoody:
The capitalization and punctuation weren't "removed" -- they were never there to begin with.  The extent_tree.c file was copied from an earlier version of seg_tree.c before those changes had been applied to seg_tree.  If we keep extent_tree.c, we could reapply those changes.

sandrain:
@tonyhutter @adammoody About the comments in seg tree and extent tree, it might have wrapped/reformatted when I cleanup the code using vim (gq command). Possibly, I only did that for the extent tree, because I haven't directly touched the seg tree code. I can unify the formatting across those two files if desired.

One problem I notice in the extent tree is  the `extent_tree_span` function, which is invoked by `unifyfs_inode_span_extents`, which is invoked by `unifyfs_fops_mdhim.c:get_local_extents`. I am not sure if we want to remove those functions. Especially, the output values returned by `extent_tree_span` are actually typed as `unifyfs_key_t` and `unifyfs_val_t`, which we have for interacting with mdhim:

```
/* given an extent tree and starting and ending logical offsets,
 * fill in key/value entries that overlap that range, returns at
 * most max entries starting from lowest starting offset,
 * sets outnum with actual number of entries returned */
int extent_tree_span(
    struct extent_tree* extent_tree, /* extent tree to search */
    int gfid,                        /* global file id we're looking in */
    unsigned long start,             /* starting logical offset */
    unsigned long end,               /* ending logical offset */
    int max,                         /* maximum number of key/vals to return */
    void* keys,             /* array of length max for output keys */
    void* vals,             /* array of length max for output values */
    int* outnum);                    /* number of entries returned */
```

It is a bit ugly because we use the void pointers to avoid including mdhim related headers (`server/src/unifyfs_metadata_mdhim.h`).

Including @MichaelBrim.

CamStan (approved):
A lot of work here. Amazing job @sandrain and @MichaelBrim!

Nothing jumped out that needs to be changed before merging.
Left a note for my future self for a couple updates after merging.

adammoody:
Thanks for all of the work on this, everyone!

client/src/unifyfs-sysio.c
```diff
@@ -1606,15 +1613,16 @@ int UNIFYFS_WRAP(lio_listio)(int mode, struct aiocb* const aiocb_list[],
         rc = unifyfs_gfid_read_reqs(reqs, reqcnt);
         if (rc != UNIFYFS_SUCCESS) {
             /* error reading data */
-            ret = -1;
+            ret = rc;
```
adammoody, line 1616: @sandrain , can ``unifyfs_gfid_read_reqs`` return something other than -1 on error?  Regardless, we want to be sure the ``lio_listio`` wrapper only returns -1.  To be clearer, I think it's best to leave this in as setting ``ret = -1``.

adammoody: Nevermind, now I see the other code below.

Issue #560, opened 2020-10-12
Problem in Unify/MPI-IO reading data across a "boundary" between two previously written ranges in a file


clmendes:
@adammoody , @kathrynmohror : After doing extensive testing with one of the failing parallel HDF5 codes from their testsuite (cchunk5), I think I finally understand why it is failing, although I haven't looked yet at how to fix it in Unify. I used Recorder (without Unify) to see exactly what that HDF5 program was doing for I/O, and then, correlating the traces to the logs from a Unify-based execution, I can see where it goes wrong with Unify (my Unify version is based on Mike's new margotree branch). Based on that, I created this reproducer program below, which only uses MPI-IO (i.e. no HDF5), as I think this will make it easier to look for a fix.

The example below must be run in 4 processors. Running without Unify, file "datafile-m" will be created, with 400 bytes. Running under Unify, file "datafile-u", which also has 400 bytes, will be created. A byte-by-byte comparison between those two files will show differences between bytes 101:107 across the two files. This is an error similar to what I saw with that HDF5 test.

In this code, each rank initially writes 100 bytes to disjoint locations in the file. Then rank-0 tries to read 12 bytes starting at offset=95. This would have to read across two ranges, one range [0:99] written previously by rank-0, and the range [100:199] previously written by rank-1. The read is done with function MPI_File_read_at.

After the MPI_File_read_at call, the number of bytes actually read is retrieved in status.count_lo (according to the MPI_Status structure of MPI). Without Unify, that number is 12. However, with Unify, that number is *5* (i.e. the amount of bytes between 95 and 99).  Those values (12 and 5) are printed by the executions of the program in stdout.

This is the test code (to be compiled for Unify, it must use -DUNIFY):

```
#include "mpi.h"
#include <stdio.h>
#include <stdlib.h>
#ifdef UNIFY
#include <unifyfs.h>
char *filename="ufs:/unifyfs/datafile-u" ;
#else
char *filename="datafile-m";
#endif

#define SIZE 100

int main(int argc, char **argv)
{
    int i, myrank, nranks, ret, count, nread;
    char buf[SIZE];
    MPI_File fh;
    MPI_Status status;
    MPI_Info info=MPI_INFO_NULL;
    MPI_Offset offset;

    MPI_Init(&argc, &argv);
    MPI_Comm_size(MPI_COMM_WORLD, &nranks);
    MPI_Comm_rank(MPI_COMM_WORLD, &myrank);
#ifdef UNIFY
[...]

PR #569, opened 2020-11-16
add option to sync after every write

adammoody:
This adds an option to sync after every (successful) write operation.  This will slow down performance, but it could be a work around for codes that are missing the necessary sync calls.

This includes a runtime option to enable the auto-sync-after-write ``UNIFYFS_CLIENT_WRITE_SYNC=1``.

It should help with https://github.com/LLNL/UnifyFS/issues/560

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -79,6 +79,11 @@ int unifyfs_client_id;
 static int unifyfs_use_single_shm = 0;
 static int unifyfs_page_size      = 0;
 
+/* Determine whether we automatically sync every write to server.
+ * This slows write performance, but it can serve as a work
+ * around for apps that do not have all necessary syncs. */
+static bool unifyfs_write_sync;
+
 static off_t unifyfs_max_offt;
 static off_t unifyfs_min_offt;
 static off_t unifyfs_max_long;
@@ -1706,6 +1711,15 @@ int unifyfs_fid_write(
             /* write succeeded, remember that we have new data
              * that needs to be synced with the server */
             meta->needs_sync = 1;
+
+            /* optionally sync after every write */
+            if (unifyfs_write_sync) {
+                int ret = unifyfs_sync(fid);
+                if (ret) {
+                    LOGERR("client sync after write failed");
+                    rc = ret;
+                }
+            }
         }
     } else {
         /* unknown storage type */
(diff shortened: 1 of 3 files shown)


adammoody:
@clmendes, I haven't tested this at all other than testing that it compiles, but I think the MPI I/O test code should work with this, and hopefully the HDF example.  If it helps, I could rebase this on the margotree file hash branch.

clmendes:
@adammoody , thanks for providing this. I can test it with the MPI-IO reproducer that I had created. However, for the "real" tests (i.e. tests from the HDF5 testsuite), I'd need to use it with at least the "margotree" branch (or after PR-566 is merged), since some of the other HDF5 tests only worked after I started to use the margotree branch from Eric.

adammoody:
@clmendes , oh it looks like it doesn't quite apply cleanly to the margotree branch proper (too old I guess).

I think I could apply it to the branch that is hosting PR #566 (new-margotree-abtsync-filehash).  Have you moved on to testing with branch yet?  If so, I'll target that.

clmendes:
@adammoody , no, I have not yet tested the PR-566's branch. Today I had a good part of the day around SC20 (we had our Blue Waters talk presented today, among other things), so not very productive.

But since this sync change could be applied to the mpi-io test separately, I was planning to do it using your 'writesync' branch, which I obtained yesterday. This would be quicker to test than the rest of the HDF5 tests (for those I'm indeed counting on that branch in PR-566).

clmendes:
@adammoody , I finally tested this new branch with my reproducer code from PR-560 (the mpi-io example), sorry for my delay. I have good and bad news from today's test.

The good news is that by setting UNIFYFS_CLIENT_WRITE_SYNC=1 in my job script the program indeed reads 12 bytes as expected. I even changed the example to print those 12 bytes, and verified that they had the correct value.

Now, the bad news is that even with this, the resulting file (datafile-u) is different from the file that I obtain without Unify (datafile-m), and the difference lies in positions [101-107]. In other words, that last write done by rank-0, of 12 bytes starting at offset=95, does not make completely to the resulting file: only the first 5 bytes are actually present in the resulting file. Bytes [101-107] still have the original value, initially written by rank-1. To fix this, I need to use the 'margotree' branch! (as a reminder, using the margotree branch, *and* adding the MPI_File_sync before the first barrier, everything works fine, i.e. 12 bytes are read and the resulting file is exactly the same as the file produced by the non-Unify version).

So, in summary, I believe your change will be valid, but to use it I'd really need to combine it with the 'margotree' branch, or with that newer branch from PR-566! As it stands right now, it fixes this write/sync issue, but there are other Unify problems that only got fixed after the margotree branch was created!

adammoody:
This writesync branch has now been rebased on dev since merging the margotree code.

MichaelBrim (approved):
Looks good to me.

adammoody:
Thanks, @MichaelBrim

adammoody:
@clmendes , this has now been merged into dev, so a fresh pull of dev should enable that option for you now.

Issue #570, opened 2020-11-16
unifyfs_sync questions


adammoody:
While testing the new sync-after-every-write option, I was tracing through the unifyfs_sync code.  A couple of questions:

I think it's possible for a user to have a file open with multiple descriptors, so there could be multiple fd slots that all share the same fid value.  Since ``unifyfs_sync(fid)`` is passed a file id, rather than a file descriptor, I think we should have it look for and flush all file descriptors that match on fid, instead of just the first entry.

https://github.com/LLNL/UnifyFS/blob/968c50dd6d377771994a2ca04a1516af62822dc7/client/src/unifyfs-fixed.c#L215-L218

Is that true?

Also, if data needs to be synced to the spillover file, we should do that *before* we call the server sync rpc.  Otherwise, I think we have a race condition where the server might try to read from the spillover file before the sync to the spill has completed.  I think we should move this code before we call the server rpc.

https://github.com/LLNL/UnifyFS/blob/968c50dd6d377771994a2ca04a1516af62822dc7/client/src/unifyfs-fixed.c#L221-L229

adammoody:
@MichaelBrim , I could look to submit a fix for these if needed, but wanted to make sure if this is right before I bother.

adammoody:
Oh, maybe it's ok to break the loop early after all, since I think we will just have one meta data structure for a given fid.  If that's right, I'll just add a comment about that so that I don't get confused again.

MichaelBrim:
@adammoody Your changes in PR #571 look ok to me.

PR #571, opened 2020-11-17
flush spillover file before invoking sync rpc on server

adammoody:
Updates ``unifyfs_sync(fid)`` to flush spillover file if needed before calling the server sync rpc.  Without doing this, the server might try to read from the spillover file before the data has been flushed out.  Though if the spillover file supports strict read-after-write POSIX data consistency, flushing the spillover file might not be required, in which case it could be skipped.

TODO: for ``unifyfs_sync(-1)``, this could be optimized to only flush the spillover file once, instead of potentially for each file.

### DIFF
--- a/client/src/unifyfs-fixed.c
+++ b/client/src/unifyfs-fixed.c
@@ -191,10 +191,68 @@ int truncate_write_meta(unifyfs_filemeta_t* meta, off_t trunc_sz)
  */
 int unifyfs_sync(int target_fid)
 {
+    int tmp_rc;
     int ret = UNIFYFS_SUCCESS;
-    off_t max_log_offset = 0;
 
-    /* For each open file descriptor .. */
+    /* if caller gave us a file id, sync that specific fid */
+    if (target_fid >= 0) {
+        /* user named a specific file id, lookup its metadata */
+        int fid = target_fid;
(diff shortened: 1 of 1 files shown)


adammoody:
For https://github.com/LLNL/UnifyFS/issues/570

MichaelBrim (approved):

client/src/unifyfs-fixed.c
```diff
@@ -191,76 +191,81 @@ int truncate_write_meta(unifyfs_filemeta_t* meta, off_t trunc_sz)
...
+                    LOGERR("failed to sync logio data");
+                    ret = tmp_rc;
                 }
+                LOGDBG("after logio spill sync");
             }
 
             /* tell the server to grab our new extents */
-            ret = invoke_client_sync_rpc(meta->gfid);
-            if (ret != UNIFYFS_SUCCESS) {
+            tmp_rc = invoke_client_sync_rpc();
```
MichaelBrim, line 235: Looks like Travis build is failing here due to the missing gfid parameter.

adammoody: Ah, yes.  Thank you, @MichaelBrim !  I botched a merge conflict here somehow and was too quick to push the PR back up.

Issue #577, opened 2020-11-23
Example to create and access a PHDF5 dataset under Unify


clmendes:
I'm creating this example to document a simple tests that creates and accesses a simple dataset in Parallel-HDF5, under Unify.

The original example is taken from the HDF5 website: https://support.hdfgroup.org/ftp/HDF5/examples/parallel/Dataset.c

Based on that example, the changes made for Unify are these:
a) Change the filename from SDS.h5 to ufs:/unifyfs/SDS.h5
b) Call to a function to mount /unifyfs after MPI_Init;
c) Call to unifyfs_umount() at the end.

On Quartz, after "load module hdf5-parallel", I build this example with 

h5pcc -o prog-gotcha mpi_prog.c -I${UNIFYFS}/include -L${UNIFYFS}/lib -lunifyfs_gotcha ${UNIFYFS}/lib64/libgotcha.so

This example, run with 8 processors on 2 nodes, used to work fine with the "new-margotree" branch of Unify. It is no longer working with the recent dev version  (as of Nov.20/2020). It only works in one node.

This is the modified source code:

```
/*
 *  This example writes data to the HDF5 file.
 *  Number of processes is assumed to be 1 or multiples of 2 (up to 8)
 */

#include "hdf5.h"
#include "stdlib.h"

#define H5FILE_NAME     "ufs:/unifyfs/SDS.h5"
#define DATASETNAME     "IntArray"
#define NX     8                      /* dataset dimensions */
#define NY     5
#define RANK   2

int
main (int argc, char **argv)
{
    /*
     * HDF5 APIs definitions
     */
    hid_t       file_id, dset_id;         /* file and dataset identifiers */
    hid_t       filespace;      /* file and memory dataspace identifiers */
    hsize_t     dimsf[] = {NX, NY};                 /* dataset dimensions */
    int         *data;                    /* pointer to data buffer to write */
    hid_t       plist_id;                 /* property list identifier */
    int         i;
    herr_t      status;

    /*
     * MPI variables
     */
    int mpi_size, mpi_rank;
    MPI_Comm comm  = MPI_COMM_WORLD;
    MPI_Info info  = MPI_INFO_NULL;

    /*
     * Initialize MPI
     */
    MPI_Init(&argc, &argv); my_unify_mount();
    MPI_Comm_size(comm, &mpi_size);
    MPI_Comm_rank(comm, &mpi_rank);

    /*
     * Initialize data buffer
     */
[...]

PR #578, opened 2020-11-25
fix: invoke correct rpc for truncate bcast tree

adammoody:
@sandrain , would you please double check me on this change?

### DIFF
--- a/server/src/unifyfs_group_rpc.c
+++ b/server/src/unifyfs_group_rpc.c
@@ -706,6 +706,7 @@ int truncate_bcast_forward(const unifyfs_tree_t* broadcast_tree,
 
         /* forward request down the tree */
         coll_request* req;
+        hg_id_t hgid = unifyfsd_rpc_context->rpcs.truncate_bcast_id;
         for (i = 0; i < child_count; i++) {
             req = requests + i;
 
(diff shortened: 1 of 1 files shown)


sandrain:
@adammoody @MichaelBrim Yes, looks like that was a bug. Thanks.

server/src/unifyfs_group_rpc.c
```diff
@@ -715,7 +715,7 @@ int truncate_bcast_forward(const unifyfs_tree_t* broadcast_tree,
                    i, child, glb_servers[child].margo_svr_addr_str);
 
             /* allocate handle */
-            rc = get_request_handle(unifyfsd_rpc_context->rpcs.truncate_id,
+            rc = get_request_handle(unifyfsd_rpc_context->rpcs.truncate_bcast_id,
```
sandrain, line 718: Isn't this line going over 80 (checkpatch may fail)?

MichaelBrim:
Yeah, I think this is something I already caught in the readrpc branch, but forgot to backport.

MichaelBrim (approved):

clmendes:
Yes, with this fix I was able to run correctly the HDF5 example (from Issue #577) on 2 processors and 2 Catalyst nodes (i.e. 1 processor/node).

However, when I try to run it with 4 processors in 2 nodes (i.e. 2 processors/node) I still get the same error as before. Thus, I'm afraid this might not yet be the last fix.

adammoody:
Thanks @sandrain and @MichaelBrim

Issue #580, opened 2020-11-29
Error in HDF5-testsuite´s cchunk5 with latest tree-based Unify


clmendes:
@adammoody , @kathrynmohror : I am creating this new issue to report/document the problem that I'm facing in one of the HDF5 tests (cchunk5), after switching to use the new Unify version based on PR#578, which implemented the new tree-based functionality.

I reproduced the problem with an isolated code that only calls cchunk5, so this is much simpler to handle. I get the same error in this code as in the original HDF5 testsuite. I can share this reproducer if there is interest.

First, a few comments. This problem does NOT occur when I use Unify with the previous "new-margotree" branch. Second, the test works fine when I run it on 1 node (it must be run with 4 processors). When I run it on 2 nodes (i.e. 2 processors/node), even using the new UNIFYFS_CLIENT_WRITE_SYNC=1 setting, and adding a " sleep 25" after starting the server, I get the following error in the application, for the first two ranks:

```
Proc 1: *** Parallel ERROR ***
    VRFY (dataset created succeeded) failed at line  166 in t_chunk.c
Proc 0: *** Parallel ERROR ***
    VRFY (dataset created succeeded) failed at line  166 in t_chunk.c
``` 
This is the section of the source code t_chunk.c where this error happens:

```
164:  dataset = H5Dcreate2(file, DSET_COLLECTIVE_CHUNK_NAME, H5T_NATIVE_INT,
165:                       file_dataspace, H5P_DEFAULT, crp_plist, H5P_DEFAULT);
166:  VRFY((dataset >= 0),"dataset created succeeded");
```

According to my non-Unify execution of this program with Recorder, this H5Dcreate2 function executes the following operations in the four ranks:
```
Rank=0:
0.067066 0.067068 0 lseek ( 15 3688 0 )
0.067073 0.067717 0 read ( 15 0x2055ff0 384 )
0.067736 0.067739 0 lseek ( 15 3688 0 )
0.068010 0.068016 0 write ( 15 0x2055ff0 384 )

Rank=1:
0.067129 0.067132 0 lseek ( 13 4072 0 )
0.067137 0.067831 0 read ( 13 0x2aaacdf670a0 384 )
0.067843 0.067847 0 lseek ( 13 4072 0 )
0.068711 0.068724 0 write ( 13 0x2aaacdf670a0 384 )

Rank=2:
0.088103 0.088105 0 lseek ( 15 4456 0 )
[...]

PR #581, opened 2020-11-30
fix: handle read extent with no registered data chunks

adammoody:
Resolves https://github.com/LLNL/UnifyFS/issues/580

### Description

This handles the case where a remote server has no extent information for a read request.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_p2p_rpc.c
+++ b/server/src/unifyfs_p2p_rpc.c
@@ -257,6 +257,7 @@ static void find_extents_rpc(hg_handle_t handle)
 
     const struct hg_info* hgi = margo_get_info(handle);
     assert(hgi);
+
     margo_instance_id mid = margo_hg_info_get_instance(hgi);
     assert(mid != MARGO_INSTANCE_NULL);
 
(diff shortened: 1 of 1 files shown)


clmendes:
@adammoody , I'm glad to report that this fix indeed fixes the error with the cchunk5 example that I had reported in issue #580.

To be clear: with this fix, and without using the UNIFYFS_CLIENT_WRITE_SYNC=1 setting, the example code presents the same behavior as with the new-margotree branch: the code goes to the end of the execution without error, but the data stored in the file has a few errors (due to the lac of the sync).  However, now, when I use UNIFYFS_CLIENT_WRITE_SYNC=1, the file ends up with the correct data!

Thus, by using UNIFYFS_CLIENT_WRITE_SYNC=1 we no longer need the H5Fflush call in the app, as designed! I verified that the example works correctly and produces correct data in the resulting file. I also tested the original HDF5 testsuite with cchunk5, and it now works fine, producing correct data as well when we have UNIFYFS_CLIENT_WRITE_SYNC=1 in the job and the original application code does NOT contain any H5Fflush!

adammoody:
Good to hear!  Thanks for testing this so quickly, @clmendes

adammoody:
Before merging, I should also check the corresponding code on the send side:
https://github.com/LLNL/UnifyFS/blob/c4c2faa5d5e954d59fe1279c1f3f32c319b80c6c/server/src/unifyfs_p2p_rpc.c#L318-L335

Seems like this should be symmetric.

MichaelBrim (approved):

MichaelBrim:
@adammoody In the case of no extents found, I believe `ret != UNIFY_SUCCESS`, and therefore we won't try to do the margo_bulk_create on the send side. Thus, I think your fix is complete.

adammoody:
I got a chance to trace the send path today.  Actually, it is also allocating a bulk transfer handle ``ret == UNIFYFS_SUCCESS`` even if no chunks are returned in this case.  I'll make the corresponding change on the send side.

I think it's interesting that margo lets one create a handle successfully with ``buf_sz = 0`` from the send side ``HG_BULK_READ_ONLY`` but not on the receive side ``HG_BULK_WRITE_ONLY``.  I don't know if it's unhappy because ``buf_sz = 0`` or perhaps it has something to do with the value of the buffer address we're passing in -- maybe it can't be pinned for write or something?

I think we'd want to code around this either way, since I never really trust the outcome of a ``malloc(0)``.  However, we could raise this question with the margo team.  I'm always a fan of symmetric interfaces myself, so that they would either both fail or both succeed.

adammoody:
@MichaelBrim , thanks for looking this over.  I force pushed after making an update to the send side.  I changed the code a bit in how we fill the output param struct and free the handle.  Let me know if that looks problematic.

MichaelBrim:
@adammoody updates seem good to me.

server/src/unifyfs_p2p_rpc.c
```diff
@@ -315,35 +316,37 @@ static void find_extents_rpc(hg_handle_t handle)
...
         LOGERR("margo_respond() failed");
     }
-    if (out.num_locations) {
-        margo_bulk_free(bulk_resp_handle);
-    }
 
     /* free margo resources */
+    if (bulk_resp_handle != HG_BULK_NULL) {
+        margo_bulk_free(bulk_resp_handle);
+    }
```
adammoody, line 349: I'm not sure if it is possible for the above ``margo_bulk_create`` to fail and set ``hret`` to something other than ``HG_SUCCESS`` but still fill in the ``bulk_resp_handle`` handle value with something other than ``HG_BULK_NULL``.  And if that is possible, do we need to call ``margo_bulk_free`` or not?  I don't know.

MichaelBrim: I would need to check the margo code, but I assume they do the right thing and don't assign to the output handle parameter when the operation fails. The only safe thing we can do is to only call `margo_bulk_free()` when `margo_bulk_create()` was successful. That means you should use the return value from the call to the create as the if condition.

MichaelBrim: I checked the margo code, it just passes straight through to mercury hg_bulk_create(), which only sets the output parameter on success.

MichaelBrim: https://github.com/mercury-hpc/mercury/blob/v1.0.1/src/mercury_bulk.c

adammoody: Cool.  Thanks, @MichaelBrim

adammoody:
@CamStan , @kathrynmohror just merged this PR

adammoody:
@clmendes , this fix is merged into dev now.

PR #592, opened 2020-12-16
Enable to control the mercury NA at runtime.

sandrain:
- prefer ofi to bmi when both are available.
- new configuration/env option (UNIFYFS_MARGO_BMI)
- documentation update

TEST_CHECKPATCH_SKIP_FILES="common/src/unifyfs_configurator.h"

### Types of changes


MichaelBrim:
@sandrain I think this is heading in the wrong direction. BMI will most likely be deprecated in Mercury, so we should not be introducing a user choice (i.e., UNIFYFS_MARGO_BMI) to select it. We can keep the code that prefers OFI over BMI when both are available.

sandrain:
@MichaelBrim In fact, by default the code will choose OFI over BMI. The code will select BMI only when user specifies UNIFYFS_MARGO_BMI or OFI is not available.

MichaelBrim (approved):
Looks good now.

adammoody:
Thanks @sandrain and @MichaelBrim

PR #593, opened 2020-12-17
fix for laminate bcast when inode does not exist

MichaelBrim:
### Description

When we laminate a file, the file owner broadcasts its extents and metadata to all other servers. However, there is no guarantee that the inode for the file existed on the other servers before the laminate, as there may not have been client writes to the file on those servers. As a result, the updates to extents and metadata previously failed with ENOENT. This PR fixes that case.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_group_rpc.c
+++ b/server/src/unifyfs_group_rpc.c
@@ -464,18 +464,32 @@ static void laminate_bcast_rpc(hg_handle_t handle)
                     }
                 }
 
+                /* update inode file attributes. first check to make sure inode
+                 * for the gfid exists. if it doesn't, create it with given
+                 * attrs. otherwise, just do a metadata update. */
+                unifyfs_file_attr_t existing_fattr;
+                ret = unifyfs_inode_metaget(gfid, &existing_fattr);
+                if (ret == ENOENT) {
+                    /* create with is_laminated=0 so extents can be added */
+                    fattr->is_laminated = 0;
+                    ret = unifyfs_inode_create(gfid, fattr);
+                    if (ret != UNIFYFS_SUCCESS) {
+                        LOGERR("inode create failed (ret=%d)", ret);
+                    }
+                    fattr->is_laminated = 1;
+                }
+
                 /* add the final set of extents */
                 ret = unifyfs_inode_add_extents(gfid, num_extents, extents);
                 if (ret != UNIFYFS_SUCCESS) {
                     LOGERR("laminate extents update failed (ret=%d)", ret);
-                }
-
-                /* update attributes only after final extents added */
-                ret = unifyfs_inode_metaset(gfid,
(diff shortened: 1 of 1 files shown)


MichaelBrim:
This breaks lamination, because you can't add extents after the setting `file_attrs.is_laminated`. I'll update and resubmit.

CamStan:
First attempt was still failing. Grabbing update now.

MichaelBrim:
pushed another update to make checkpatch happy

CamStan:
But no functionality change it the checkpatch push?

This appears to be working now with my initial test (running twice, two nodes with 1ppn). Will run a few more.
One thing that jumped out in the server logs now is this, just before the laminate happens for the second file:
```
# node with write server
2020-12-17T12:22:13 tid=135925 @ unifyfs_invoke_broadcast_truncate() [server/src/unifyfs_group_rpc.c:839] truncate_bcast_forward failed: (ret=2)

# node with read server
2020-12-17T12:22:13 tid=39990 @ truncate_bcast_forward() [server/src/unifyfs_group_rpc.c:701] unifyfs_inode_truncate(gfid=1346058476, size=0) failed - ret=2
```

I'll let you know result my other runs.

MichaelBrim:
That truncate_bcast_forward() error (ENOENT) is weird. Did you enable pre/post truncate in the write client? Still, it could be another place where we need to create the inode before applying the metadata update.

CamStan:
With this fix, data check is now passing on all the producer/consumer cases in which it was previously failing for me.

---
>Did you enable pre/post truncate in the write client?

I double checked and I don't have pre/post truncate enabled in the write client.

When I checked client log, the only truncate I'm seeing is during file creation. The timestamp for the failure matched with the server logs above. 
```
2020-12-17T12:22:13 tid=135920 @ invoke_client_truncate_rpc() [client/src/margo_client.c:521] invoking the truncate rpc function in client
2020-12-17T12:22:13 tid=135920 @ invoke_client_truncate_rpc() [client/src/margo_client.c:534] Got response ret=2
2020-12-17T12:22:13 tid=135920 @ __wrap_open() [client/src/unifyfs-sysio.c:1126] using fds (internal=0, external=16384) for fid 0 file /unifyfs/write-gotcha-2.app
DEBUG: finished create
```

However, "truncate" doesn't appear in the client logs when running my same tests with the current `dev` branch.

I can send the logs if wanted. However, I switched to running two examples back to back in a non-producer/consumer setting (both hosts write and read) and this failure happens on the second run there as well. So might be able to reproduce on your end and it's possibly a different issue this fix uncovered which could be addressed separately.

MichaelBrim:
Leaving this PR as is for now since it is fixing Cameron's test cases. I'll look for the truncate issue and file a separate PR if I find a fix.

adammoody (approved):
Thanks, @MichaelBrim !

PR #598, opened 2021-01-07
Use RPC for read request replies

MichaelBrim:
### Description

This PR implements read replies and completion of read requests via Margo RPCs from server to client, and removes our custom shmem region and signaling previously used for read replies. It also includes a few fixes for bugs encountered while testing the readrpc branch.

Additionally, two new features are introduced.
1) `UNIFYFS_LOG_ON_ERROR` - a new configuration variable that will enable verbose logging upon the first call to `LOGERR()`
2) `UNIFYFS_SERVER_CORES` - a new environment variable used when launching servers with the `unifyfs` utility that specifies the number of processor cores to reserve for each server process (default is 1, same as prior behavior)

### Motivation and Context

Previously, we created a per-client shmem region for returning read data from the server to the client. We also relied on polling of a shmem location for signaling data availability and consumption. Margo RPCs within a node also use shmem for large data buffers, and provide a much simpler abstraction for notifications. Switching to a Margo RPC also opens up the future potential for supporting direct client replies from a remote server.

### How Has This Been Tested?

Tested on Summit using the examples with up to 128 nodes (8 ppn).

### Types of changes

### Checklist:


MichaelBrim:
This is aggravating. I'm fairly sure none of my changes are the root cause for this Travis failure when building the static (i.e., ld wrap) tests. I didn't even touch `t/Makefile.am` until I started to workaround the failure. The short summary of the problem is that the linker is attempting to wrap syscalls in object files from libunifyfs_common.a, but the definitions for the `_wrap_<syscall>` functions are in libunifyfs.a, which precedes libunifyfs_common on the link line. I have yet to find a workaround, and have no clue why this suddenly started happening. I would appreciate if someone else could look into this as well to make sure I'm not missing something obvious.

adammoody:
I also don't know why this suddenly started happening.  We've got a circular dependency between libunifyfs.a and libunifyfs_common.a because of the wrapped I/O calls.

- libunifyfs.a needs unifyfs_config_init() from libunifyfs_common.a
- libunifyfs_common.a needs __wrap_fopen() from libunifyfs.a

I'm trying to hack in the ld flags for ``--start-group`` and ``--end-group`` as one fix, though that's not straight-forward either.

As another option, and perhaps a cleaner one, we could skip building a libunifyfs_common at all and just list all of its source files as part of the client and server builds.

adammoody:
Yep, this is frustrating, still banging my head on the wall with it, too.  I have tried to repeat the libraries in our LDADD list to handle the circular reference.  However, libtool seems to be stripping those out, probably realizing there are duplicate entries.

There is a ``--preserve-dup-deps`` option one can pass to libtool to prevent that, but that is not yet working for me either.  Perhaps due to the libtool version on travis?

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=508658

```
$ libtool --version
libtool (GNU libtool) 2.4.6
```

Still hunting for a work around.

CamStan:
@adammoody, looks like your latest build worked. Made it to the tests.

adammoody:
Thanks @CamStan .  I think I ended up getting something to work by just listing all common source files as part of the input source files to libunifyfs and then dropping its dependency on libunifyfs_common.  That means we compile some files twice, but it eliminates the circular dependency problem.  Submitted a PR to @MichaelBrim 's branch.

MichaelBrim:
I merged Adam's work against my readrpc branch, did a little cleanup, and rebased/pushed. Let's see if we make it further.

MichaelBrim:
Seems to be building the tests ok, I will look into the new test failures.

client/src/Makefile.am
```diff
@@ -36,11 +49,56 @@ CLIENT_COMMON_LDFLAGS = \
   $(MARGO_LDFLAGS)
 
 CLIENT_COMMON_LIBADD = \
-  $(top_builddir)/common/src/libunifyfs_common.la \
+  $(OPT_LIBS) \
   $(MARGO_LIBS) \
-  -lrt -lpthread
+  -lm -lrt -lcrypto -lpthread
+
+LIBCOMMON_SRCS = \
```
sandrain, line 56: I think this breaks the `__FILE__` macro in log printing in the client. The pathnames are not correctly cut:

```
2021-01-14T14:02:42 tid=118463 @ invoke_client_mount_rpc() [client.c:270] invoking the mount rpc function in client
2021-01-14T14:02:42 tid=118463 @ invoke_client_mount_rpc() [client.c:287] Got response ret=0
2021-01-14T14:02:42 tid=118463 @ invoke_client_mount_rpc() [client.c:297] My client id is 33
2021-01-14T14:02:42 tid=118463 @ get_page_size() [common/src/unifyfs_logio.c:63] returning page size 65536 B
2021-01-14T14:02:42 tid=118463 @ unifyfs_init() [s.c:1754] FD limit for system = 4096
2021-01-14T14:02:42 tid=118463 @ init_superblock_structures() [s.c:1542] Meta-stacks initialized
```

sandrain: Looks like the `common` dir is expanded as `/autofs/nccs-svm1_techint/home/hs2/projects/UnifyFS/client/src/../../common/src/unifyfs_log.c`. Because of `/client/src../..`, `unifyfs_log_print` cuts extra characters from the pathname.

MichaelBrim: Yeah, I noticed that effect too. I will fix it.

MichaelBrim:
Still trying to reproduce the remaining 18 test errors (all read related) outside of Travis. The errors are all coming from the sysio suite, and are consistent across static/gotcha versions. When I run those test suites on Summit, they work fine, so I'm still trying to figure out what's going wrong.

MichaelBrim:
Not sure this will fix the Travis tests, but I found out that the settings I updated in the example unifyfs.conf file were actually being used. I pushed a fix to avoid that.

MichaelBrim:
I now have a Docker container where I can reproduce the same failures as Travis. Hopefully I'll be able to get some better insight into the failures.

MichaelBrim:
In my container, the read failures are due to failed margo_bulk_transfer()s from server to client. The bulk buffer size is 1 MiB for all the failed transfers, but other tests using reads of the same size complete successfully.

sandrain:
@MichaelBrim  When I was fixing the bulk problem between servers, I could not find a reliable way to figure out the max bulk buffer size associated with the transport (#598). But I am not sure if this is the cause because you also observe successful cases with the same size.

MichaelBrim:
For these transfers, the data should be going through shmem NA, so I'm doubtful about it being a buffer size issue. I also discovered a few of the failed transfers are only 512KiB - not exclusive to 1 MiB.

MichaelBrim:
To debug the bulk transfers, I added a test that cycles through various read sizes, starting at 1KiB and going up to 64MiB by doubling. Only transfers of less than 4KiB complete successfully in my Docker container. I'm going to push the new test to see if Travis has the same size limitation.

MichaelBrim:
Looks like Travis has the same limitation, only reads less than 4KiB passed. Now we just need to figure out what is different in these VM/container hosts versus the Summit nodes where things work.

adammoody:
Nice debugging work, guys!  Is this using shared memory?  Do things work in bulk mode with TCP?

MichaelBrim:
For posterity, I will summarize the problem. We use the Mercury sm (shared memory) NA for client-server RPCs. The implementation of the sm NA requires Linux cross-memory attach (CMA) support (available in v3.2+). However, there is also a Linux Security Module named YAMA that controls the scope of ptrace attach (and thus CMA). The sysctl name of this setting is `kernel.yama.ptrace_scope`. On Ubuntu systems, like our Travis CI VM, the default is to restrict this to only parent-child processes (value=1). The workaround is to restore the original ptrace attach behavior (value=0), which permits any two processes running with the same uid to use ptrace attach.

MichaelBrim:
Looks like the tests are passing again in Travis, but make distcheck is failing at the distclean stage. Seems related to the new method of directly including common sources in the client library.

adammoody:
> 
> 
> For posterity, I will summarize the problem. We use the Mercury sm (shared memory) NA for client-server RPCs. The implementation of the sm NA requires Linux cross-memory attach (CMA) support (available in v3.2+). However, there is also a Linux Security Module named YAMA that controls the scope of ptrace attach (and thus CMA). The sysctl name of this setting is `kernel.yama.ptrace_scope`. On Ubuntu systems, like our Travis CI VM, the default is to restrict this to only parent-child processes (value=1). The workaround is to restore the original ptrace attach behavior (value=0), which permits any two processes running with the same uid to use ptrace attach.

Thanks, @MichaelBrim .  We should capture this somewhere in our user guide, as well, since some end user might run into this on their system.  Also, we should add an item on our TODO list to search for the recommended work around in case one cannot get CMA working.

That's cool that Mercury is using CMA.  That was a feature added for faster MPI communication between processes on the same node, and most of the MPI libraries use it now.

One thing we've run into with in the MPI libraries that we should check here.  Even if CMA is enabled for users in general, the OS can disable it in certain situations.  For example, if the read bit is not set on the executable, I think CMA is disabled.  We have run into this where some binaries have their execute bit set but not their read bit.  That enables someone to run a code but not read any of its bits.  I wonder if we might see the same thing if we disable the read bit on the UnifyFS client application.

MichaelBrim:
As Phil Carns noted in his email, mercury 2.x has an explicit check for this that will tell users if CMA was detected as supported by the configuration but not runtime enabled. Another reason we should try to move to newer margo/mercury/argobots versions.

MichaelBrim:
The fix for the make distclean problem (missing `.deps/*.Plo` files) seems to be to remove the 'subdir-objects' automake init option. There are tons of projects out there that hit this same bug, and it's considered a blocker to automake v2.0's release. Note that this does result in autogen spitting out a bunch of warnings during execution of automake, but they are harmless.

CamStan:
Added a commit to update the bootstrap script for building autotools if needed. Used a version of the script from #601.

With this, was able to get UnifyFS to build and run, as well as unit tests to pass, on the systems here. So should be able to merge when ready.

Still end up with a segfault that crashes the server when running the CI tests. Still digging into this, but it can be addressed in a separate PR if necessary.

adammoody:
> 
> 
> Added a commit to update the bootstrap script for building autotools if needed. Used a version of the script from #601.
> 
> With this, was able to get UnifyFS to build and run, as well as unit tests to pass, on the systems here. So should be able to merge when ready.
> 
> Still end up with a segfault that crashes the server when running the CI tests. Still digging into this, but it can be addressed in a separate PR if necessary.

Great work, @CamStan .  Nice touch on the conditional install, too.

adammoody (approved):
Thanks a ton, @MichaelBrim !  I'll try to review this more closely post-merge when I get a chance.

PR #601, opened 2021-01-25
add script to download and build autotools

adammoody:
@CamStan , this helps with builds of @MichaelBrim 's readrpc branch on LLNL systems.  I recycled this "buildme" script from another project.  It pulls versions of autotools that are dated, but newer than what we have on the system.

First, run the autotools buildme script:
```
cd autotools
./buildme
```

Then, modify your PATH (to point to your autotools install) and PKG_CONFIG_PATH (to point to system pkg-config directories) before building UnifyFS:
```
export PATH=`pwd`/autotools/install/bin:$PATH
export PKG_CONFIG_PATH=/usr/share/pkgconfig:/usr/lib64/pkgconfig:$PKG_CONFIG_PATH
./autogen.sh
...
```

It still prints warnings and feels hacky with PKG_CONFIG.  It needs work, but it lets me build.

### DIFF
--- a/autotools/buildme
+++ b/autotools/buildme
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+set -x
+
+#mkdir -p autotools
+#cd autotools
+
+installdir=`pwd`/install
+
+# add autotools install bin to our path
+export PATH=${installdir}/bin:$PATH
+
+# build autoconf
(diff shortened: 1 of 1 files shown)


CamStan:
Added a version of this script to the bootstrap script in #598 to build autotools if the version of Automake detected on the system is older than 1.15.

PR #603, opened 2021-01-27
avoid truncate on creating a new file

sandrain:
- avoid triggering truncate on creation of a new file in `open(2)`
- setting `EISDIR` for `truncate(2)` on a directory (when locally detected)

### Description

This patch fixes `unifyfs_fid_open` not to trigger the truncate operation after successfully creating a new file.

### Types of changes

### DIFF
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -1105,6 +1105,11 @@ int unifyfs_fid_truncate(int fid, off_t length)
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
     assert(meta != NULL);
 
+    /* truncate is not valid for directories */
+    if (S_ISDIR(meta->mode)) {
+        return EISDIR;
+    }
+
     if (meta->is_laminated) {
         /* Can't truncate a laminated file */
         return EINVAL;
@@ -1217,6 +1222,9 @@ int unifyfs_fid_open(
      * if O_APPEND, set pos to file size
      */
 
+    /* flag indicating whether file should be truncated */
+    int need_truncate = 0;
+
     /* determine whether we are creating a new file
      * or opening an existing one */
     if (flags & O_CREAT) {
@@ -1267,6 +1275,11 @@ int unifyfs_fid_open(
                 LOGERR("Failed to get metadata on existing file %s (fid:%d)",
                     path, fid);
             }
+
+            /* check for truncate if the file exists already */
+            if ((flags & O_TRUNC) && open_for_write && !gfattr.is_laminated) {
+                need_truncate = 1;
+            }
         }
         if (ret != UNIFYFS_SUCCESS) {
             LOGERR("Failed to populate the global meta entry for %s (fid:%d)",
@@ -1277,12 +1290,6 @@ int unifyfs_fid_open(
                 unifyfs_fid_delete(fid);
             }
             return ret;
-        }
-
-        /* successfully opened the file with O_CREAT, truncate if needed
-         * and file is not laminated */
-        if ((flags & O_TRUNC) && open_for_write && !gfattr.is_laminated) {
-            ret = unifyfs_fid_truncate(fid, 0);
         }
     } else {
         /* trying to open without creating, file must already exist,
@@ -1330,6 +1337,11 @@ int unifyfs_fid_open(
         /* Successful in fetching metadata for existing file.
          * Update our local cache using that metadata. */
         unifyfs_fid_update_file_meta(fid, &gfattr);
+
+        /* check if we need to truncate the existing file */
+        if ((flags & O_TRUNC) && open_for_write && !gfattr.is_laminated) {
+            need_truncate = 1;
+        }
     }
 
     /* if given O_DIRECTORY, the named file must be a directory */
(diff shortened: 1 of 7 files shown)


sandrain:
https://github.com/LLNL/UnifyFS/blob/9a2f9ced1dab24d3bd2c976d430143b4a638467f/server/src/unifyfs_p2p_rpc.c#L918

@MichaelBrim Do we still need the truncate broadcasting? Truncate is only possible before lamination when the file metadata exists only in a single server. If the file is laminated (metadata copied in all servers), truncate isn't allowed anymore (so no need for bcast)? Also, it looks like this (truncate bcast on laminated file) will fail anyway because inode rejects:

https://github.com/LLNL/UnifyFS/blob/9a2f9ced1dab24d3bd2c976d430143b4a638467f/server/src/unifyfs_inode.c#L222

MichaelBrim:
@sandrain truncate broadcast is still necessary (before laminate) to ensure a consistent file view. what currently happens is the truncate goes to the owner, and then the owner initiates the truncate broadcast. this allows the other servers to delete the extents that are no longer valid due to the truncate.

sandrain:
@MichaelBrim I understand now. Yes, we would need to discard data segments if truncated. Thanks for clarification.

MichaelBrim (changes requested):
Please fix the missing bracket problem

client/src/unifyfs.c
```diff
@@ -1960,12 +1965,13 @@ int unifyfs_fid_open(
                 unifyfs_fid_delete(fid);
             }
             return ret;
-        }
 
-        /* successfully opened the file with O_CREAT, truncate if needed
-         * and file is not laminated */
-        if ((flags & O_TRUNC) && open_for_write && !gfattr.is_laminated) {
-            ret = unifyfs_fid_truncate(fid, 0);
+            /* successfully opened the file with O_CREAT, truncate if needed
```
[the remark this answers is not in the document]
sandrain: Something went wrong here. I will fix.

sandrain:
We do not have examples for testing this specific changes (various open w/ O_TRUNC cases). I will try to add an example.

sandrain:
The following cases have been tested:

- opening/creating a new file: no truncate broadcasting is triggered
- opening an existing laminated file: O_EXCL fails (EEXISTS), O_TRUNC fails (EROFS)
- opening an existing non-laminated file: O_EXCL fails (EEXISTS), O_TRUNC truncates the file to 0

sandrain:
Somehow github doesn't allow me to remove the WIP label.

client/src/unifyfs.c
```diff
@@ -2013,6 +2018,11 @@ int unifyfs_fid_open(
         /* Successful in fetching metadata for existing file.
          * Update our local cache using that metadata. */
         unifyfs_fid_update_file_meta(fid, &gfattr);
+
+        /* check if we need to truncate the existing file */
+        if ((flags & O_TRUNC) && open_for_write && !gfattr.is_laminated) {
+            need_truncate = 1;
+        }
```
adammoody, line 2025: Nice catch!

client/src/unifyfs.c
```diff
@@ -1899,6 +1904,7 @@ int unifyfs_fid_open(
      *   check that local and global info are consistent
      * if O_APPEND, set pos to file size
      */
+    int need_truncate = 0;
```
adammoody, line 1908: Separating this line of code from the preceding comment, which is independent.
```suggestion
     */
     
     /* flag indicating whether file should be truncated */
    int need_truncate = 0;

```

client/src/unifyfs.c
```diff
@@ -2059,6 +2069,15 @@ int unifyfs_fid_open(
...
 
+    /* truncate the file, if we have to */
+    if (need_truncate) {
+        ret = unifyfs_fid_truncate(fid, 0);
+        if (ret != UNIFYFS_SUCCESS) {
+            LOGERR("Failed to truncate the file %s", path);
+            return ret;
+        }
+    }
+
```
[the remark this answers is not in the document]
sandrain: Yea, good point. I was not thinking that O_APPEND can be somehow specified together with O_TRUNC.  I will address this and your suggestion above as well. Thanks, @adammoody

MichaelBrim (approved):

client/src/unifyfs.c
```diff
@@ -1105,6 +1105,11 @@ int unifyfs_fid_truncate(int fid, off_t length)
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(fid);
     assert(meta != NULL);
 
+    /* truncate is not valid for directories */
+    if (S_ISDIR(meta->mode)) {
+        return EISDIR;
+    }
+
```
adammoody, line 1112: We don't want to fix this here, but this code sequence reminds me that we still have some work to do to make sure our local cache of metadata is consistent after unlink.  For example in this case, I think the following sequence will cause problems for us:
```
rank 0:
mkdir("/unifyfs/foo")

rank 1:
remove("/unifyfs/foo")
fd = open("/unifyfs/foo", O_CREAT)
write(fd)
close(fd)

// at this point, the directory foo has been replaced by a file so truncate should work,
// but the cache in rank 0 will still see it as a directory so it will fail
rank 0:
truncate("/unifyfs/foo")
```

adammoody (approved):

adammoody:
Thanks @sandrain

Issue #442, opened 2020-01-22
Extent test suite to test different configurations


adammoody:
As we start to have configuration parameters that enable/disable capabilities or new optimizations, we should be sure our test suite runs through all combinations.  Note that this is complicated in that some test cases may no longer work, or the expected behavior may change.

For example, we may have a configuration option that changes whether stat() returns 0, local size, or the current global size pre-lamination.  The test cases that check file size with stat will need to know this and know to check the correct expected value under each configuration.

adammoody:
@CamStan , @tonyhutter , got an issue started here we can use to discuss this.

MichaelBrim:
We need to test I/O shuffle (i.e., read different data than written by the local rank) with data integrity checking turned on. For the writeread example, the relevant options are `-x -k`.

CamStan:
Test the various ways of configuring UnifyFS to make sure they all work:

- envars
-config file
-command line options

PR #606, opened 2021-02-08
client: define unifyfs super magic

adammoody:
This defines a ``UNIFYFS_SUPER_MAGIC`` value in ``unifyfs.h``.  The ``statfs`` wrapper returns this value by default in the ``struct statfs f_type`` field.  For clients that query ``statfs`` but do not yet handle a response of ``UNIFYFS_SUPER_MAGIC``, one can configure UnifyFS to return ``TMPFS_MAGIC`` instead by setting ``UNIFYFS_CLIENT_SUPER_MAGIC=0``.

```
UNIFYFS_CLIENT_SUPER_MAGIC=1 -> UNIFYFS_SUPER_MAGIC (default)
UNIFYFS_CLIENT_SUPER_MAGIC=0 -> TMPFS_MAGIC
```

A couple of proposed options for ``UNIFYFS_SUPER_MAGIC``:
```
0x554E4659 spells "UNFY" in ASCII (32-bit)
0x556E696679465321 spells "UnifyFS!" in ASCII (64-bit)
```

Other file system magic values can be seen with ``man statfs``:

```
              ADFS_SUPER_MAGIC      0xadf5
              AFFS_SUPER_MAGIC      0xADFF
              BEFS_SUPER_MAGIC      0x42465331
              BFS_MAGIC             0x1BADFACE
              CIFS_MAGIC_NUMBER     0xFF534D42
              CODA_SUPER_MAGIC      0x73757245
              COH_SUPER_MAGIC       0x012FF7B7
              CRAMFS_MAGIC          0x28cd3d45
              DEVFS_SUPER_MAGIC     0x1373
              EFS_SUPER_MAGIC       0x00414A53
              EXT_SUPER_MAGIC       0x137D
              EXT2_OLD_SUPER_MAGIC  0xEF51
              EXT2_SUPER_MAGIC      0xEF53
              EXT3_SUPER_MAGIC      0xEF53
              EXT4_SUPER_MAGIC      0xEF53
              HFS_SUPER_MAGIC       0x4244
              HPFS_SUPER_MAGIC      0xF995E849
              HUGETLBFS_MAGIC       0x958458f6
              ISOFS_SUPER_MAGIC     0x9660
              JFFS2_SUPER_MAGIC     0x72b6
              JFS_SUPER_MAGIC       0x3153464a
              MINIX_SUPER_MAGIC     0x137F /* orig. minix */
              MINIX_SUPER_MAGIC2    0x138F /* 30 char minix */
              MINIX2_SUPER_MAGIC    0x2468 /* minix V2 */
              MINIX2_SUPER_MAGIC2   0x2478 /* minix V2, 30 char names */
              MSDOS_SUPER_MAGIC     0x4d44
              NCP_SUPER_MAGIC       0x564c
              NFS_SUPER_MAGIC       0x6969
              NTFS_SB_MAGIC         0x5346544e
              OPENPROM_SUPER_MAGIC  0x9fa1
              PROC_SUPER_MAGIC      0x9fa0
              QNX4_SUPER_MAGIC      0x002f
              REISERFS_SUPER_MAGIC  0x52654973
              ROMFS_MAGIC           0x7275
              SMB_SUPER_MAGIC       0x517B
              SYSV2_SUPER_MAGIC     0x012FF7B6
              SYSV4_SUPER_MAGIC     0x012FF7B5
              TMPFS_MAGIC           0x01021994
              UDF_SUPER_MAGIC       0x15013346
              UFS_MAGIC             0x00011954
              USBDEVICE_SUPER_MAGIC 0x9fa2
              VXFS_SUPER_MAGIC      0xa501FCF5
              XENIX_SUPER_MAGIC     0x012FF7B4
              XFS_SUPER_MAGIC       0x58465342
              _XIAFS_SUPER_MAGIC    0x012FD16D
```

### Types of changes

### Checklist:

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -93,7 +93,7 @@ if [ "$automake_sub_version" -lt "15" ]; then
 
     autotools_repos=(http://ftp.gnu.org/gnu/autoconf/autoconf-2.69.tar.gz
         http://ftp.gnu.org/gnu/automake/automake-1.15.tar.gz
-        http://mirror.team-cymru.org/gnu/libtool/libtool-2.4.2.tar.gz
+        http://mirror.team-cymru.org/gnu/libtool/libtool-2.4.6.tar.gz
         https://pkg-config.freedesktop.org/releases/pkg-config-0.27.1.tar.gz
     )
 
@@ -129,7 +129,7 @@ if [ "$automake_sub_version" -lt "15" ]; then
 
     # build libtool
     echo "### building libtool ###"
-    pushd libtool-2.4.2
+    pushd libtool-2.4.6
         ./configure --build=$sys_name --prefix=$INSTALL_DIR
         make
         make install
@@ -138,7 +138,7 @@ if [ "$automake_sub_version" -lt "15" ]; then
     # build pkg-config
     echo "### building pkg-config ###"
     pushd pkg-config-0.27.1
-        ./configure --build=$sys_name --prefix=$INSTALL_DIR
+        ./configure --build=$sys_name --prefix=$INSTALL_DIR --with-internal-glib
         make
         make install
     popd
@@ -146,7 +146,6 @@ else
     echo "### detected automake version is greater than or equal to 1.15 ###"
     echo "### skipping autotools build ###"
 fi
-
 
 echo "### building GOTCHA ###"
 cd GOTCHA
--- a/client/src/unifyfs.c
+++ b/client/src/unifyfs.c
@@ -88,6 +88,9 @@ static off_t unifyfs_min_long;
 int    unifyfs_max_files;  /* maximum number of files to store */
 bool   unifyfs_local_extents;  /* track data extents in client to read local */
 
+/* whether to return UNIFYFS (true) or TMPFS (false) magic value from statfs */
+bool unifyfs_super_magic;
+
 /* log-based I/O context */
 logio_context* logio_ctx;
 
@@ -1746,6 +1749,17 @@ static int unifyfs_init(void)
             rc = configurator_bool_val(cfgval, &b);
             if (rc == 0) {
                 unifyfs_write_sync = (bool)b;
+            }
+        }
+
+        /* Determine SUPER MAGIC value to return from statfs.
+         * Use UNIFYFS_SUPER_MAGIC if true, TMPFS_SUPER_MAGIC otherwise. */
+        unifyfs_super_magic = true;
+        cfgval = client_cfg.client_super_magic;
+        if (cfgval != NULL) {
+            rc = configurator_bool_val(cfgval, &b);
+            if (rc == 0) {
+                unifyfs_super_magic = (bool)b;
             }
         }
 
--- a/t/0110-statfs-gotcha.t
+++ b/t/0110-statfs-gotcha.t
@@ -1,0 +1,13 @@
+#!/bin/bash
+#
+# Source sharness environment scripts to pick up test environment
+# and UnifyFS runtime settings.
+#
+. $(dirname $0)/sharness.d/00-test-env.sh
+. $(dirname $0)/sharness.d/01-unifyfs-settings.sh
+
+# disable statfs from returning UnifyFS! super magic value,
+# return tmpfs magic instead
+export UNIFYFS_CLIENT_SUPER_MAGIC=0
+
+$JOB_RUN_COMMAND $UNIFYFS_BUILD_DIR/t/sys/statfs-gotcha.t
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -114,6 +114,7 @@ test_cppflags = \
 sys_sysio_gotcha_t_SOURCES = \
   sys/sysio_suite.h \
   sys/sysio_suite.c \
+  sys/statfs.c \
   sys/creat-close.c \
   sys/creat64.c \
   sys/mkdir-rmdir.c \
@@ -133,6 +134,7 @@ sys_sysio_gotcha_t_LDFLAGS = $(test_gotcha_ldflags)
 sys_sysio_static_t_SOURCES = \
   sys/sysio_suite.h \
   sys/sysio_suite.c \
+  sys/statfs.c \
   sys/creat-close.c \
   sys/creat64.c \
   sys/mkdir-rmdir.c \
--- a/t/Makefile.am
+++ b/t/Makefile.am
@@ -10,11 +10,13 @@ TESTS = \
 if HAVE_GOTCHA
   TESTS += \
     0100-sysio-gotcha.t \
+    0110-statfs-gotcha.t \
     0200-stdio-gotcha.t
 endif
 
 TESTS += \
   0500-sysio-static.t \
+  0510-statfs-static.t \
   0600-stdio-static.t \
   0700-unifyfs-stage-full.t \
   9005-unifyfs-unmount.t \
@@ -28,6 +30,7 @@ TESTS += \
 check_SCRIPTS = \
   0001-setup.t \
   0500-sysio-static.t \
+  0510-statfs-static.t \
   0600-stdio-static.t \
   0700-unifyfs-stage-full.t \
   9005-unifyfs-unmount.t \
@@ -41,6 +44,7 @@ check_SCRIPTS = \
 if HAVE_GOTCHA
   check_SCRIPTS += \
     0100-sysio-gotcha.t \
+    0110-statfs-gotcha.t \
     0200-stdio-gotcha.t
 endif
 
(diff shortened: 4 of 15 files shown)


MichaelBrim:
@adammoody are 64-bit values supported? looks like all the existing values are 32-bit

adammoody:
Yeah, good eye, @MichaelBrim .  That's not clear to me either .

When I wrote this up, I noted in the CORAL 2 ``man statfs`` that it defines the ``f_type`` field conditionally like this:
```
           #if __WORDSIZE == 32          /* System word size */
           # define __SWORD_TYPE           int
           #else /* __WORDSIZE == 64 */
           # define __SWORD_TYPE         long int
           #endif

           struct statfs {
               __SWORD_TYPE f_type;    /* type of file system (see below) */
```
It does seem to be a 64-bit value on lassen, which I could see under totalview.  Though I don't know if it's common to define that as 64 bits.  If there are still lots of machines out there using a 32-bit field, we should stick with the 32-bit value to be more portable.

t/sys/statfs.c
```diff
@@ -0,0 +1,73 @@
...
+     * errno=ENOENT */
+    struct statfs fs;
+    errno = 0;
+    int rc = statfs(path, &fs);
+    int err = errno;
+    ok(rc == 0,
+       "statfs %s (rc=%d, errno=%d): %s",
+       path, rc, err, strerror(err));
+
+    /* Verify we can create a new file. */
```
MichaelBrim, line 53: Should probably updated this comment to reflect what the test really does.

t/sys/statfs.c
```diff
@@ -0,0 +1,73 @@
...
+ * desired results. */
+int statfs_test(char* unifyfs_root, int expect_unifyfs_magic)
+{
+    /* Diagnostic message for reading and debugging output */
+    diag("Starting UNIFYFS_WRAP(statfs) tests");
+
+    /* call statfs on the mountpoint */
+    const char* path = unifyfs_root;
+
+    /* Verify opening a non-existent file without O_CREAT fails with
```
MichaelBrim, line 43: Delete this comment

t/sys/statfs.c
```diff
@@ -0,0 +1,73 @@
...
+#include <linux/limits.h>
+#include "t/lib/tap.h"
+#include "t/lib/testutil.h"
+
+/* for statfs */
+#include <sys/vfs.h>
+
+#include "unifyfs.h"
+
+/* This function contains the tests for UNIFYFS_WRAP(open) found in
```
MichaelBrim, line 29: Should update this comment to reflect what test actually does.

adammoody:
Thanks, @MichaelBrim .  That reminds me that I wrote the test to be able to check both values.  However, it's only checking the default.

To check the TMPFS value, we should run the test again after setting ``UNIFYFS_CLIENT_SUPER_MAGIC=0`` and starting a fresh client instance and a new call to ``unifyfs_mount``.  I meant to bring that up on the call today as a discussion point.  Adding @CamStan .

Do we have any tests that check behavior under different configurations like this yet?

CamStan:
> To check the TMPFS value, we should run the test again after setting `UNIFYFS_CLIENT_SUPER_MAGIC=0` and starting a fresh client instance and a new call to `unifyfs_mount`.
> 
> Do we have any tests that check behavior under different configurations like this yet?

We currently don't have any tests that do this, but at first glance, both frameworks (unit tests and CI tests) should already be set up to handle this. At least in regards to only needing a fresh client.

Currently the unit tests use the default configs if they're not changed in [t/sharness.d/00-test-env.sh](https://github.com/LLNL/UnifyFS/blob/dev/t/sharness.d/00-test-env.sh) or [t/sharness.d/01-unifyfs-settings.sh](https://github.com/LLNL/UnifyFS/blob/dev/t/sharness.d/01-unifyfs-settings.sh).

---

In the unit tests, each suite is currently set up as a fresh client. In the case of this PR, the suite is [t/sys/sysio_suite.c](https://github.com/LLNL/UnifyFS/blob/dev/t/sys/sysio_suite.c):
https://github.com/LLNL/UnifyFS/blob/9620621aedb9d24338e86b20230646b6a510d4d7/t/sys/sysio_suite.c#L56

This is added to the test framework in part by creating a `.t` file (e.g., [t/0100-sysio-gotcha.t](https://github.com/LLNL/UnifyFS/blob/dev/t/0100-sysio-gotcha.t)):
https://github.com/LLNL/UnifyFS/blob/9620621aedb9d24338e86b20230646b6a510d4d7/t/0100-sysio-gotcha.t#L1-L8

So I think all we'd need to do is create additional suites to test different client configurations where we set the desired configs in the `.t` file after sourcing the sharness.d scripts.

There might be an easier/cleaner way to do this but haven't thought through it.

CamStan:
Tagging #442

bootstrap.sh
```diff
@@ -93,7 +93,7 @@ if [ "$automake_sub_version" -lt "15" ]; then
 
     autotools_repos=(http://ftp.gnu.org/gnu/autoconf/autoconf-2.69.tar.gz
         http://ftp.gnu.org/gnu/automake/automake-1.15.tar.gz
-        http://mirror.team-cymru.org/gnu/libtool/libtool-2.4.2.tar.gz
+        http://mirror.team-cymru.org/gnu/libtool/libtool-2.4.6.tar.gz
```
adammoody, line 96: I had to bump libtool up a level because margo seemed to require it.  To get that to work, I had to wipe out my automake install to force this code to rerun.

More autotools fun...

I'm wondering if we shouldn't just have bootstrap always build a particular set of autotools.  That would ensure that everyone building through bootstrap is using a consistent set of tools.

@CamStan , are there problems if disable the automake version check so that this always runs?

CamStan: I don't think there'd be any problems disabling the version check. I only did that to avoid installing additional things if not needed.

FYI: I was able to get bootstrap to work using an install of Autotools in a shared location on our systems as well. I updated that one to use libtool v2.4.6 per this change.

t/0110-statfs-gotcha.t
```diff
@@ -0,0 +1,12 @@
...
+# Source sharness environment scripts to pick up test environment
+# and UnifyFS runtime settings.
+#
+. $(dirname $0)/sharness.d/00-test-env.sh
+. $(dirname $0)/sharness.d/01-unifyfs-settings.sh
+
+# disable statfs from returning UnifyFS! super magic value, return tmpfs magic instead
+export UNIFYFS_CLIENT_SUPER_MAGIC=0
+
+$JOB_RUN_COMMAND $UNIFYFS_BUILD_DIR/t/sys/sysio-gotcha.t
```
CamStan, line 12: I think this is the only thing that needs to be fixed for the new `statfs` suite.
```suggestion
$JOB_RUN_COMMAND $UNIFYFS_BUILD_DIR/t/sys/statfs-gotcha.t
```

CamStan (changes requested):
Nice! Excited to see if this works.

t/0110-statfs-gotcha.t
adammoody: Thanks @CamStan .  Nice catch!

t/Makefile.am
```diff
@@ -158,6 +163,24 @@ std_stdio_gotcha_t_SOURCES = \
...
+sys_statfs_gotcha_t_LDFLAGS = $(test_gotcha_ldflags)
+
+std_stdio_gotcha_t_SOURCES = std/stdio_suite.h \
+                             std/stdio_suite.c \
+                             std/fopen-fclose.c \
+                             std/fseek-ftell.c \
+                             std/fwrite-fread.c \
+                             std/fflush.c \
+                             std/size.c
+>>>>>>> accbacf (test: add configuration to check tmpfs magic)
```
CamStan, line 182: This might be the causing the current failure of 
```
config.status: error: cannot find input file: `t/Makefile.in'
```

One other thing to check is, for some reason, my editor insisted on using tabs instead of whitespace only within this makefile, with was causing other issues for me.

adammoody: Thanks, @CamStan .  Yep, good eye.  I had a stray line from a bad merge conflict resolution.

CamStan (approved):
@adammoody, this looks good to me! Just saw one typo.
Did you run into anything unexpected in the processes to test using different configurations?

If you like, we can merge this and get it in the 0.9.2 release, unless you want to wait till next week to give others a change to look over it.

client/src/unifyfs.c
```diff
@@ -1749,6 +1752,17 @@ static int unifyfs_init(void)
             }
         }
 
+        /* Determine SUPER MAGIC value to return from statfs.
+         * Use UNIFYFS_SUPER_MAIGIC if true, TMPFS_SUPER_MAGIC otherwise. */
```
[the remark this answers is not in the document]
adammoody: I like MAIGIC is better.

adammoody:
Thanks @CamStan and @MichaelBrim .

@sandrain , just a heads up that this is now merged.  It might confuse IOR if you pull dev and rebuild since it now returns a different magic value by default.  You should be able to revert back to the tmpfs magic value by setting ``UNIFYFS_CLIENT_SUPER_MAGIC=0``

sandrain:
@adammoody Thanks for the explanation!

Issue #584, opened 2020-12-03
Update Mochi stack dependencies to later versions


CamStan:
We currently depend on Margo v0.4.3 and Mercury v1.0.1 which are a bit behind their latest versions (v0.9 and v2.0.0 respectively).

Initially ran into this in issue #573 as spackages changed.

It'll be better for us to be on newer software now that we're trying to scale up and debug performance issues and deadlocks that we may get stuck on.

The Mochi team suggested we reach out when we're ready to update. They stated:
> The [Margo] API changes are very minor but I'm not sure which ones you might bump into. There are some nice internal improvements in Margo to look forward to once you do, though, including some performance improvements (in conjunction with the newer Mercury especially), ability to profile RPCs, ability to set/inspect configuration via json, and some other odds and ends.

sandrain:
@CamStan I've tested with mercury-2.0.0 and margo-0.9, and it seems like unifyfs runs fine when using bmi+tcp. It failed with ofi+tcp. My test was on Summit without using spack.

sandrain:
It seems like changing protocol from ```ofi+tcp://``` to ```ofi+sockets://``` makes unifyfs runs fine with new mercury, at least in my testing environment on Summit:

```
diff --git a/server/src/margo_server.c b/server/src/margo_server.c
index 36e74cb..2cdc438 100644
--- a/server/src/margo_server.c
+++ b/server/src/margo_server.c
@@ -41,7 +41,7 @@ static const char* PROTOCOL_MARGO_SHM   = "na+sm://";
 static const char* PROTOCOL_MARGO_TCP = "bmi+tcp://";
 static const char* PROTOCOL_MARGO_RMA = "bmi+tcp://";
 #elif defined(NA_HAS_OFI)
-static const char* PROTOCOL_MARGO_TCP = "ofi+tcp://";
+static const char* PROTOCOL_MARGO_TCP = "ofi+sockets://";
 static const char* PROTOCOL_MARGO_RMA = "ofi+verbs://";
 #else
 #error No supported Mercury NA plugin found (please use one of: 'BMI', 'OFI')
```

In the previous mercury (v1.0.1), the ```ofi+tcp://``` is transformed to ```ofi+sockets://```:

https://github.com/mercury-hpc/mercury/blob/0de22781cd08931416aba8050c5a6647c23abcb9/src/na/na_ofi.c#L127

but in v2.0, ```ofi+tcp://``` becomes ```tcp;ofi+rxm://```

[...]

PR #607, opened 2021-02-17
mochi-margo support

MichaelBrim:
### Description

This PR updates the code for use with the latest releases of the Mochi packages (margo, mercury, argobots). Note that the 'margo' spack package has been deprecated and replaced with 'mochi-margo'.

The Travis CI configuration has also been updated to use mochi-margo with a libfabric dependency, rather than bmi.

We are also now making use of mercury 2.0's support for redirecting its log messages to our own server logs.

### Types of changes

### Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,7 +67,7 @@ before_install:
 install:
   - . $HOME/spack/share/spack/setup-env.sh
   - spack install gotcha@1.0.3 && spack load gotcha@1.0.3
-  - spack install margo@0.4.3^mercury@1.0.1+bmi~ofi~boostsys && spack load argobots && spack load mercury && spack load margo
+  - spack install mochi-margo ^libfabric fabrics=rxm,sockets,tcp && spack load argobots && spack load mercury && spack load mochi-margo
   - spack install spath && spack load spath
   # prepare build environment
   - eval $(./scripts/git_log_test_env.sh)
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -3,23 +3,65 @@
 # This is an easy-bake script to download and build all UnifyFS's dependencies.
 #
 
+make_nproc=4
+
 ROOT="$(pwd)"
 
 mkdir -p deps
 mkdir -p install
+
 INSTALL_DIR=$ROOT/install
+export PATH=$INSTALL_DIR/bin:$PATH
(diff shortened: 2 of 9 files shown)


.travis.yml
```diff
@@ -67,7 +67,7 @@ before_install:
 install:
   - . $HOME/spack/share/spack/setup-env.sh
   - spack install gotcha@1.0.3 && spack load gotcha@1.0.3
-  - spack install margo@0.4.3^mercury@1.0.1+bmi~ofi~boostsys && spack load argobots && spack load mercury && spack load margo
+  - spack install mochi-margo ^libfabric fabrics=rxm,sockets,tcp && spack load argobots && spack load mercury && spack load mochi-margo
```
CamStan, line 70: For future reference and clarification when I update our spackage and build docs:

1. We now work with `^mercury+bmi~ofi`, `^mercury+ofi~bmi`, or both enabled, correct?
2. Do we conflict if `^mercury~bmi~ofi`?
3. Do we now require the specific  `^libfabric fabrics=rxm,sockets,tcp` dependency all the time? Only if `+ofi`? Or will we now always break with the default libfabric that Spack installs for Mercury?

CamStan (approved):
@MichaelBrim, this is great! Thank you for doing this.

A couple questions below to help me when testing our Spack install.

CamStan:
Unless we want to update our `bootstrap.sh` as part of this PR with the changes mentioned in https://github.com/LLNL/UnifyFS/issues/584#issuecomment-744890743, I can test and add that when I update our spackage and build docs.

.travis.yml
MichaelBrim: 1. Yes (but I personally didn't test the both case - Hyogi did that awhile back)
2. Yes
3. That specific `libfabrics fabrics=rxm,sockets,tcp` config is necessary for my Docker builds and possibly Travis since they are both Ubuntu Bionic. However, Mercury v2.0 'ofi+tcp' requires `libfabrics fabrics=rxm,tcp`, but their spackage doesn't properly capture that dependency. The former spec will work on more machines, so it is probably what we want in our spackage.

MichaelBrim:
@CamStan Good catch on the bootstrap script. I always forget about it because I don't use it. I will add another commit that updates bootstrap.sh.

CamStan:
> I always forget about it because I don't use it.

Same. Only remembered since I was just messing with it last week.

> I will add another commit that updates bootstrap.sh.

Sounds good! Thank you!

CamStan (approved):

Issue #634, opened 2021-05-28
Write examples fail with default options and default spillover size


CamStan:
### Describe the problem you're observing
Running the _write_ or _writeread_ examples (static or gotcha) on a single node and with a single process, using the default options and default `spill_size` (1 GiB), fails while trying to write MiB 318.

The example succeeds when increasing the `spill_size` to 2 GiB (possibly less).
 
The default values for these examples are:
https://github.com/LLNL/UnifyFS/blob/dcc3984f6de5b5d758388e22e1dbe57a13026c40/examples/src/testutil.h#L113-L115

and the default `shmem_size` and `spill_size` are:
https://github.com/LLNL/UnifyFS/blob/dcc3984f6de5b5d758388e22e1dbe57a13026c40/common/src/unifyfs_const.h#L79-L80

The expected behavior is that 512 MiB should fit within the default `shmem_size` and `spill_size`.

### Describe how to reproduce the problem
Get a single node allocation, set a few config options, and start the server:
```
export UNIFYFS_DAEMONIZE=off
export UNIFYFS_LOG_DIR=${MY_JOBDIR}/logs
export UNIFYFS_LOG_VERBOSITY=5
export UNIFYFS_LOGIO_SPILL_DIR=$NODE_LOCAL_STORAGE_PATH
export UNIFYFS_SHAREDFS_DIR=${MY_JOBDIR}/share

$UNIFYFS_INSTALL/bin/unifyfs start -d -S $UNIFYFS_SHAREDFS_DIR -e $UNIFYFS_INSTALL/bin/unifyfsd &> $UNIFYFS_LOG_DIR/unifyfs.start.out
```

Run the default `write-static` example with a single process:
```
jsrun -r1 -n1 -a1 -k $UNIFYFS_LOG_DIR/write-static-1GBspill.err -o $UNIFYFS_LOG_DIR/write-static-1GBspill.out $UNIFYFS_EXAMPLES/write-static
```

### Include any warning or errors or releveant debugging data

The stdout log for the client starts getting errors on write req 318 (each request here is 1MiB)
```
$ head ${UNIFYFS_LOG_DIR}/write-static-1GBspill.out
[0] write() failed (errno=28, No space left on device)
[0] write req 318 failed (errno=28, No space left on device)
[0] write() failed (errno=28, No space left on device)
[0] write req 319 failed (errno=28, No space left on device)
⋮
```

[...]

PR #638, opened 2021-06-12
Bugfix: unlink global file not found locally

CamStan:
### Description

Add logic to unlink/remove a file that exists globally when it is not found locally.

Add option to reuse a file name, and a function to remove a file to _testutil.h_.
Add check for reusing a file name to _write_, _writeread_, and _cr_ examples.

Change default `UNIFYFS_LOGIO_SPILL_SIZE` to 4GIB.
Closes: #634 

Minor documentation clarifications.

### Motivation and Context

Idea was brought up to allow rerunning the examples multiple times with the same filename.

Attempting this resulted in being able to stat a file to show it existed, but when attempting to remove a file created by a different client, unifyfs would claim the file didn't exist.

### How Has This Been Tested?

Ran _write-gotcha/static_ example with and without the new `-r|--reuse-filename` option.
E.g.,
```
jsrun -r1 -n2 -a1 -c1 -e individual -k $UNIFYFS_LOG_DIR/write-gotcha.err -o $UNIFYFS_LOG_DIR/write-gotcha.out $UNIFYFS_EXAMPLES/write-gotcha -v -r -f mytestfile

jsrun -r1 -n2 -a1 -c1 -e individual -k $UNIFYFS_LOG_DIR/write-gotcha-reuse.err -o $UNIFYFS_LOG_DIR/write-gotcha-reuse.out $UNIFYFS_EXAMPLES/write-gotcha -v -r -f mytestfile
```

### Types of changes

### Checklist:


client/src/unifyfs-sysio.c
```diff
@@ -616,27 +638,51 @@ int UNIFYFS_WRAP(remove)(const char* path)
...
-            errno = unifyfs_rc_errno(ret);
-            return -1;
+            /* check that it's not a directory */
+            if (attr.mode & S_IFDIR) {
+                /* TODO: shall be equivalent to rmdir(path) */
+                /* ERROR: is a directory */
+                LOGDBG("Attempting to unlink a directory %s in UNIFYFS", upath);
+                errno = EISDIR;
+                return -1;
+            }
```
adammoody, line 660: I think it's valid to call ``remove`` on a directory.  I see we also had this check for a directory in our ``remove`` wrapper before this change.
```
NAME
       remove - remove a file or directory

SYNOPSIS
       #include <stdio.h>

       int remove(const char *pathname);

DESCRIPTION
       remove() deletes a name from the file system.  It calls unlink(2) for files, and rmdir(2) for directories.
```

adammoody (approved):

adammoody:
Thanks @CamStan .  Nice work tracking this down and finding the fix.  This looks good to me, other than the remove-on-directory item.

BTW, our local file cache can get stale at times.  That requires a bigger fix than what we'd want to do here, but this provides a good opportunity to highlight a problem.  For example, I think this related 2-process test won't work like we'd want it to.
```
char[] path = "/unifyfs/foo"
if (rank == 1) {
  mkdir(path) // local cache on rank 1 will record path as a direcotry
}

MPI_Barrier(MPI_COMM_WORLD)

if (rank == 0) {
  rmdir(path) // rank 0 deletes item
  int fd = open(path, O_WRONLY | O_CREAT) // recreates it as a file
  close(fd)
}

MPI_Barrier(MPI_COMM_WORLD)

if (rank == 1) {
  unlink(path) // cache on rank 1 still sees item as directory, even though it's now actually a file, so unlink will fail
}
```

client/src/unifyfs-sysio.c
CamStan: Updated this for when the file is found locally with the following. However, haven't added anything to address this when the dir is found globally and this brought up a few questions. I'll explain this in my comment below.
``` C
         } else { /* got the file locally */
            /* if directory, is it empty? */
            if (unifyfs_fid_is_dir(fid) && !unifyfs_fid_is_dir_empty(upath)) {
                /* ERROR: is a directory, but isn't empty */
                LOGDBG("Attempting to remove non-empty dir %s in UNIFYFS",
                        upath);
                errno = ENOTEMPTY;
                 return -1;
             }
```

CamStan:
@adammoody, I went to update this for having `remove()` behave like `rmdir()` and noticed there may be a few other bugs here.

1. The `rmdir()` wrapper has the same issue that this PR attempts to fix with `unlink()` and `remove()`, in that the existence of the directory is only checked for locally.

2. The `unifyfs_fid_is_dir_empty()` function only checks that the directory is local and in use.
https://github.com/LLNL/UnifyFS/blob/2a0352a3f2a66959157ae828cb16412186f182d3/client/src/unifyfs.c#L653-L658
Would we want to update this to also check globally, write a different function, or just assume that if the directory is only found globally, that it's not in use and is okay to remove (this is assuming 3 is fine as it is)?

3. Realized that our unit test for removing a non-empty dir fails, despite the `rmdir()` wrapper checking for this:
https://github.com/LLNL/UnifyFS/blob/2a0352a3f2a66959157ae828cb16412186f182d3/client/src/unifyfs-sysio.c#L136-L140
I'm assuming this is because of the `in_use` check mentioned in 2. However, if this is the behavior we want, then the solution to just let `remove()`work on a directory only found globally is easy.
```
    PASS: 0100-sysio-gotcha.t 20 - ../../../t/sys/mkdir-rmdir.c:125 creat subfile /tmp/tmp.S9GVtl4fS6/mount/PTTcS._fpC4Z4ztupby0s4@BOxRm+FoOYAyy./subfile in existing dir (fd=30000): Success
    XFAIL: 0100-sysio-gotcha.t 26 - ../../../t/sys/mkdir-rmdir.c:174 rmdir non-empty directory /tmp/tmp.S9GVtl4fS6/mount/PTTcS._fpC4Z4ztupby0s4@BOxRm+FoOYAyy. should fail (errno=0): Success # TODO mkdir_5: unifyfs currently creates all paths as separate entities
```
I can create an issue for these (and the one two-process one you mentioned) so we don't lose them, unless we want to try to address some/all in this PR.

CamStan:
This is ready for another review.
Checkpatch is failing because of my middle commit that I've only kept for reference. If we want to use this approach, and when this is ready to merge, I'll squash these.

MichaelBrim (commented):
I see a couple places in where you check the shared logic function's mask against S_IFLNK, but no calls to the function set that flag in the mask.

CamStan:
> I see a couple places in where you check the shared logic function's mask against S_IFLNK, but no calls to the function set that flag in the mask.

True. Checking for this came up in one of my discussions with Adam, but was before I had thought through the logic.

I just noticed we don't seem to wrap `symlink()` and at first thought, could UnifyFS even have symlinks in it, the way it's currently set up?

I'm not treating them any different than regular files here as the only extra thing our current removes wrappers really care about right now is if something is a directory or not.
Should probably just remove that check then, yeah? Then if we end up supporting them, anything special about their removal can be added.

MichaelBrim:
I don't think we should worry about symlink support at this time. I can't think of a practical use case where an HPC app would need to use a symlink. And you're right that supporting them using our current codebase would actually be quite difficult.

client/src/unifyfs-sysio.c
```diff
@@ -109,42 +109,124 @@ int UNIFYFS_WRAP(mkdir)(const char* path, mode_t mode)
...
 }
 
-int UNIFYFS_WRAP(rmdir)(const char* path)
+/* Function for shared logic between rmdir, remove, and unlink wrappers.
+ *
+ * Remove the target at upath based on the input file mode indicated by mask.
+ * If mask == 0 (e.g., remove wrapper made request), the target doesn't need
+ * any mode specific checks before attempting to remove.
+ * */
+static int unifyfs_remove_mask(const char* upath, mode_t mask)
```
MichaelBrim, line 118: For code readability, I personally think this function should be named differently. If you just read the name, you would think it "removes a mask" from the file, which it doesn't. Perhaps just call it "unifyfs_remove"? The function comment clearly states the purpose of the mask parameter.

CamStan: Good call. Changed this.

MichaelBrim (approved):

PR #641, opened 2021-06-14
Update semantics description in documentation

kathrynmohror:
### Description
Our description of UnifyFS semantics in our documentation is woefully out of date. This PR fleshes out the description and updates it to be current with the implementation.

### Motivation and Context
Reorganize section, add more detail, clarify semantics because sorely out of date.

### How Has This Been Tested?
Not tested

### Types of changes

- [x ] Documentation (a change to man pages or other documentation)

### Checklist:

- [x ] I have updated the documentation accordingly.

### DIFF
--- a/docs/assumptions.rst
+++ b/docs/assumptions.rst
@@ -77,17 +77,17 @@ of UnifyFS.
 The use of synchronization operations are required for applications that exhibit
 I/O accesses that deviate from the bulk synchronous I/O pattern.
 There are two types of synchronization that are required for correct execution
-of parallel I/O on UnifyFS: *local synchronization* and *global synchronization*.
+of parallel I/O on UnifyFS: *local synchronization* and *inter-process synchronization*.
 Here, *local synchronization* refers to synchronization operations performed
 locally by a process to ensure that its updates to a file are visible
 to other processes. For example, a process may update a region of a file
 and then execute fflush() so that a different process can read the updated
 file contents.
-*Global synchronization* refers to synchronization
+*Inter-process synchronization* refers to synchronization
 operations that are performed to enforce ordering of conflicting I/O operations
 from multiple processes.
-These global synchronizations occur outside of normal file I/O operations and
-typically involve interprocess communication, e.g., with MPI. For example,
+These inter-process synchronizations occur outside of normal file I/O operations and
+typically involve inter-process communication, e.g., with MPI. For example,
 if two processes need to update the same file region and it is important to
 the outcome of the program that the updates occur in a particular order, then
 the program needs to enforce this ordering with an operation like an MPI_Barrier()
@@ -114,7 +114,7 @@ requirements of UnifyFS.
         or by supplying the client.write_sync configuration parameter to UnifyFS
         on startup, which will cause an implicit "flush" operation after
         every write (note: use of the client.write_sync mode can significantly slow down
-        write performance.). In this case, global synchronization is still required
+        write performance.). In this case, inter-process synchronization is still required
         for applications that perform conflicting updates to files.
 
 During a write phase, a process can deviate from the bulk synchronous
@@ -142,7 +142,7 @@ In summary, reading the local data (which has been written by processes
 executing on the same compute node) will always be faster than reading
 remote data.
 
-Note that, as we discuss above, commit semantics also require global synchronization
+Note that, as we discuss above, commit semantics also require inter-process synchronization
 for potentially conflicting
 write accesses. If an application does not enforce sequential ordering of file
 modifications during a write phase, e.g., with MPI synchronization,
@@ -164,9 +164,6 @@ laminated, it becomes permanently read-only and its data is accessible across
 all the compute nodes in the job without further synchronization.
 Once a file is laminated, it cannot be further modified,
 except for being renamed or deleted.
-.. Is the next sentence true? Does more need to be added?
-If a failure occurs during a job before a file is laminated, the file
-contents may be unrecoverable.
 
 A typical use case for lamination is for checkpoint/restart.
 An application can laminate checkpoint files after they have
--- a/docs/assumptions.rst
+++ b/docs/assumptions.rst
@@ -21,8 +21,8 @@ The system requirements to run UnifyFS are:
 Application Behavior
 ---------------------------
 
-UnifyFS is specifically designed to support the bulk synchronous I/O patterns
-that are typical in HPC applications, e.g., checkpoint/restart or output dumps.
+UnifyFS is specifically designed to support the bulk synchronous I/O pattern
+that is typical in HPC applications, e.g., checkpoint/restart or output dumps.
 In bulk synchronous I/O, I/O operations occur in separate write and read phases,
 and files are not read and written simultaneously.
 For example, files are written during checkpointing (a write phase)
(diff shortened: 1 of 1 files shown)


docs/assumptions.rst
```diff
@@ -1,63 +1,159 @@
...
+executing on the same compute node) will always be faster than reading
+remote data.
+
+Note that commit semantics also require synchronization for potentially conflicting
+write accesses. If an application does not enforce sequential ordering of file
+modifications during a write phase, e.g., with MPI synchronization,
+and multiple processes write concurrently to the same file offset or to an
+overlapping region, the result is undefined and may
+reflect the result of any of the processes' operations to that offset or region.
+**I don't think this paragraph is true. I think we won't return the last write even if synchronization is applied**
```
MichaelBrim, line 131: I think the paragraph is mostly right in that concurrent writes to the same offset/region will result in undefined behavior, but perhaps it's not as clear as it could be. The problem with the current discussion is we're overloading "synchronization operations" to mean two things that are not equivalent. One type is local synchronization (i.e., flush extents to the server), the other type is global synchronization across app processes (which must be provided by something like MPI since Unify does not provide anything for that). For a consistent distributed sequential ordering of conflicting writes, there must be global sync ops between each of the local sync ops.

docs/assumptions.rst
```diff
@@ -1,63 +1,159 @@
...
+'''''''''''''''''''''''''''
+
+The other consistency model that UnifyFS employs is called "lamination
+semantics" which is intended to be applied once a file is done being modified
+at the end of a write phase of an application.  After a file is
+laminated, it becomes permanently read-only and its data is accessible across
+all the compute nodes in the job without further synchronization.
+Once a file is laminated, it cannot be further modified,
+except for being renamed or deleted.
+.. Is the next sentence true? Does more need to be added?
```
MichaelBrim, line 146: I think we need a separate section on failures, because it really depends on where the failure occurs, regardless of lamination state. If a node goes down, it takes the server with it, and all data from the node is lost. If the application has a terminal failure, and the server is still running, another local or remote client can still access the data. If the server encounters a terminal failure, all data on the node is lost.

docs/assumptions.rst
```diff
@@ -66,18 +162,32 @@ file. A typical checkpoint will look like:
...
+By assuming that processes do not need to access file data modified
+by other processes before lamination,
+UnifyFS can optimize write performance by buffering all metadata and
+file data for processes locally, instead of performing costly exchanges of
+metadata and file data between compute nodes on every write.
+Also, since file contents cannot change after lamination,
 aggressive caching may be used during the read phase with minimal locking.
 Further, since a file may be lost on application failure unless laminated, data
 redundancy schemes can be delayed until lamination.
+.. do we need to define our failure behavior better?
```
MichaelBrim, line 186: yes, see above comment

docs/assumptions.rst
```diff
@@ -125,12 +231,12 @@ The additional behavior of UnifyFS can be summarized as follows.
       options.
 
     - UnifyFS also can be coupled with SymphonyFS_, high level I/O libraries, or
-      a checkpoint library (VeloC_) to move data to PFS periodically.
+      a checkpoint library like (SCR_) or (VeloC_) to move data to the PFS periodically.
 
-    - UnifyFS can be used with checkpointing libraries (VeloC_) or other I/O
-      libraries to support shared files on burst buffers.
+    - UnifyFS can be used with checkpointing libraries like (SCR_) or (VeloC_),
+      or with I/O libraries libraries like (HDF5_) to support shared files on burst buffers.
```
MichaelBrim, line 237: "libraries libraries"

docs/assumptions.rst
```diff
@@ -139,9 +245,15 @@ The additional behavior of UnifyFS can be summarized as follows.
 
     - UnifyFS survives across multiple application runs within a job.
 
-    - UnifyFS will transparently intercept system level I/O calls of
+    - If a failure occurs during a job before a file is laminated, the file
+      contents may be unrecoverable.
+.. is this adequate to describe failure behavior?
```
MichaelBrim, line 250: no, see above comment

MichaelBrim (changes requested):

docs/assumptions.rst
kathrynmohror: Thanks @MichaelBrim. I added a new section on failure behavior

docs/assumptions.rst
```diff
@@ -1,63 +1,159 @@
...
+a file, including remote data that has been written by processes
+executing on remote compute nodes in the job.
+However, the performance will differ based on which process wrote the data that
+is being read:
+      - If the bytes being read were written by the same process that is reading
+        the bytes, UnifyFS offers the fastest performance and no synchronization
+        operations are needed. This kind of access is typical in some I/O
+        libraries, e.g., HDF5, where file metadata may be updated and read by
+        the same process.
+      - If the bytes being read were written by a process executing on the same compute
```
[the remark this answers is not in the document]
kathrynmohror: I agree. I misread our design document and got it in my head that the library could read any of the local extent trees (from any process). I fixed it to say that sync ops are needed for this case.

docs/assumptions.rst
kathrynmohror: I added a discussion of local and global synchronization near the top of this section. I think it helps, thanks for suggesting it. 

I still wonder about whether the most recent update will always be returned, even if correct global synchronization is used. Perhaps, though, this was a problem with MDHIM and is fixed with our new approach? I think with MDHIM we ended up with multiple key,value pairs for the same offset and we didn't know which one was most recent so one would be returned (but perhaps not the most recent). Perhaps now with our new approach, there is just one location where data for an offset is stored and when a sync operation is applied, that location is updated with the most recent data. Is that right?

docs/assumptions.rst
```diff
@@ -1,63 +1,180 @@
...
-workloads. In such a workload, the expected application behavior is as follows.
+UnifyFS is specifically designed to support the bulk synchronous I/O pattern
+that is typical in HPC applications, e.g., checkpoint/restart or output dumps.
+In bulk synchronous I/O, I/O operations occur in separate write and read phases,
+and files are not read and written simultaneously.
+For example, files are written during checkpointing (a write phase)
+and read during recovery/restart (a read phase).
+Additionally, parallel writes and reads to shared files occur systematically,
+where processes access computable, regular offsets of files, e.g., in strided or
+segmented access patterns, in contrast to random, interleaved, small writes and reads.
```
adammoody, line 32: maybe also call out "overlapped writes" in the list of things not to do in bulk synchronous mode?

docs/assumptions.rst
```diff
@@ -1,63 +1,180 @@
...
+Commit Consistency Semantics in UnifyFS
+'''''''''''''''''''''''''''
+.. _commit_consistency_label:
+
+Commit consistency semantics require
+explicit "commit" operations to be performed before updates to a file
+are globally visible.
+We chose commit consistency semantics for UnifyFS because it is sufficient
+for correct execution of typical HPC applications that adhere to
+the bulk synchronous I/O pattern, and enables UnifyFS to provide better
```
adammoody, line 68:
```suggestion
the bulk synchronous I/O pattern, and it enables UnifyFS to provide better
```

docs/assumptions.rst
adammoody: For "global sync", the word "global" is a bit strong.  What really matters is "inter-process sync", just the subset of processes that care about some shared offset, which might just be two procs out of many.  It's more like "file system synchronization" and "process synchronization".

adammoody: That's right, we don't have the problem with two key/val pairs existing for a single file offset anymore.  A single owner server is responsible for updating the main extent tree for a file.  Old entries are replaced by new entries.

Another thing regarding "undefined".  From the user's point of view, if two procs write to the same region, the result could be some strange mix of bytes from the two, even if each process issued one write.  For example, consider that two procs each do a single large write of say 1GB to the same region of the file.  Internally, each single large write may be broken into a bunch of smaller chunks/extents, which are sent to the owner server maintaining the extent tree in numerous rpc calls.  Those then get played on the server in random order.  The result is that we might get a mix of data from the two procs for that region.  It's not as if the single large write from one of the two procs wins and the other loses.  Our current implementation might actually do better than what I'm describing, but I think it'd be good to stress the "undefined" bit until users complain since it gives us some flexibility.

docs/assumptions.rst
```diff
@@ -66,21 +183,50 @@ file. A typical checkpoint will look like:
...
-lamination occurs.  Also, since file contents cannot change after lamination,
+supporting the typical requirements of bulk synchronous I/O.
+Recall that for bulk synchronous I/O patterns, reads and writes typically occur in
+distinct phases. This means that for the majority of the time,
+processes do not need to read arbitrary
+bytes of a file until the write phase is completed, which in practice is
+when the file is done being modified and closed and can be safely made
+read-only with lamination.
+By assuming that processes do not need to access file data modified
+by other processes before lamination,
```
adammoody, line 199: Maybe tweak this to highlight that "well-behaved" apps can benefit from the extreme write buffering optimization.
```suggestion
For applications in which processes do not need to access file data modified
by other processes before lamination,
```

docs/assumptions.rst
adammoody: Just for the team... I wouldn't change the text here, but technically we should be able to handle random and interleaved writes just fine.

We could also handle small writes efficiently.  Chunk space utilization and our extent tree can run into problems.  For extreme cases, we have it on our roadmap that we could also support a stripe-based storage model like Lustre rather than using a log-based I/O model.  The stripe-based method eliminates the need for extents and it can fully pack the storage, though it means we need to transfer data bytes on sync operations rather than extent info.  The write data could still be buffered locally until the sync/laminate, though.

Small local reads should also be decent in our current model, but lots of small remote reads will be bad.

adammoody: On when to switch to a striped-based method, there are a few interesting options:
1) allow user to set that as a config
2) use log-I/O during write, reformat to a stripe-based layout at laminate
3) track whether the extents for a file consume more space than the data as the file is being written, and switch layouts on the fly

docs/assumptions.rst
```diff
@@ -1,63 +1,180 @@
...
+        operations explicitly in their codes.
+      - With explicit synchronization. If an application does not use a compliant
+        parallel I/O library or if the developer wishes to perform explicit
+        synchronization, local synchronization can be achieved through adding
+        explicit "flush" operations with calls to fflush(), close(), or fsync()
+        in the application source code,
+        or by supplying the client.write_sync configuration parameter to UnifyFS
+        on startup, which will cause an implicit "flush" operation after
+        every write (note: use of the client.write_sync mode can significantly slow down
+        write performance.). In this case, global synchronization is still required
```
MichaelBrim, line 117: remove period before closing paren

docs/assumptions.rst
MichaelBrim: +1

docs/assumptions.rst
kathrynmohror: I made some updates to introduce the concept of overlapping writes and to drive home that the result in that case is undefined.

adammoody:
@MichaelBrim , the merge button is blocked pending a change request from you.  Is that good to go now?  If so, do you see where you can sign off again?

MichaelBrim (approved):

MichaelBrim:
@adammoody should be good to merge now

Issue #645, opened 2021-07-12
avoid unnecessary fsync when sync'ing extents with server


adammoody:
In ``unifyfs_sync_extents``, the client fsyncs all data to disk *before* it calls the rpc to sync extents with the server:

https://github.com/LLNL/UnifyFS/blob/fddedf29335cf3359c51be47017616a61a1f51cd/client/src/unifyfs-fixed.c#L192

We sync data to disk by calling the logio_sync call here:

https://github.com/LLNL/UnifyFS/blob/fddedf29335cf3359c51be47017616a61a1f51cd/client/src/unifyfs-fixed.c#L226

Then we tell the server to copy extents from the client extent buffer with an rpc call here:

https://github.com/LLNL/UnifyFS/blob/fddedf29335cf3359c51be47017616a61a1f51cd/client/src/unifyfs-fixed.c#L235

There are two main reasons we sync extents with the server:
1) the application has written enough extents that we've exhausted the extent buffer, in which case, we flush that buffer to the server so that the client can handle more writes
2) the application has explicitly called something like fsync() or fflush()

In case 1), when we're using a POSIX filesystem as our backing store, which will be the case for tmpfs and the file system on most SSDs, then we can avoid syncing data to disk completely.  POSIX ensures that the server process will read the correct data even though it was written by another process, without requiring an fsync.

I think we could similarly avoid fsync'ing data to disk on calls to close().  We currently sync on close, but that's mainly to send extents to the server, not really to store data to disk.

Even in case 2), we could likely sync data to disk *after* we call the rpc to expose the extents to the server.  That might allow for some overlap in letting the server fetch the extents while data is being sync'd to disk so that those two operations can happen in parallel.  It may require adding a second rpc to actually allow overlap.

The fix for case 1) to avoid calling fsync at all should speed up UnifyFS performance for applications that write lots of extents and rarely explicitly call fsync.

To support this, we likely need to add another argument to ``unifyfs_sync_extents`` that tells the function whether an fsync is needed.  Then all callers need to be updated to add that argument in their call.  The implementation could use the new unifyfs client library, which has split those into separate calls.

CamStan:
Addressed by #651?

adammoody:
[...]

PR #655, opened 2021-07-27
Ensure logio data is persisted on user fsync()

MichaelBrim:
### Description

This PR makes it so we persist logio data to storage when a client uses `fsync()`, unless the `client.fsync_persist` configuration option is used to disable this behavior.

### Motivation and Context

See issue #645

### How Has This Been Tested?

Tested in Ubuntu Docker container

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -1933,13 +1933,20 @@ int UNIFYFS_WRAP(fsync)(int fd)
         }
 
         /* invoke fsync rpc to register index metadata with server */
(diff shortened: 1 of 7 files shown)


adammoody:
Great!  Thanks, @MichaelBrim !

Since this is moved up into the wrapper, I think we'll want to capture any other calls that do something similar.  I think ``fdatasync`` and ``sync`` would be on that list, assuming we have wrappers for them.

I also checked ``fflush``, but its man page says that it doesn't sync data to disk.  Instead, one must call ``fflush`` and then follow that with something like ``fsync``.

adammoody (approved):

Issue #197, opened 2018-11-09
Add support for unlink()


adammoody:
We need to be able to delete files.  Currently calling unlink() is a NOP:

- delete all index info associated with file
- free storage in all superblocks (memory and spill over)

This is tricky in that with traditional parallel file systems, only one process must call unlink().  For us to match that behavior, the unlink() from a single client must send a message to the delegators, which must then reach into the superblock file structures on each client to free off storage resources.  To do that, the delegator will need to lock the superblock for write access so it doesn't conflict with changes the client might be making.

adammoody:
An alternative design that will simplify things is to restructure things so that the server manage the storage chunks.  Writes from the client would then transfer data from client to server, and the server would write data.  Reads would not be changed.  This will enable the server to delete a file and free its storage without needing to share locks with the client.

PR #656, opened 2021-07-29
release logio storage for extents on file unlink

MichaelBrim:
### Description

Each server is responsible for releasing the local extents for the target file. For extents that have not yet been synced to the local server, the client must release the storage.

Also includes a bug fix for extents that happened to span the shmem and spill portions of the log. The bug was revealed by the new library API test that makes sure we can properly reclaim storage for deleted files (`t/api/storage-reuse.c`).

### Motivation and Context

See issue #197

### How Has This Been Tested?

Tested in Ubuntu Docker

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs_fid.c
+++ b/client/src/unifyfs_fid.c
@@ -92,6 +92,24 @@ static int fid_storage_free(unifyfs_client* client,
     unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(client, fid);
     if ((meta != NULL) && (meta->fid == fid)) {
         if (meta->storage == FILE_STORAGE_LOGIO) {
+            /* client needs to release unsynced write extents, since server
+             * does not know about them */
+            seg_tree_rdlock(&meta->extents_sync);
+            struct seg_tree_node* node = NULL;
+            while ((node = seg_tree_iter(&meta->extents_sync, node))) {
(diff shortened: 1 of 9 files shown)


adammoody:
Nice work, @MichaelBrim !  That was fast!

There are are few more aspects that I think we'll need to handle.  But this is already a big improvement, so we could do these other items as future work.

Since the slot map is shared for read/write between the server and client, we'll want to lock it to control concurrent access.  The client might be write to a file and allocate new slots at the same time the server is releasing slots due to an unlink.  Actually, based on your code here, I'm guessing adding locking might be pretty straight-forward.

I think overwriting a file region will leak slots.  The server will only hold the most recent extents for all regions.  I think we currently forget about (drop) extents for the old data.  One fix would be to free old slots as they are overwritten, though that sounds like it would slow down our writes, especially since those old extents could be remote.  Alternatively, we could accumulate a list of old extents on the server and delete those on unlink.  I think there is a similar issue on the client if it happens to overwrite a region before it has synced its extents with the server.

There will be metadata on the clients about the unlinked file that we probably need to clean up or invalidate.  In particular, the metadata cache and the client extent / segment trees.  That's less to do with the data storage, but it's one more thing that comes to mind.

adammoody (approved):

adammoody:
Thanks again, @MichaelBrim !

PR #659, opened 2021-08-03
use ABT after calling margo_init

adammoody:
This removes the call to ``ABT_init`` and defers that task to Margo.  Margo configures ABT in a particular way.

In particular, it increases the default ABT thread stack size from 4KB to 2MB:
https://github.com/mochi-hpc/mochi-margo/blob/ff094a5a26af19041860f9bba1e0e938c0194ae5/src/margo-init.c#L1533-L1534

which resolves some problems as described in:
https://github.com/pmodels/argobots/issues/333

Alternatively, we could call ``ABT_init`` before ``margo_init``.  However, we likely would need to configure ABT to be compatible with what Margo expects.  We need to set some ABT environment variables before calling ``ABT_init`` with something like:
```
    char env_str[64];
    sprintf(env_str, "%d", 2*1024*1024);
    setenv("ABT_THREAD_STACKSIZE", env_str, 0);

    ABT_init(argc, argv);
```
similar to what is done in margo here:
https://github.com/mochi-hpc/mochi-margo/blob/ff094a5a26af19041860f9bba1e0e938c0194ae5/src/margo-util.c#L29-L36

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_server.c
+++ b/server/src/unifyfs_server.c
@@ -396,8 +396,6 @@ int main(int argc, char* argv[])
     }
 
     LOGDBG("initializing rpc service");
-    ABT_init(argc, argv);
-    ABT_mutex_create(&app_configs_abt_sync);
     rc = configurator_bool_val(server_cfg.margo_lazy_connect,
                                &margo_lazy_connect);
     rc = configurator_bool_val(server_cfg.margo_tcp,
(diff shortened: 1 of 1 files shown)


CamStan (approved):
Looks good! Assuming @MichaelBrim doesn't have a reason we need to initialize Argobots first.

adammoody:
Also, we should remember that if we're seeing strange segfaults in the future to try increasing ``ABT_THREAD_STACK_SIZE``.  Considering that, it's kind of a bummer that margo calls ``setenv`` with overwrite set to ``1``:
```
setenv("ABT_THREAD_STACKSIZE", env_str, 1);
```
https://github.com/mochi-hpc/mochi-margo/blob/ff094a5a26af19041860f9bba1e0e938c0194ae5/src/margo-util.c#L34

That means their setting will overwrite anything the user has set in their environment.  Ugh.

I think they do read values from their config file, though:
https://github.com/mochi-hpc/mochi-margo/blob/ff094a5a26af19041860f9bba1e0e938c0194ae5/src/margo-init.c#L1528

I think environment variables are easier to work with if we have to change things.  We could ask the margo team about switching the ``setenv`` to use ``overwrite=0``.

MichaelBrim (approved):
At one time, there was a reason to call ABT_init() separately. I don't remember why now, and can't find a code path that should require that now, so this is OK by me.

Issue #660, opened 2021-08-03
Memory leaks in unifyfsd


rgmiller:
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | Linux
OS Version             | Ubuntu 20.04
Architecture           | x86_64
UnifyFS Version        | dev branch

### Describe the problem you're observing
I've been running unifyfsd with valgrind, and it (valgrind) is reporting a number of memory leaks.  Most of the errors seem to be pretty benign: allocate some space at program startup and forget to free it before the program ends.  It looks like some, though, do have the potential to leak memory continuously during a run.

### Describe how to reproduce the problem
I've been using `valgrind --leak-check=full --show-leak-kinds=all`, so it's reporting pretty much everything.

### Include any warning or errors or relevant debugging data

Below is a partial list of the specific errors I've confirmed so far:

- Never calling `unifyfs_config_fini()` to free up configuration variables
- ~~Never calling `ABT_finalize()` to free up the Argobots data~~
  - Fixed by #659 
- Not calling `ABT_mutex_free()` to free up the `app_configs_abt_sync` mutex
- ~~Not freeing the `elems` pointer when cleaning up `arraylist_t` structs~~
  -  Fixed by ea2f5f5e
- Global `server_info_t` array allocated but never freed
  - margo_server.c, lines 37 & 506
- When the request manager struct (reqmgr_thrd_t) is freed, the `client_reqs` array list it (probably) allocated is not freed
  - Happens every time a request manager thread is cleaned up
  - Same is true for the `client_callbacks` array list
- `unifyfs_inode_tree_clear()` deletes inode structs without first deleting the allocated members of those structs
- Request manager thread doesn't free the `reqs_sync` mutex when the thread exits
- `fattr.filename` in `create_mountpoint_dir()` is allocated using `strdup()`, but never freed.
  - This one is going to be tricky because there appears to be no good place to do the `free()`.  See the comment below for details.

CamStan:
> * Never calling ABT_finalize() to free up the Argobots data

Will #659 solve this one as UnifyFS will no longer be directly calling `ABT_init()`?

rgmiller:
> Will #659 solve this one as UnifyFS will no longer be directly calling `ABT_init()`?

I think so.  I'll update the list above.

rgmiller:
Fixing the `fattr.filename` leak is going to be tricky because there appears to be no good place to free it.

[...]

PR #673, opened 2021-08-19
Fix various memory leaks in unifyfsd

rgmiller:
This is a collection of commits that fix a variety of memory leaks in the daemon.
### Description

There were a variety of memory leaks found by running valgrind against the daemon.  Most of the leaks were minor (failing to release memory when the daemon exited), but a few did seem to leak memory continuously as the daemon ran.

### Motivation and Context

Resolves issue #660 

### How Has This Been Tested?

Testing was done by running unifyfsd within the context of valgrind on an Ubuntu workstation.  A few simple clients were run to read & write data and then the daemon was killed.  The resulting valgrind report was analyzed.

### Types of changes

### Checklist:

### DIFF
--- a/server/src/unifyfs_server.c
+++ b/server/src/unifyfs_server.c
@@ -53,6 +53,9 @@ size_t glb_num_servers;     // size of glb_servers array
 
 unifyfs_cfg_t server_cfg;
 
+/* arraylist to track failed clients */
+arraylist_t* failed_clients; // = NULL
+
 static ABT_mutex app_configs_abt_sync;
 static app_config* app_configs[UNIFYFS_SERVER_MAX_NUM_APPS]; /* list of apps */
 static size_t clients_per_app = UNIFYFS_SERVER_MAX_APP_CLIENTS;
@@ -640,6 +643,9 @@ static int find_rank_idx(int my_rank)
 static int unifyfs_exit(void)
 {
     int ret = UNIFYFS_SUCCESS;
+    /* Note: ret could potentially get overwritten a few times.  Since this
+       is shutdown/cleanup code, we'll do as much cleanup as we can and just
+       return the most recent value of ret. */
 
     /* iterate over each active application and free resources */
     LOGDBG("cleaning application state");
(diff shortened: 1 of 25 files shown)


MichaelBrim (changes requested):
For the `unifyfs_inode_tree_clear()` changes, use`unifyfs_inode_destroy(oldnode)` instead of freeing the allocations inside the inode directly.

server/src/unifyfs_server.c
```diff
@@ -675,6 +678,11 @@ static int unifyfs_exit(void)
     LOGDBG("stopping rpc service");
     margo_server_rpc_finalize();
 
+    ret = ABT_mutex_free(&app_configs_abt_sync);
```
MichaelBrim, line 681: Not sure about this, but I think if `margo_finalize()` is being called as part of `margo_server_rpc_finalize()` above, and we're relying on Margo to init/fini Argobots, then we may not be able to safely call any ABT_xxx methods after that.

rgmiller:
> For the `unifyfs_inode_tree_clear()` changes, use`unifyfs_inode_destroy(oldnode)` instead of freeing the allocations inside the inode directly.

That will require some changes in unifyfs_inode.c:  `unifyfs_inode_destroy()` is currently a static function.  If we want to call it from `unifyfs_inode_tree_clear()`, we'd have to change that.  I don't know if that will cause issues elsewhere or not.

MichaelBrim:
@rgmiller if you are happy with the current code, please do a rebase of this versus current dev and squash down to one commit.

rgmiller:
> @rgmiller if you are happy with the current code, please do a rebase of this versus current dev and squash down to one commit.

@MichaelBrim Done - sort of.  It seems to have squashed down to 4 commits.  The fskv_init thing has already been merged in and two others were merges of the latest dev branch changes back into the dev branch on my fork.  I'm guessing that's why they didn't get squashed.  Why the fskv_init commit is showing up at all still baffles me.

rgmiller:
I resolved the conflict in unifyfs_inode.c.  Not sure about unifyfs_fid.c.

MichaelBrim:
@rgmiller Because of the intermediate merges, it doesn't appear there is an easy way to get rgmiller:dev back in a state where there is a single commit versus LLNL:dev. My suggestion would be to start over with a new branch based on LLNL:dev and re-do the necessary changes in that branch, and then submit a new PR.

rgmiller:
What about just squashing the commits during the merge?  GitHub has an option to do that, doesn't it?

MichaelBrim:
Currently this has a bunch of "changes" unrelated to your memory leak fixes, which makes it hard to track where the code came from and why it changed. In general though, we try to avoid having merge commits in the history. I can help clean this up if you want me to.

rgmiller:
OK, I've created a new branch (in my forked repo) with just the memory leak fixes.  I'll close this pull request and create a new one using the new branch.

Issue #715, opened 2022-06-08
Error building with GCC11 on Summit with PMIx enabled


rgmiller:
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | RHEL
OS Version             | 8.2
Architecture           | ppc64le
UnifyFS Version        | Git dev branch

### Describe the problem you're observing
When building UnifyFS on Summit with PMIx enabled, GCC 11 will issue a warning about a size mismatch on one of the parameters.

The problem is that the parameter is defined as 'const char[512]`, but the caller is expected to pass in one of several #defined strings, all of which are less than 512 bytes.  GCC 11 complains about possibly reading 512 bytes from an area that is smaller than that.

(The bug might be applicable to other implementations of PMIx.  It is notably *NOT* a problem for OpenPMIx implementation, because that implementation defines the parameter in question slightly differently.)

### Describe how to reproduce the problem
Configure UnifyFS to enable PMIx and use gcc 11.x as the compiler
Attempt to build on a system that uses IBM's Spectrum MPI

PR #716, opened 2022-06-08
Disable GCC11 warning on call to  PMIx_Get()

rgmiller:
This PR adds `#pragma` statements to suppress a specific warning generated by gcc 11 when the `PMIx_Get()` function is called.

### Description

The version of PMIx included with Spectrum MPI defines the PMIx_Get()
function to take a parameter that ends up being a 'const char[512]'.
However, the caller is expected to pass in a static const string that is
less than 512 bytes long.  GCC11 issues a warning about this size
mismatch, and since we build with `-Werror`, that breaks the build.

This commit adds some #pragma lines to disable that particular check
for the call to PMIx_Get() when building with GCC 11.

### Motivation and Context

See issue #715

### How Has This Been Tested?

Built on Summit with GCC 10.2.0 & 11.2.0

### Types of changes

Note:  the bug is actually in IBM's implementation of PMIx, not UnifyFS.
### Checklist:

### DIFF
--- a/common/src/unifyfs_keyval.c
+++ b/common/src/unifyfs_keyval.c
@@ -372,7 +372,25 @@ int unifyfs_pmix_init(void)
     PMIX_PROC_CONSTRUCT(&proc);
     strlcpy(proc.nspace, pmix_myproc.nspace, PMIX_MAX_NSLEN);
     proc.rank = PMIX_RANK_WILDCARD;
+
+// This is a kludge for working around a bug in the version of PMIx installed
+// on Summit (which is part of Spectrum MPI 10.4.0.3).  Specifically, gcc 11
+// complains that the 2nd parameter - which is type 'const pmix_key_t', which
+// in turn is 'char [512]' - is longer than the const string that actually gets
+// passed in (which is itself defined in pmix_common.h).
+// It would be nice to limit this to just Summit, but there doesn't seem to be
+// an easy way to do that (no pre-defined macro we can check, for example).
+// Once the bug is fixed and the version on Summit is updated, we can get rid
+// of this test and just use the line in the 'else' clause
+#if (__GNUC__ >= 11)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-overread"
     rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &valp);
+#pragma GCC diagnostic pop
+#else
+    rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &valp);
+#endif
+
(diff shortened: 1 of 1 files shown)


MichaelBrim:
@rgmiller Your current solution is too hacky. Why not do this instead?
```
char key[512];
strcpy(key, PMIX_JOB_SIZE);
PMIx_Get(...,key,...);
```

rgmiller:
@MichaelBrim 

Yeah, it's hacky.  In my defense, I do not want that code to be there forever.  I want to rip it all out once IBM fixes their PMIx code.  That said, we could do an explicit copy, though it would have to be something more like:
```
char key[PMIX_MAX_KEYLEN+1];
strcpy(key, PMIX_JOB_SIZE);
PMIx_Get(&proc, key, NULL, 0, &valp);
```
The PMIx spec only says that `PMIX_MAX_KEYLEN` should be at least 63 bytes, so you can't count on it being 512.

Another, cleaner option would be:
```
pmix_key_t key = PMIX_JOB_SIZE;
rc = PMIx_Get(&proc, key, NULL, 0, &valp);
```
That would work for Summit, but you'd run into problems on systems where PMIx didn't define a `pmix_key_t`.  This includes the OpenPMIx reference implementation (which is what Spack builds).  

I think the proper fix is to change all the defines in pmix_common.h from lines like:
```
#define PMIX_JOB_SIZE                       "pmix.job.size"         // (uint32_t) #procs in this job
```
to lines like:
```
const pmix_key_t PMIX_JOB_SIZE = "pmix.job.size";         // (uint32_t) #procs in this job
```
That's something that IBM developers have to do, though.  (I need to figure out how to file a bug report with them, though.)

common/src/unifyfs_keyval.c
```diff
@@ -372,7 +372,15 @@ int unifyfs_pmix_init(void)
...
     strlcpy(proc.nspace, pmix_myproc.nspace, PMIX_MAX_NSLEN);
     proc.rank = PMIX_RANK_WILDCARD;
-    rc = PMIx_Get(&proc, PMIX_JOB_SIZE, NULL, 0, &valp);
+
+    // Note: we do an extra copy because passing PMIX_JOB_SIZE directly to
+    // PMIx_Get() causes gcc 11 to generate a warning due to the fact that
+    // PMIX_JOB_SIZE evaluates to a 14 byte char array while pmix_key_t is
+    // (at least) 64 bytes.
+    pmix_key_t key;
+    strcpy(key, PMIX_JOB_SIZE);
```
MichaelBrim, line 381: use `strlcpy(key, PMIX_JOB_SIZE, sizeof(key))`

MichaelBrim (approved):

PR #725, opened 2022-07-20
Changes to support Node Local extents.

hariharan-devarajan:
### Description
Currently, unifyfs clients read node-local data from the server thread which is a bottleneck.
In the case of Laminated files, we can fetch the node-local extents from the server and perform read locally.
With the new proposed changes
- The client has a new configuration for node_local_extents
- On the first read of a laminated file if client.node_local_extents is enabled we fetch extents from the server and update the local extents
- This allows reads to be executed locally
- on server we get our extents using unifyfs_invoke_find_extents_rpc

In these cases, we do only one RPC per request group and fetch extents for all laminated files.

### Motivation and Context
This feature would enable faster reads by clients in the case of laminated files.
Some examples of this are,

- read-only workload where data is cached in unifyfs node-locally.
- producer/consumer use cases like MuMMI are colocated in the same node.

### How Has This Been Tested?
In the API test case, I have enabled node_local_extents to true. This works with existing tests well as for non-laminated files this optimization is not applied. Whereas for laminated files it is.

### Types of changes

- [X] New feature (non-breaking change which adds functionality)
- [X] Performance enhancement (non-breaking change which improves efficiency)

### Checklist:

- [X] My code follows the UnifyFS code style requirements.
- [X] I have read the **CONTRIBUTING** document.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.
- [X] All commit messages are properly formatted.


client/src/client_read.c
```diff
@@ -597,9 +597,67 @@ int process_gfid_reads(unifyfs_client* client,
...
+                cur->next = NULL;
+            }
+            meta->needs_extents_sync = 0;
+        }
+        if (num_request_selected > 0) {
+            /* There are files which are laminated and
+             * require sync of extents */
+            int extent_count = 0;
+            extents_list_t extents = NULL;
+            int ret =
```
MichaelBrim, line 633: redefinition of 'int ret'

MichaelBrim: perhaps use 'int rc' instead. conventionally, 'ret' should be the function return value.

common/src/unifyfs_client_rpcs.h
```diff
@@ -282,6 +283,105 @@ MERCURY_GEN_PROC(unifyfs_heartbeat_in_t,
 MERCURY_GEN_PROC(unifyfs_heartbeat_out_t, ((int32_t)(ret)))
 DECLARE_MARGO_RPC_HANDLER(unifyfs_heartbeat_rpc)
 
+/* unifyfs_node_local_extents_get_rpc (client => server)
+ *
+ * returns node local extents for a file
+ * given a global file id */
+MERCURY_GEN_STRUCT_PROC(unifyfs_index_t,
+                 ((uint32_t)(file_pos))
```
MichaelBrim, line 291: uint32_t is too small for file/log positions and extent length. use hg_size_t or uint64_t

common/src/unifyfs_client_rpcs.h
```diff
@@ -282,6 +283,105 @@ MERCURY_GEN_PROC(unifyfs_heartbeat_in_t,
...
+                 ((uint32_t)(log_pos))
+                 ((uint32_t)(length))
+                 ((int32_t)(gfid)))
+/* List of extents used in the output of RPC call*/
+typedef struct extents_list {
+    unifyfs_index_t value;
+    struct extents_list* next;
+} *extents_list_t;
```
MichaelBrim, line 299: hiding the fact that this is a pointer-to-struct in the typedef is confusing for later uses of extents_list_t. I suggest using extents_list_t as the typedef for the structure, and then using explicit pointer logic in the associated code.

common/src/unifyfs_client_rpcs.h
MichaelBrim: You're not tracking the client_id for the returned extents. The client_id is necessary to know which data log to read. With the current code, I think you have a bug where you are always reading the log for the current client, rather than the log for another client on the same node. There is additional work that will need to be done to setup the logio state for other clients and use it when reading extents from those clients. I think this also means you can't use the existing `extents` seg_tree in the client's file metadata to store the node-local extents, since `seg_tree` has no notion of other clients.

t/api/init-fini.c
```diff
@@ -19,11 +19,14 @@ int api_initialize_test(char* unifyfs_root,
...
     diag("Starting API initialization test");
 
-    int n_configs = 1;
-    unifyfs_cfg_option chk_size = { .opt_name = "logio.chunk_size",
-                                    .opt_value = "32768" };
+    int n_configs = 2;
+    unifyfs_cfg_option options[2];
+    options[0].opt_name = "logio.chunk_size";
+    options[0].opt_value = "32768";
+    options[1].opt_name = "client.node_local_extents";
```
MichaelBrim, line 26: I would prefer not to set this option in the API init test.

MichaelBrim (changes requested):
You should make this a draft PR until we can resolve the "reading other client's data" issue I mentioned.

client/src/client_read.c
```diff
@@ -653,15 +653,15 @@ int process_gfid_reads(unifyfs_client* client,
             /* There are files which are laminated and
              * require sync of extents */
             int extent_count = 0;
-            extents_list_t extents = NULL;
-            int ret =
```
hariharan-devarajan: Changed ret to rc as requested

t/api/init-fini.c
```diff
@@ -19,12 +19,10 @@ int api_initialize_test(char* unifyfs_root,
 {
     diag("Starting API initialization test");
 
-    int n_configs = 2;
-    unifyfs_cfg_option options[2];
+    int n_configs = 1;
+    unifyfs_cfg_option options[1];
     options[0].opt_name = "logio.chunk_size";
     options[0].opt_value = "32768";
-    options[1].opt_name = "client.node_local_extents";
```
hariharan-devarajan: removed from node-local-extents from default cases.

common/src/unifyfs_client_rpcs.h
```diff
@@ -288,26 +288,28 @@ DECLARE_MARGO_RPC_HANDLER(unifyfs_heartbeat_rpc)
...
+                 ((uint64_t)(length))
                  ((int32_t)(gfid))
                  ((int32_t)(log_app_id))
                  ((int32_t)(log_client_id)))
 /* List of extents used in the output of RPC call*/
 typedef struct extents_list {
     unifyfs_client_index_t value;
     struct extents_list* next;
-} *extents_list_t;
+} extents_list_t;
```
hariharan-devarajan, line 301: removed ptr in typedef and changed relevant code.

common/src/unifyfs_client_rpcs.h
```diff
@@ -288,26 +288,28 @@ DECLARE_MARGO_RPC_HANDLER(unifyfs_heartbeat_rpc)
...
+
+typedef extents_list_t* extents_list_t_ptr;
```
hariharan-devarajan, line 303: However, the macros don't work with a pointer so typedef is still needed over a pointer. I only use this here and nowhere else.

client/src/client_read.c
hariharan-devarajan: I fixed this check my comment below

common/src/unifyfs_client_rpcs.h
hariharan-devarajan: I fixed this check my comment below

common/src/unifyfs_client_rpcs.h
hariharan-devarajan: I fixed this check my comment below

t/api/init-fini.c
hariharan-devarajan: I fixed this check my comment below

client/src/client_read.c
```diff
@@ -597,9 +621,72 @@ int process_gfid_reads(unifyfs_client* client,
...
+                for (int j = 0; j < extent_count; ++j) {
+                    if (fetched_extents_cur->value.log_app_id ==
+                        client->state.app_id) {
+                        int fid = unifyfs_fid_from_gfid(client,
+                                          fetched_extents_cur->value.gfid);
+                        /* get meta for this file id */
+                        unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(
+                                client,
+                                fid);
+                        seg_tree_add(&meta->extents,
```
MichaelBrim, line 674: probably need to verify `meta != NULL` before adding to `meta->extents`

common/src/unifyfs_client_rpcs.h
```diff
@@ -282,6 +283,109 @@ MERCURY_GEN_PROC(unifyfs_heartbeat_in_t,
...
+                 ((uint64_t)(log_pos))
+                 ((uint64_t)(length))
+                 ((int32_t)(gfid))
+                 ((int32_t)(log_app_id))
+                 ((int32_t)(log_client_id)))
+/* List of extents used in the output of RPC call*/
+typedef struct extents_list {
+    unifyfs_client_index_t value;
+    struct extents_list* next;
+} extents_list_t;
```
MichaelBrim, line 301: I guess I'm still a little confused by your choice of a list-of-structures here, rather than a simple array-of-structures. Seems like it adds a bunch of extra encoding/decoding overhead versus just transferring the bulk array and using simple indexing.

hariharan-devarajan: So here, I need to send a list of extents on input as well as the output. It was more efficient to transfer it directly as its only metadata.

client/src/client_read.c
hariharan-devarajan: ok i added that.

client/src/client_read.c
```diff
@@ -597,9 +621,74 @@ int process_gfid_reads(unifyfs_client* client,
...
     read_req_t* reqs = NULL;
+    if (client->use_node_local_extents) {
+        extents_list_t* list = calloc(1, sizeof(struct extents_list));
+        struct extents_list* cur = list;
+        int num_request_selected = 0;
+        for (int i = 0; i < in_count; ++i) {
+            int fid = unifyfs_fid_from_gfid(client, in_reqs[i].gfid);
+            /* get meta for this file id */
+            unifyfs_filemeta_t* meta = unifyfs_get_meta_from_fid(client, fid);
+            if (!meta->attrs.is_laminated || !meta->needs_extents_sync) {
```
MichaelBrim, line 632: need to check `meta != NULL` here too

client/src/unifyfs_api_internal.h
```diff
@@ -37,6 +37,7 @@ typedef struct {
     int pending_unlink;           /* received unlink callback */
 
     int needs_sync;               /* have unsynced writes */
```
MichaelBrim, line 39: For clarity, I think we should rename these as:
`needs_sync` -> `needs_write_sync`
`needs_extents_sync` -> `needs_read_sync`

common/src/unifyfs_client_rpcs.h
MichaelBrim: I think you misunderstood my suggestion. If you just use an array of `unifyfs_client_index_t`, you don't need to write the custom serialization routine and do all the list pointer work. See for example `unifyfs_mread_rpc` where we pass an array of `unifyfs_extent_t` as bulk data.

client/src/unifyfs_api_internal.h
hariharan-devarajan: needs_read_sync feels like read data is synchronizing. Maybe we should call it needs_read_metadata/extents_sync

common/src/unifyfs_client_rpcs.h
hariharan-devarajan: I can use the bulk data while sending unifyfs_client_index_t but the return doesn't work for me. I am forced to write another RPC from server to client like in case of mread_rpc.

Can you point me to some documentation on how I can return bulk data from an RPC?

client/src/client_read.c
hariharan-devarajan: did the change

client/src/unifyfs_api_internal.h
hariharan-devarajan: did the change

common/src/unifyfs_client_rpcs.h
MichaelBrim: For client-server RPCs, it's all going through shared memory, not using RDMA. My experience is that put/get of whole arrays in shmem is way faster than iterating over a list where you encode/decode each list item.

The size of the array can be dynamic, you only need it to calculate the bulk size (`array size = # elements * sizeof(element)`).

To return bulk data, you just add the bulk args (size and data) to the output struct member list in the `MERCURY_GEN_PROC()`, same as is done for the input struct.

hariharan-devarajan: Ok, I actually missed pulling data from the request as well as response bulk data that's why it was not working for me.

I fixed that. I had to move the pull function to the common rpc_util file.

I have done the changes and tested it as well

client/src/client_read.c
```diff
@@ -597,9 +621,74 @@ int process_gfid_reads(unifyfs_client* client,
...
+                                         extents[j].file_pos,
+                                         extents[j].file_pos +
+                                         extents[j].length - 1,
+                                         extents[j].log_pos,
+                                         extents[j].log_client_id);
+                        }
+                    }
+                }
+            }
+            free(extents);
```
MichaelBrim, line 686: should probably move this inside the if block, or check for NULL before calling free()

client/src/margo_client.c
```diff
@@ -888,6 +889,102 @@ int invoke_client_mread_rpc(unifyfs_client* client,
...
+        ret = (int) out.ret;
+        if (ret == (int) UNIFYFS_SUCCESS) {
+            *extent_count = out.extent_count;
+
+            void* out_buffer = pull_margo_bulk_buffer(handle, out.bulk_data,
+                                                  out.bulk_size, NULL);
+            *extents = calloc(out.extent_count, sizeof(unifyfs_client_index_t));
+            unifyfs_client_index_t* out_extents  =
+                    (unifyfs_client_index_t*) out_buffer;
+            for (int i = 0; i < out.extent_count; ++i) {
```
MichaelBrim, line 961: why do you need this extra array allocation and copy for loop? just use 
`*extents = (unifyfs_client_index_t*) out_buffer;`

common/src/unifyfs_client_rpcs.h
```diff
@@ -282,6 +283,31 @@ MERCURY_GEN_PROC(unifyfs_heartbeat_in_t,
...
+MERCURY_GEN_STRUCT_PROC(unifyfs_client_index_t,
+                 ((uint64_t)(file_pos))
+                 ((uint64_t)(log_pos))
+                 ((uint64_t)(length))
+                 ((int32_t)(gfid))
+                 ((int32_t)(log_app_id))
+                 ((int32_t)(log_client_id)))
+MERCURY_GEN_PROC(unifyfs_node_local_extents_get_in_t,
+                 ((int32_t)(app_id))
+                         ((int32_t)(client_id))
```
MichaelBrim, line 299: please fix the alignment for L299-302

common/src/unifyfs_client_rpcs.h
```diff
@@ -282,6 +283,31 @@ MERCURY_GEN_PROC(unifyfs_heartbeat_in_t,
...
+                         ((hg_size_t)(num_req))
+                         ((hg_bulk_t)(bulk_data))
+                         ((hg_size_t)(bulk_size)))
+
+MERCURY_GEN_PROC(unifyfs_node_local_extents_get_out_t,
+                 ((int32_t)(ret))
+                         ((hg_size_t)(extent_count))
```
MichaelBrim, line 306: please fix the alignment for L306-308

MichaelBrim (changes requested):
@hariharan-devarajan Thanks, this is getting really close to being ready - just a few more requested changes. After that, we'll need to rebase/squash down to a single commit.

client/src/margo_client.c
hariharan-devarajan: Done

common/src/unifyfs_client_rpcs.h
hariharan-devarajan: Done

common/src/unifyfs_client_rpcs.h
hariharan-devarajan: Done

hariharan-devarajan:
> @hariharan-devarajan Thanks, this is getting really close to being ready - just a few more requested changes. After that, we'll need to rebase/squash down to a single commit.

I have rebased it :)

client/src/margo_client.c
```diff
@@ -888,6 +889,89 @@ int invoke_client_mread_rpc(unifyfs_client* client,
...
+    hret = margo_get_output(handle, &out);
+    if (hret == HG_SUCCESS) {
+        LOGDBG("Got response ret=%" PRIi32, out.ret);
+        ret = (int) out.ret;
+        if (ret == (int) UNIFYFS_SUCCESS) {
+            *extent_count = out.extent_count;
+            void* out_buffer = pull_margo_bulk_buffer(handle, out.bulk_data,
+                                                  out.bulk_size, NULL);
+            *extents = (unifyfs_client_index_t*) out_buffer;
+
```
MichaelBrim, line 958: remove extra line here

docs/configuration.rst
```diff
@@ -63,17 +63,18 @@ a given section and key.
...
+   Key                  Type    Description
+   ==================   ======  =================================================================
+   cwd                  STRING  effective starting current working directory
+   fsync_persist        BOOL    persist data to storage on fsync() (default: on)
+   local_extents        BOOL    service reads from local data if possible (default: off)
+   max_files            INT     maximum number of open files per client process (default: 128)
+   super_magic          BOOL    whether to return UNIFYFS (on) or TMPFS (off) statfs magic (default: on)
+   write_index_size     INT     maximum size (B) of memory buffer for storing write log metadata
+   write_sync           BOOL    sync data to server after every write (default: off)
+   node_local_extents   BOOL    service reads from node local data if possible for laminated files (default: off)
```
MichaelBrim, line 76: move this up between max_files and super_magic - table is sorted by key

server/src/unifyfs_request_manager.c
```diff
@@ -1440,6 +1444,94 @@ static int process_unlink_rpc(reqmgr_thrd_t* reqmgr,
...
 }
 
+static int process_node_local_extents_get_rpc(reqmgr_thrd_t* reqmgr,
+                                              client_rpc_req_t* req)
+{
+    int ret = UNIFYFS_SUCCESS;
+    unifyfs_node_local_extents_get_in_t* in = req->input;
+    assert(in != NULL);
+    size_t num_req = in->num_req;
+    void* buffer = pull_margo_bulk_buffer(req->handle, in->bulk_data,
```
MichaelBrim, line 1454: You should pull the bulk buffer in the RPC handler (`unifyfs_node_local_extents_get_rpc()`) so that the transfer happens in a lightweight argothread. Follow the pattern from `unifyfs_mread_rpc()`, which is do the pull, then on success, assign `client_rpc_req_t.bulk_buf & .bulk_sz`

MichaelBrim (changes requested):
Found a couple more minor things to update.

server/src/unifyfs_request_manager.c
hariharan-devarajan: Done

docs/configuration.rst
hariharan-devarajan: Done

client/src/margo_client.c
hariharan-devarajan: Done

server/src/unifyfs_request_manager.c
```diff
@@ -1440,6 +1444,90 @@ static int process_unlink_rpc(reqmgr_thrd_t* reqmgr,
...
+    out.ret = (int32_t) ret;
+    out.extent_count = total_chunks;
+    size_t extents_size = total_chunks*sizeof(unifyfs_client_index_t);
+    unifyfs_client_index_t* extents_buffer =
+            malloc(extents_size);
+    extents_list* cur = &out_list;
+    for (int i = 0; i < total_chunks; ++i) {
+        extents_buffer[i].file_pos = cur->value.file_pos;
+        extents_buffer[i].length = cur->value.length;
+        extents_buffer[i].gfid = cur->value.gfid;
```
MichaelBrim, line 1505: I think you are missing assignments to the log-related fields of the `unifyfs_client_index_t` here

hariharan-devarajan: Done

client/src/unifyfs_api.c
```diff
@@ -122,7 +131,17 @@ unifyfs_rc unifyfs_initialize(const char* mountpoint,
             client->use_local_extents = (bool)b;
         }
     }
-
+    /* Determine if we should track all extents on the node-local clients
```
MichaelBrim, line 134: Looks like this code block is a duplicate of the one above.

server/src/unifyfs_client_rpc.c
```diff
@@ -911,7 +911,7 @@ static void unifyfs_get_gfids_rpc(hg_handle_t handle)
         /* don't need the input struct any more, so free it now. */
         free(in);
         in = NULL;
-    }
+        }
```
MichaelBrim, line 914: fix spacing here

server/src/unifyfs_group_rpc.c
```diff
@@ -148,7 +149,6 @@ static int get_child_response(coll_request* coll_req,
 
             margo_free_output(chdl, out);
         }
-        free(out);
```
MichaelBrim, line 151: Why was this removed?

server/src/unifyfs_group_rpc.c
```diff
@@ -387,6 +367,8 @@ static int collective_forward(coll_request* coll_req)
     return ret;
 }
 
+
```
MichaelBrim, line 370: remove these empty lines

server/src/unifyfs_group_rpc.c
```diff
@@ -433,8 +415,7 @@ void collective_set_local_retval(coll_request* coll_req, int val)
     }
 }
 
-/* finish collective process by waiting for any child responses and
- * sending parent response (if applicable) */
+/* Forward the collective request to any children */
```
MichaelBrim, line 418: Revert to the previous comment

server/src/unifyfs_group_rpc.c
```diff
@@ -479,9 +462,10 @@ int invoke_bcast_progress_rpc(coll_request* coll_req)
     }
 
     /* get handle to local rpc function */
+    hg_handle_t handle;
```
MichaelBrim, line 465: Why did this change to a local variable? Revert to using the handle saved in the the coll_request

server/src/unifyfs_group_rpc.c
```diff
@@ -491,8 +475,7 @@ int invoke_bcast_progress_rpc(coll_request* coll_req)
          * by a ULT */
         bcast_progress_in_t in;
         in.coll_req = (hg_ptr_t) coll_req;
-        hret = margo_iforward(coll_req->progress_hdl, &in,
-                              &(coll_req->progress_req));
+        hret = margo_forward(handle, &in);
```
MichaelBrim, line 478: Revert to using margo_iforward()

server/src/unifyfs_request_manager.c
```diff
@@ -84,6 +86,8 @@ do { \
  * to process the read replies. It iterates with the client until
  * all incoming read replies have been transferred. */
 
+typedef struct extents_list extents_list;
```
MichaelBrim, line 89: this typedef should be in the same place as the struct definition

t/api/init-fini.c
```diff
@@ -20,10 +20,11 @@ int api_initialize_test(char* unifyfs_root,
     diag("Starting API initialization test");
 
     int n_configs = 1;
-    unifyfs_cfg_option chk_size = { .opt_name = "logio.chunk_size",
-                                    .opt_value = "32768" };
+    unifyfs_cfg_option options[1];
```
MichaelBrim, line 23: revert to prior code

MichaelBrim (changes requested):
I think most of the requested changes are due to unintended rebase effects.

server/src/unifyfs_client_rpc.c
hariharan-devarajan: > fix spacing here

I Checked the bracket matches the else indent.

t/api/init-fini.c
hariharan-devarajan: fixed

server/src/unifyfs_request_manager.c
hariharan-devarajan: fixed

server/src/unifyfs_group_rpc.c
hariharan-devarajan: fixed

server/src/unifyfs_group_rpc.c
hariharan-devarajan: fixed

server/src/unifyfs_group_rpc.c
hariharan-devarajan: fixed

server/src/unifyfs_group_rpc.c
hariharan-devarajan: fixed

client/src/unifyfs_api.c
hariharan-devarajan: fixed

MichaelBrim: the below block is still a duplicate. the above block should be reverted to update client->use_local_extents

server/src/unifyfs_group_rpc.c
```diff
@@ -54,7 +55,7 @@ static int forward_child_request(void* input_ptr,
     int ret = UNIFYFS_SUCCESS;
 
     /* call rpc function */
-    double timeout_ms = margo_server_server_timeout_msec;
+    double timeout_ms = UNIFYFS_MARGO_SERVER_SERVER_TIMEOUT_MSEC;
```
MichaelBrim, line 58: You need to revert this whole file to the current dev version and then add your new `#include unifyfs_rpc_util.h`

t/api/init-fini.c
```diff
@@ -21,7 +21,7 @@ int api_initialize_test(char* unifyfs_root,
 
     int n_configs = 1;
     unifyfs_cfg_option chk_size = { .opt_name = "logio.chunk_size",
-                                    .opt_value = "32768" };
+            .opt_value = "32768" };
```
MichaelBrim, line 24: revert this file to the dev version

MichaelBrim (changes requested):
There's still a few places where the rebase went wrong. See prior comments

client/src/unifyfs_api.c
MichaelBrim: now there is no duplication, but the original block that updates `client->use_local_extents` is missing

MichaelBrim (changes requested):
Just the one remaining issue. Almost there :)

client/src/unifyfs_api.c
hariharan-devarajan: ok added that back.

MichaelBrim (approved):
All looks good now. Thanks @hariharan-devarajan

adammoody (approved):

PR #736, opened 2022-09-08
unifyfs-ls util

rgmiller:
A first attempt at an `ls` like utility for UnifyFS
### Description

Creates a new `unifyfs-ls` utility.  In order to support this tool, this PR adds a new client-to-server RPC and two new API functions.  It also adds new unit tests.

### Motivation and Context

It would be very useful to have a stand-alone utility that can provide information about files in the UnifyFS filesystem.

### How Has This Been Tested?

Tested by running the stand-alone tool and by running the new unit test.  Test environment is an Ubuntu workstation.

### Types of changes

### Checklist:


rgmiller:
Marked the PR as a WIP.  There's a few TODO comments in the code that I'd like to discuss with reviewers.

client/src/margo_client.c
```diff
@@ -1211,3 +1212,118 @@ static void unifyfs_unlink_callback_rpc(hg_handle_t handle)
...
+    }
+
+    // copy the pointer so it can be returned to the caller
+    // (Caller will need to eventually call free() on this pointer!)
+    *gfid_list = _gfid_list;
+
+exit:  // normally, we use "out", but that's also the name of a variable
+    /* free resources */
+    if (ret != UNIFYFS_SUCCESS) {
+        free(_gfid_list);
```
adammoody, line 1311: Could this have an uninitialized value, which would then cause an error in free()?  For example, if an error is detected at line 1249?

server/src/unifyfs_client_rpc.c
```diff
@@ -863,3 +863,68 @@ static void unifyfs_mread_rpc(hg_handle_t handle)
...
+                req->bulk_buf = NULL;
+                req->bulk_sz = 0;
+                ret = rm_submit_client_rpc_request(&ctx, req);
+            }
+
+            if (ret != UNIFYFS_SUCCESS) {
+                if (NULL != req) {
+                    free(req);
+                }
+            }
```
adammoody, line 908: If ``ret`` is set to an error on line 901, I think the guard here prevents freeing ``req`` even though it may have been allocated successfully.

server/src/unifyfs_request_manager.c
```diff
@@ -1440,6 +1440,51 @@ static int process_unlink_rpc(reqmgr_thrd_t* reqmgr,
...
+
+    /* initialize bulk handle for the gfid_list */
+    hg_size_t segment_sizes[1] = { num_gfids * sizeof(int) };
+    void* segment_ptrs[1] = { (void*)gfid_list };
+    hg_return_t hret = margo_bulk_create(unifyfsd_rpc_context->shm_mid,
+                                         1, segment_ptrs, segment_sizes,
+                                         HG_BULK_READ_ONLY, &out.bulk_gfids);
+    /* Note: unifyfsd_rpc_context defined in margo_server.h */
+    if (hret != HG_SUCCESS) {
+        return UNIFYFS_ERROR_MARGO;
```
adammoody, line 1468: Need to free ``gfid_list`` in this case?

client/src/margo_client.c
rgmiller: Yep.  Good catch.  I'll make sure `_gfid_list` gets initialized to NULL up at the top of the function.  (Calling free() on NULL is explicitly allowed.)

server/src/unifyfs_client_rpc.c
rgmiller: I think the logic here is correct:   If the return value is NOT `UNIFYFS_SUCCESS`, then we need to free the request.

Testing for `NULL` before calling `free()` isn't necessary, though.

server/src/unifyfs_request_manager.c
rgmiller: Yep, you're right.  Another good catch.

adammoody (approved):

adammoody:
Thanks, @rgmiller !

Issue #744, opened 2022-12-01
unlink race condition in PnetCDF nc_test


adammoody:
In PnetCDF, the ``nc_test`` has a number of quick running tests that each create a scratch file, execute I/O operations on that file, and then delete the scratch file.  The consecutive sequence of tests shown here complete quickly and all use the same filename for the scratch file.

https://github.com/Parallel-NetCDF/PnetCDF/blob/6c71a30cd95f575c01025c0c926fc06dc9157774/test/nc_test/nc_test.c#L414-L420

The ``MPI_File_open()`` call of one of these tests fails when it attempts to sync extents with the server during an internal call to ``close()``.  ROMIO's ``MPI_File_open()`` calls ``open()`` and ``close()`` on each file.  The ``close()`` tries to sync extents because the file was opened for writing.

https://github.com/Parallel-NetCDF/PnetCDF/blob/6c71a30cd95f575c01025c0c926fc06dc9157774/test/nc_test/test_write.m4#L394

The extent sync fails because the file has been deleted so that ``meta->fid (-1) != fid (2)`` at this check:

https://github.com/LLNL/UnifyFS/blob/a13edaf779c755b3314f1bd7ce7f798d532d9951/client/src/unifyfs_fid.c#L1086-L1096

This situation happens because the prior test deleted the scratch file via ``MPI_File_delete() -> unlink()``.

https://github.com/Parallel-NetCDF/PnetCDF/blob/6c71a30cd95f575c01025c0c926fc06dc9157774/test/nc_test/test_write.m4#L437

That ``unlink()`` invokes an unlink rpc from client-to-server, which induces a later server-to-client unlink callback rpc that comes back to the client from the server at some future point in time, which happens to be in the middle of ``MPI_File_open``'s ``open()`` and ``close()`` calls.  The ``open()`` successfully recreates the file, then it is deleted due to the unlink callback, and then the ``close()`` fails to sync extents:

https://github.com/LLNL/UnifyFS/blob/a13edaf779c755b3314f1bd7ce7f798d532d9951/client/src/unifyfs-sysio.c#L2322-L2330

PR #745, opened 2022-12-05
unlink: optional sleep after calling client-to-server unlink rpc

adammoody:
In testing PnetCDF, some tests fail when creating a file after deleting a file by the same name (see https://github.com/LLNL/UnifyFS/issues/744).  As a work around, this adds an optional sleep immediately after a client calls the client-to-server unlink rpc to give the unlink operation more time to complete before the client returns from its call to ``unlink()``.

To enable this option, one can set a new config parameter:

```
export UNIFYFS_CLIENT_UNLINK_USECS=1000000
```

For the first test case that was failing, which was a serial program (single-process MPI job), a value of 1000000 (1 second) was sufficient.  Higher sleep times may be required for parallel jobs.

This is a hack, but it helps for now.

A better fix would be to implement a mode where the ``unlink()`` wrapper blocks at the calling client until all servers have indicated that the unlink operation has completed.  That may require a round trip between each server with each of its clients, since each client has to do some work to support unlink.  That change will be a more substantial effort, and so it is saved for future work.  Once added, this particular work around could be removed.

### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs_api.c
+++ b/client/src/unifyfs_api.c
@@ -183,6 +183,18 @@ unifyfs_rc unifyfs_initialize(const char* mountpoint,
     client->max_write_index_entries =
         client->write_index_size / sizeof(unifyfs_index_t);
 
+    /* Number of microsecs to sleep after calling client-to-server unlink rpc.
+     * This is a work around (hack) to try to give the unlink operation time
+     * to complete before the client returns from calling its unlink wrapper. */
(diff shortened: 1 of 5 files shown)


CamStan (approved):
Thanks @adammoody!

adammoody:
Thanks for taking a look, @CamStan !

PR #760, opened 2023-01-27
test: fopen existing and truncate on open

adammoody:
This adds a few test cases related to fixes in https://github.com/LLNL/UnifyFS/pull/746

1) Check that an existing file is truncated with ``open()`` in write mode and ``O_TRUNC``.
2) Check that ``fopen(file, "w")`` succeeds on an existing file.
3) Check that ``fopen(file, "w")`` truncates an existing file to 0 bytes.


adammoody:
Thanks @CamStan .  You pointed me right to the fix.

CamStan:
Oh good! Glad it worked!

CamStan (approved):

PR #779, opened 2023-06-23
define stat wrappers depending on whether symbol is found

adammoody:
### Types of changes

### Checklist:

### DIFF
--- a/client/src/unifyfs-stdio.c
+++ b/client/src/unifyfs-stdio.c
@@ -3043,9 +3043,10 @@ match_failure:
  * considered part of the scanset.
  */
 static const u_char*
-__sccl(tab, fmt)
-char* tab;
-const u_char* fmt;
+__sccl(char* tab, const u_char* fmt)
+//__sccl(tab, fmt)
+//char* tab;
+//const u_char* fmt;
 {
     int c, n, v;
 
--- a/client/src/unifyfs-sysio.c
+++ b/client/src/unifyfs-sysio.c
@@ -788,6 +788,11 @@ static int __stat(const char* path, struct stat* buf)
     return 0;
 }
 
+#if defined(HAVE_STAT64)     || \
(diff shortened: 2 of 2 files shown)


adammoody:
@wangvsa , I'm hitting some link time errors when building on tioga due to missing symbols ``stat64`` and ``fstat64``.  Have you tried to build there?

I'm building everything direct (no spack), which might be part of the problem.  I did find that I needed to set some variables to pick up the Cray compilers instead of the system gcc: ``./configure CC=cc CXX=CC ...``.  I made some tweaks to ``bootstrap.sh``, as well.  I can pass any of that along if you're doing the same.

What's strange is that configure claims to find these functions, but then I hit the link errors when building some of the examples.  I haven't tracked it down just yet.  I did find there are some definitions for these in ``/usr/lib64/libc_nonshared.a``, which maybe is where configure found them during its check.

Also, I made a few changes to some of the wrapper code.  Would you please review those changes (second commit)?

wangvsa:
I just tried on Tioga and got the same link issue. Adding the conditional check doesn't seem to solve the issue, since configure was able to find stat64/fstat64. Are there other changes you made?

adammoody:
No, I haven't figured it out yet, but it's useful to know that it's not just me.  I had some other tests to run, so I just commented out those wrappers for now as a work around.

I'm guessing it has something to do with the Cray compilers.  Are you building directly or via spack?

wangvsa:
I'm building it directly (without spack). I also set CC=cc CXX=CC so that bootstrap and configure pick the cray compiler.

MichaelBrim:
Not sure if it's relevant or helpful, but when I'm building from source on Crusher/Frontier I have recently needed to set CC=craycc and CXX=crayCC

MichaelBrim (approved):
looks good to me

Issue #782, opened 2023-07-06
Size mismatch in write() call (in test code)


rgmiller:
### System information

Type                   | Version/Name
---                    |     ---
Operating System       | Ubuntu
OS Version             | 22.04
Architecture           | x86_64
UnifyFS Version        | latest from git

### Describe the problem you're observing
https://github.com/LLNL/UnifyFS/blob/f95abed7c1aa91864d047b7c0ad38ac05e5719a7/t/sys/write-read.c#L379

The call to `write()` on this line specifies a count of 300, but the data to be written is only 2 bytes long.  Newer versions of gcc (at least 11.3.0, possibly somewhat earlier) will issue a warning about this.  And by default, we build with `-Werror` enabled, so the warning turns into an error and the build stops.

Temporary work-around is to compile with `-Wno-error=stringop-overread`.

### Describe how to reproduce the problem
Just look at the code.

### Include any warning or errors or releveant debugging data

The exact error from GCC:
```
./../t/sys/write-read.c: In function ‘write_pre_existing_file_test’:
../../t/sys/write-read.c:379:16: error: ‘write’ reading 300 bytes from a region of size 2 [-Werror=stringop-overread]
  379 |     rc = (int) write(fd, "a", 300);
      |                ^~~~~~~~~~~~~~~~~~~
```

adammoody:
Good catch.  I think maybe that ``buf`` was intended for this parameter instead of ``"a"``?

https://github.com/LLNL/UnifyFS/blob/f95abed7c1aa91864d047b7c0ad38ac05e5719a7/t/sys/write-read.c#L364

rgmiller:
That buffer is initialized to all zeros.  Is that sufficient for this test?  Or should we have different values in that buffer?

Also, the code uses the literal 300 in a couple of different places.  Should we replace that with a `#define` or `const int`?

MichaelBrim:
@rgmiller  Based on what it's testing, I don't think the zeroes are a problem. I agree that we should use a constant or #define for the buffer size.

PR #786, opened 2023-07-13
Fix warning about too small buffer in write() call

rgmiller:
Fix a warning about a write() call having too small a buffer for the number of bytes it was supposed to write.

Also replaced a pair of literal numbers that were used repeatedly with #defines.

Resolves issue #782

### Description

Simple bug fix:  use the already existing buffer as the source for a `write()` call instead of the too short string literal.

This change adds `#defines` for the test length and for a shorter test length and uses them instead of the integer literals `300` and `100` that are sprinkled throughout the function.

### Motivation and Context

Newer versions of GCC (starting with at least v11.3.0 and possibly earlier) issue a warning because of this bug.  Since the default for UnifyFS is to compile with `-Werror`, that warning aborts the build.  See issue #782.

### How Has This Been Tested?

Re-compile and note that the warning goes away.
Re-run the unit tests and make sure the changed test still passes and all the printed messages look correct.

### Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)
- [x ] Code cleanup (non-breaking change which makes code smaller or more readable)

### Checklist:

- [x ] My code follows the UnifyFS code style requirements.
- [x ] All new and existing tests passed.
- [x ] All commit messages are properly formatted.

### DIFF
--- a/t/sys/write-read.c
+++ b/t/sys/write-read.c
@@ -360,8 +360,11 @@ int write_pre_existing_file_test(char* unifyfs_root)
 {
     diag("Starting write-to-pre-existing-file tests");
 
+#define TEST_LEN 300
+#define TEST_LEN_SHORT 100
+/* TEST_LEN_SHORT must be less than TEST_LEN */
     char path[64];
-    char buf[300] = {0};
+    char buf[TEST_LEN] = {0};
     int fd = -1;
     int err, rc;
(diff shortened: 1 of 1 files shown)


adammoody (approved):

adammoody:
Thanks, @rgmiller

PR #798, opened 2023-09-19
Add resource manager support for Flux

wangvsa:
### Description

Tioga uses Flux to schedule jobs and has limited support for srun. Also Tioga does not have the `scontrol` command, which we use to retrieve the allocated node list.

This PR includes native support for Flux. It uses `flux run` to run clients and servers and uses `flux resource` to retrieve the number of nodes and the node list.
PS: `flux resource` returns a condensed node list, e.g., tioga[3-10, 12, 14]. The existing `parse_hostfile()` function can't handle this format, I added some code to parse it manually.

### How Has This Been Tested?

Tested on Tioga with 1, 2, 4 nodes. Also tested `unifyfs-ls`, `unifyfs-stage` and stage-in/out features.

### Types of changes

### Checklist:

### TODO

Unlike slurm where `SLURM_JOBID` can be used to determine a slurm allocation, flux only sets environment variables such as `FLUX_JOB_ID` for each `flux run` job (a flux job is similar to a slurm step). At the time of executing `unifyfs` (batch level), those variables have not been set yet.

A short flux script example: 
```bash
#!/bin/bash
# flux: -N4 -n256 -t 5m
# flux: --job-name="UnifyFS"
# flux: --queue="pdebug"

export UNIFYFS_LOGIO_SPILL_DIR=/tmp
export UNIFYFS_LOG_DIR=`pwd`
export UNIFYFS_LOG_VERBOSITY=3
export UNIFYFS_MOUNTPOINT=/unifyfs

# Here FLUX_JOB_ID are FLUX_JOB_NNODES are not set

unifyfs start -d --stage-in=`pwd`/manifest-in.txt --share-dir=`pwd`
flux run -n128 -N4 -c1 $UNIFYFS_DIR/libexec/write-static -m /unifyfs -f myTestFile
flux run -n128 -N4 -c1 $UNIFYFS_DIR/libexec/write-gotcha -f workflowTestFile
flux run -n128 -N4 -c1 $UNIFYFS_DIR/libexec/read-gotcha  -f workflowTestFile
flux run -n16 -N4 -c1 unifyfs-stage --parallel --status-file=/tmp/stage-out-status.txt `pwd`/manifest-out.txt 
unifyfs-ls 
unifyfs terminate --cleanup

```

***As a result, currently I use `FLUX_EXEC_PATH` to determine if the system has flux scheduler. I feel this is not optimal but I couldn't figure out a better way.***

### DIFF
--- a/util/unifyfs/src/unifyfs-rm.c
+++ b/util/unifyfs/src/unifyfs-rm.c
@@ -617,6 +617,102 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
     return ret;
 }
 
+/**
+ *
+ * @brief Get list of nodes using Flux resource
+ *
+ * @param resource  The job resource record to be filled
+ *
+ * @return 0 on success, negative errno otherwise
+ */
+static int flux_read_resource(unifyfs_resource_t* resource)
+{
+    size_t n_nodes = 0;
+    char  num_nodes_str[10] = {0};
+    char  nodelist_str[1024] = {0};
+    char* ret = NULL;
+    FILE* pipe_fp = NULL;
+    char** nodes = NULL;
+    int node_idx = 0;
+
+    // get num nodes using flux resource command
+    pipe_fp = popen("flux resource list --states=free -no {nnodes}", "r");
+    ret = fgets(num_nodes_str, sizeof(num_nodes_str), pipe_fp);
+    if (ret == NULL) {
+        pclose(pipe_fp);
+        return -EINVAL;
+    }
+    n_nodes = (size_t) strtoul(num_nodes_str, NULL, sizeof(num_nodes_str));
+    pclose(pipe_fp);
+
+    nodes = calloc(sizeof(char*), n_nodes);
+    if (nodes == NULL) {
+        return -ENOMEM;
+    }
+
+    // get node list using flux resource command
+    // the returned list is in a condensed format
+    // e.g., tioga[18-19,21,32]
+    // TODO: is it safe to assume flux resource only
+    // returns a single line?
+    pipe_fp = popen("flux resource list --states=free -no {nodelist}", "r");
+    ret = fgets(nodelist_str, sizeof(nodelist_str), pipe_fp);
+    if (ret == NULL) {
+        pclose(pipe_fp);
+        return -EINVAL;
+    }
+    pclose(pipe_fp);
+
+    // remove the trailing ']'
+    nodelist_str[strlen(nodelist_str)-1] = 0;
+    // get the node ids, i.e., the list in []
+    char* node_ids = strstr(nodelist_str, "[");
+    if (node_ids) {
+        char* host = calloc(1, (node_ids-nodelist_str)+2);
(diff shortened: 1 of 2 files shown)


util/unifyfs/src/unifyfs-rm.c
```diff
@@ -617,6 +617,102 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
+    pipe_fp = popen("flux resource list --states=free -no {nodelist}", "r");
+    ret = fgets(nodelist_str, sizeof(nodelist_str), pipe_fp);
+    if (ret == NULL) {
+        pclose(pipe_fp);
+        return -EINVAL;
+    }
+    pclose(pipe_fp);
+
+    // remove the trailing ']'
+    nodelist_str[strlen(nodelist_str)-1] = 0;
```
adammoody, line 667: Is the trailing ']' still here when only allocating one node?

adammoody: I think it's missing, actually:
```
>>: flux alloc -q pdebug -N 1
>>: flux resource list --states=free -no '{nodelist}'
tioga20
```

adammoody: If you want to avoid parsing directly, another option is to rely on ``/bin/hostlist``.
```
>>: /bin/hostlist -e 'tioga[18-19,21,32]'
tioga18,tioga19,tioga21,tioga32
>>:  /bin/hostlist -e 'tioga20'
tioga20
```
Though I think that command might just be available on LLNL systems and is not distributed with flux.

I think I've seen that flux has some python packages that parse hostlists, too.  I can dig that up if you're interested.

I'm also fine with parsing the hostlist directly.

util/unifyfs/src/unifyfs-rm.c
```diff
@@ -617,6 +617,102 @@ static int slurm_read_resource(unifyfs_resource_t* resource)
...
+                nodes[node_idx++] = strdup(nodename);
+            }
+            token = strtok_r(NULL, ",", &end_str);
+        }
+    } else {
+        // no '[' in the string, meaning it has a single node
+        if (n_nodes != 1) {
+            return -EINVAL;
+        }
+        nodes[0] = strdup(nodelist_str);
```
adammoody, line 708: Oh, now I see you're already accounting for the single node case here.  Do you just need to move the 
```
 // remove the trailing ']'
    nodelist_str[strlen(nodelist_str)-1] = 0;
```
lines inside of the if block?

util/unifyfs/src/unifyfs-rm.c
wangvsa: It turns out the last character of the buffer returned from fgets is either \n or EOF. So my code actually removes that instead of ], and it just happens that the rest of the code still works fine with ] in the buffer. Anyway, I have fixed this and now I will remove the last two characters all together.

adammoody (approved):

adammoody:
Thanks, @wangvsa

